PR #1413, opened 2026-04-20
fix(infra): unblock FleetDM deploy via remediation + resource management

devantler:
## Problem

FleetDM failed to deploy on the dev cluster. The `apps` Flux Kustomization is blocked by a cascade failure:

```
Kyverno policy sets retries:3 without remediateLastFailure:true
  → Dex HelmRelease upgrade times out (default 5m on 96-98% memory CX23 nodes)
  → After 4 attempts exhausted → dex permanently Stalled
  → infrastructure-controllers health check fails
  → infrastructure blocked → apps blocked → FleetDM never created
```

**CI didn't catch this** because the Docker provider has more memory available and the permanent-stalling bug only manifests after multiple retries on a long-lived cluster.

### Phase 1 — Unblock the deployment chain
- **Kyverno remediation policy**: Add `remediateLastFailure: true` so HelmReleases self-heal after exhausting retries
- **Dex HelmRelease**: Add `spec.timeout: 10m` (default 5m too short for memory-constrained nodes; also resets the current Stalled state via generation bump)

### Phase 2 — Replace Goldilocks with Kyverno auto-VPA
- Remove Goldilocks controller (saves 3 pods)
- Strip `goldilocks.fairwinds.com/*` annotations from all namespaces
- New `auto-vpa` ClusterPolicy auto-generates VPAs for every Deployment/StatefulSet with `controlledResources: ["memory"]` — VPA only right-sizes memory, CPU stays under KEDA control
- Remove Goldilocks routes from auth-proxy and reference-grant

### Phase 3 — KEDA for autoscaling + scale-to-zero
- Install KEDA as infrastructure controller
- ScaledObjects with `minReplicaCount: 0` for dashboards (homepage, headlamp, whoami) — scale to zero when idle
- ScaledObjects with `minReplicaCount: 1` for always-on services (dex, oauth2-proxy, fleetdm) — horizontal autoscale on CPU
- Docker provider patches pin all ScaledObjects to 1 replica for deterministic CI

## References

- [Running HPA and VPA together on Kubernetes](https://brtkwr.com/posts/2026-02-07-running-hpa-and-vpa-together-on-kubernetes/)
- [KEDA documentation](https://keda.sh/docs/latest/)


devantler:
@copilot+claude-opus-4.6 create a stacked pr on top of this one, that adds true scaling-to-zero with https://github.com/kedacore/http-add-on

Make sure to consider if the addon is available via the existing HelmRelease, or if we must add another.

PR #1641, opened 2026-05-28
refactor(cd): use installer actions for cosign and syft

devantler:

## Why

[#1640](https://github.com/devantler-tech/platform/pull/1640) is the minimal fix for the broken CD pipeline — it bumps two pinned SHA256 digests that Renovate version-only bumps left stale. But the underlying problem repeats every time Renovate touches `COSIGN_VERSION` or `SYFT_VERSION`: the contract documented in the workflow comment ("Renovate bumps the version only; the digest must follow in the same PR — intentional") relies on a human noticing, and auto-merge keeps slipping past it.

This PR removes the contract entirely by switching both binaries to their **official installer actions**, which manage binary integrity themselves:

| Action | What it replaces | Integrity check |
|---|---|---|
| [`sigstore/cosign-installer@v4.1.2`](https://github.com/sigstore/cosign-installer) | curl + sha256sum + sudo install of `cosign-linux-amd64` | Keyless-verifies the cosign binary against Fulcio + Rekor on every run |
| [`anchore/sbom-action@v0.24.0`](https://github.com/anchore/sbom-action) | curl + sha256sum + tar of `syft_*.tar.gz`, plus the raw `syft scan` invocation | Installer manages syft's binary integrity; we just pass `syft-version` |

Both actions are themselves SHA-pinned (matching the repo's existing convention for `actions/checkout`, `actions/attest-sbom`, etc.).

## What this preserves

- **`COSIGN_VERSION` / `SYFT_VERSION` env vars stay**, with the same `# renovate:` annotations. The custom regex manager in `.github/renovate.json` (which already targets `*_VERSION` env vars) keeps proposing bumps for both, and those bumps are now single-value — no second value to forget.
- **All existing signing behaviour:** keyless cosign sign + recursive, GHCR registry login, digest capture in `steps.cosign-sign.outputs.digest`, the same SBOM at `sbom.cdx.json` consumed by `actions/attest-sbom`, the same SLSA provenance attestation, same OIDC permissions.
- **No new `docker/login-action` dependency** — kept the bare `docker login` shell step (split out so the registry credentials are visible in one place rather than re-used inside an action input).

## What this removes

- `COSIGN_SHA256` and `SYFT_SHA256` env vars and their explanatory comments.
- The `curl + sha256sum -c + sudo install` blocks for both tools.
- The defensive `jq -e '.bomFormat == "CycloneDX"' sbom.cdx.json` sanity check — `anchore/sbom-action` either succeeds (and writes a valid file) or fails the step; the home-rolled defence was insurance against a home-rolled syft invocation silently emitting garbage, which the action doesn't do.

Net: `cd.yaml` is **−45 / +38 lines**.

## Relationship to #1640

- **Either PR alone unblocks CD.** [#1640](https://github.com/devantler-tech/platform/pull/1640) does it with the minimal possible diff; this PR does it structurally and removes future toil.
- **They will conflict** (both touch the same two step blocks). Pick one merge order:
  - **Merge #1640 first** (minimal patch ships, CD goes green) → then this PR (refactor on top, trivial conflict).
  - **Merge this PR first** → close [#1640](https://github.com/devantler-tech/platform/pull/1640) as superseded.
- I have no preference; the maintainer choice depends on how confident you want to be in the installer actions vs. the simpler patch.

## Verification

Workflow-only change. YAML parses (`ruby -ryaml -e 'YAML.load_file'` → OK). Both action commits are the upstream-published commits for their respective tags:

- `sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6` → `refs/tags/v4.1.2`
- `anchore/sbom-action@e22c389904149dbc22b58101806040fa8d37a610` → `refs/tags/v0.24.0`

Real validation is the next `v*` tag deploy — CD has no static cluster-free way to test the signing path.

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -128,6 +128,20 @@ jobs:
           GHCR_TOKEN: ${{ secrets.GHCR_TOKEN }}
           HCLOUD_TOKEN: ${{ secrets.HCLOUD_TOKEN }}
 
+      - name: ⚙️ Install cosign
+        # sigstore/cosign-installer is the official installer action — it
+        # downloads cosign for the requested release and verifies its own
+        # Sigstore signature against Fulcio + Rekor (keyless), so there is
+        # no per-version SHA256 to track in this repo. Bump cosign by
+        # editing COSIGN_VERSION below; Renovate proposes the bump via the
+        # custom regex manager in .github/renovate.json.
+        env:
+          # renovate: datasource=github-releases depName=sigstore/cosign extractVersion=^v(?<version>.+)$
+          COSIGN_VERSION: "3.0.6"
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+        with:
+          cosign-release: 'v${{ env.COSIGN_VERSION }}'
+
       - name: 🖋️ Sign OCI manifest with cosign (keyless)
         # Signs the platform OCI artifact published by `ksail workload push`
         # using Sigstore keyless signing: Fulcio mints a short-lived cert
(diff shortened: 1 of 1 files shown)


devantler:
Superseded by #1642 — same refactor, rebased onto post-#1640 main. Closing here because force-push to the original branch wasn't available; #1642 has the identical resolved content (`refactor(cd): use installer actions for cosign and syft`, −45/+38 in cd.yaml).

🤖 Generated by the Daily AI Assistant

PR #1646, opened 2026-05-28
ci: discount Unhealthy probe warnings on deleted pods in the deploy gate

devantler:

Follow-up to #1637 (the deterministic fix for the rollout teardown race discussed there).

## Problem

The merge-queue **Deploy to Prod** gate (`.github/actions/check-event-warnings`) fails on *any* `Warning` event in its 90s settle window. During a rollout, kubelet fires one last liveness/readiness probe ~1s after Cilium tears down a *terminating* pod's route:

```
Unhealthy: …:9003/healthz: connect: no route to host
```

That event is on a pod that is **already deleted** — a teardown artifact, not a steady-state fault — yet the gate counted it and failed the deploy. It hit OpenCost in #1637 and is a recurring flaky-failure source for any PR that rolls a pod. The gate's own description says it targets warnings "still firing at steady state"; a one-shot probe on a gone pod isn't that.

## Fix

Snapshot live pods alongside the events, and partition off (don't count) `Unhealthy` warnings whose involved Pod is absent from the snapshot.

| property | behaviour |
|---|---|
| Scope | **only** `reason == "Unhealthy"` on a `Pod` not in the live snapshot |
| Real crash-loops / persistent probe failures | **still counted** — those pods *exist* (CrashLoopBackOff, etc.) |
| Other reasons (`BackOff`, `FailedMount`, …) on deleted pods | **still counted** (not discounted) |
| Snapshot unavailable | **fail-safe** — keep every warning (never silently hide because the snapshot broke) |
| Transparency | discounted warnings printed in a report-only `::group::` so nothing is masked without a trace |
| Event shapes | both core/v1 (`.involvedObject`) and events.k8s.io/v1 (`.regarding`) |

## Why this is safe

The gate's purpose is to catch problems *still firing at steady state*. By definition those occur on pods that **exist** — a crash-looping pod is in `CrashLoopBackOff`, a broken-readiness pod is `Running`-but-unready. Only the artifacts of pods that have **ceased to exist** are discounted, and even then only the probe-failure (`Unhealthy`) reason. Everything else is untouched, and the discounted set is logged.

## Validation

- jq partitioning **unit-tested** across every case: deleted-pod `Unhealthy` → dropped; live-pod `Unhealthy`, non-`Unhealthy` on deleted pods, non-Pod warnings, pre-marker events → correct; `null` snapshot → keep all.
- `shellcheck` clean on the embedded run script; `actionlint` clean; `yq` parses.

Note: this changes CI tooling only (no platform manifests), so it's a `ci:` change and shouldn't trigger a platform release.

### DIFF
--- a/.github/actions/check-event-warnings/action.yaml
+++ b/.github/actions/check-event-warnings/action.yaml
@@ -5,8 +5,12 @@ description: >
   inspects Warning events whose most-recent occurrence is at/after the marker —
   i.e. warnings still firing at steady state (crash loops, repeated probe
   failures, image back-off, etc.). Transient one-shot warnings emitted during
-  bootstrap are ignored because they fired before the marker. The full Warning
-  history is always printed for context.
+  bootstrap are ignored because they fired before the marker. "Unhealthy"
+  (probe-failure) warnings on pods that no longer exist are discounted as
+  rollout teardown artifacts (a final probe after the CNI removed a terminating
+  pod's route); real crash loops and persistent probe failures occur on pods
+  that still exist and remain counted. The full Warning history — and anything
+  discounted — is always printed for context.
 
 inputs:
   context:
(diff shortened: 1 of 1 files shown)


devantler:
I do not want this. It is a hacky workaround

PR #1665, opened 2026-05-29
fix(longhorn): reduce guaranteed-instance-manager-cpu 12% → 6%

devantler:
> ⛔ **HOLD — do not merge until the worker-2 replacement (4th worker) is back and Longhorn has replica-rebuild headroom.** Applying this **restarts the Longhorn data plane** (instance-managers). The cluster is currently at **3 workers with replica count 3 (no N+1)** after the worker-2 incident, so a data-plane restart now risks volume availability. Kept as a draft for exactly this reason.

## Summary

From a live per-container measurement of prod (2026-05-29, read-only): **6 Longhorn `instance-manager` pods each reserve ~474m (12% of a cx33) while using 8–30m** — ~2.7 cores locked, and instance-managers are **not** VPA-managed (so the auto-vpa policy can't right-size them).

## Change

`guaranteedInstanceManagerCPU` 12% → **6%** in the Longhorn HR `defaultSettings` (v1 + v2). 6% ≈ 240m on a cx33 — still **8–16× headroom** over observed usage — and frees **~1.4 cores**, while leaving margin for the CPU spikes during Longhorn replica rebuilds.

## Why a separate, held PR

The companion right-sizing change ([#1662](https://github.com/devantler-tech/platform/pull/1662)) is GitOps-safe to land anytime. This one is **not**: changing the guaranteed CPU restarts instance-managers (the volume data plane). It must wait until the cluster is back to 4 workers and stable.

## Validation

`kubectl kustomize k8s/clusters/prod/` builds. (Helm values aren't rendered by Kustomize; the `guaranteedInstanceManagerCPU` v1/v2 map form matches the 1.11.2 chart's `longhorn.multiTypeSetting` helper.)


devantler:
> 🤖 Generated with Claude Code — investigated the merge-queue failure.

**Why it failed (51 Warning events):** dominated by Longhorn **data-plane churn** — `longhorn-manager` pods down, nodes reported `down: manager pod not running`, `Waiting for disk … to be ready/schedulable`, `engine-image … cannot access /data/longhorn`, `longhorn-csi-plugin FailedPreStopHook` — plus collateral readiness/liveness failures on pods riding the churning worker-2/worker-3 nodes (homepage, wedding-app, fleetdm, dex, oauth2-proxy, keda, vpa). This is exactly what changing `guaranteedInstanceManagerCPU` does: **Longhorn recreates every instance-manager pod to apply the new CPU guarantee**, briefly disrupting the volumes they serve. The `check-event-warnings` gate flags those events.

(The `add-ns-quota` terminating-namespace PolicyErrors in the list are incidental — this branch predates the precondition fix in #1662.)

**Root cause:** a Longhorn *data-plane* change applied while the cluster is fragile — **3 workers with `longhorn_replica_count: 3` = no N+1 rebuild headroom** (worker-2 was removed during the incident; the 4th is blocked on a Hetzner cx33 shortage). With no spare replica, simultaneous instance-manager restarts can't keep volumes available → the storm. Prod has since recovered (the candidate rolled back) — no lasting harm.

**There is no manifest change that makes recreating instance-managers non-disruptive**, so forcing this green isn't possible, and each merge-queue attempt re-disrupts prod. I've therefore **converted this PR back to draft** (it should not have left the HOLD state).

**Prerequisites before it can merge cleanly:**
1. Clean the orphaned `kubevirt`/`cdi` APIServices + webhooks so those namespaces finish terminating (also clears the `add-ns-quota` warnings both PRs inherit).
2. Restore the 4th worker → Longhorn N+1, so instance-manager restarts roll one-at-a-time while volumes stay available.
3. Rebase on `main` to inherit #1662's `add-ns-quota` precondition.
4. Even then, instance-manager restarts emit some warnings — best applied in a maintenance window.

**Worth reconsidering the ROI:** this frees ~1.4 cores of *CPU* reservation, but the workers are **memory-bound** (CPU is only ~28% utilized), so it doesn't relieve the binding constraint. Given the deploy friction, deferring indefinitely — or dropping it in favour of the GitOps-safe floor reclaim in #1662 — is a reasonable call.

PR #1661, opened 2026-05-29
feat(cluster-policies): require workloads to spread across nodes

devantler:
## What

The `spread-pods` ClusterPolicy is synced verbatim from upstream `kyverno/policies` by [`.github/workflows/sync-cluster-policies.yaml`](.github/workflows/sync-cluster-policies.yaml) and adapted for this cluster via a Kustomize patch in [`k8s/bases/infrastructure/cluster-policies/kustomization.yaml`](k8s/bases/infrastructure/cluster-policies/kustomization.yaml). This change hardens that patch so spreading workloads across nodes is **strictly required** rather than best-effort:

- `whenUnsatisfiable: ScheduleAnyway` → **`DoNotSchedule`** — every `Deployment`/`StatefulSet` (outside `kube-system`, carrying an `app.kubernetes.io/name` label) must spread across nodes (`topologyKey: kubernetes.io/hostname`, `maxSkew: 1`).
- Added **`matchLabelKeys: [pod-template-hash]`**.

## Why

`DoNotSchedule` + `maxSkew: 1` on its own deadlocks rolling updates: during a rollout the new revision's surge pod is counted against the old revision's pods, so it can't be placed. `matchLabelKeys: [pod-template-hash]` scopes the skew calculation per ReplicaSet revision, keeping rollouts flowing while enforcement stays strict.

- Requires Kubernetes ≥ 1.27 (GA in 1.34). This cluster runs k8s ~1.34 via Talos `v1.12.4`, so it is on by default.
- StatefulSets have no `pod-template-hash` label; Kubernetes ignores absent keys and falls back to the label selector — fine, since StatefulSet updates roll one pod at a time.

## Reviewer notes

- The change is confined to the **Kustomize patch**; the upstream-synced sample (`samples/other/spread-pods-across-topology/`) is untouched, so it keeps receiving upstream updates for free.
- **Behavioral impact:** pods can now go `Pending` if they genuinely cannot spread (e.g. a worker is drained/down on the 3-worker local cluster and a multi-replica app needs a distinct node). Single-replica apps are unaffected; rollouts are protected by `matchLabelKeys`.
- Workloads without an `app.kubernetes.io/name` label are skipped (the policy needs it to group replicas) — unchanged behavior.
- Companion `insert-pod-antiaffinity` policy left soft (preferred) — topology spread is now the hard enforcer.

## Validation

- `ksail workload validate` — local + prod, 256 files each, exit 0
[...]

PR #1677, opened 2026-06-01
fix(ci): flag only event warnings still firing after the settle, not transients

devantler:
## Root cause

The `check-event-warnings` deploy gate is what failed the recent #1659 reconcile:

```
New Warning events since 2026-06-01T21:09:00Z (4)
  [flux-system] Pod/helm-controller-…      Unhealthy: Readiness probe failed: …:9440/readyz: connection refused
  [flux-system] Pod/kustomize-controller-… Unhealthy: Readiness probe failed: …:9440/readyz: connection refused
  [flux-system] Kustomization/infrastructure-controllers  HealthCheckFailed: … context canceled
  [flux-system] Pod/notification-controller-…  Unhealthy: Readiness probe failed: …:9440/readyz: connection refused
```

The action's docstring says it *ignores transient one-shot warnings*. But the implementation records the marker **before** the settle and flags any Warning with `lastTimestamp >= marker`. So a one-shot warning that fires **during** the settle window and then stops is flagged anyway — it only actually ignores transients that fired *before the deploy started*.

That false-positives on **any deploy that rolls a workload during the settle**. Concretely: applying the Flux-controller `topologySpreadConstraints` (#1659) — or any controller config change — restarts the controllers. While a controller restarts, its `/readyz` returns `connection refused` for a few seconds (leader-election handoff), and `kustomize-controller` restarting cancels the in-flight `infrastructure-controllers` health check (`context canceled`). All four events are transient and self-heal within seconds — the controllers are `1/1 Running` immediately after — yet the gate failed the deploy.

## Fix

Record the marker **after** the settle, then observe for a short window (new `observe-seconds` input, default `60`):

```
settle (90s, grace — rollouts converge)  →  marker  →  observe (60s)  →  sample
```

Only warnings whose most-recent occurrence is at/after the post-settle marker are flagged — i.e. **still firing once the cluster has settled**, which is the documented "steady state" intent. Persistent problems (crash loops, `FailedAttachVolume`, missing PriorityClass — all of which fire continuously) keep firing into the observe window and are **still caught**; one-shot rollout blips that drain during the settle are not.

This is a correctness fix, not a loosening: it makes the gate do what its docstring already claims.

## Validation

- `action.yaml` parses; embedded bash passes `bash -n`.
- Fed the gate's `jq` filter the **exact `21:09:17` readiness-blip event** (before a post-settle marker) plus a synthetic `BackOff ×42` crash-loop (after it): only the crash-loop is flagged. ✅
- No caller changes needed — all three call sites (`ci.yaml` system-test + merge_group, `cd.yaml`) use defaults; `observe-seconds` defaults to 60.

## Relationship to other PRs

This unblocks **#1659** (Flux controller spread) and any future controller-config deploy. Note #1661 ("require workloads to spread across nodes") merged recently — worth confirming whether #1659 is still needed or is now redundant with that policy.

### DIFF
--- a/.github/actions/check-event-warnings/action.yaml
+++ b/.github/actions/check-event-warnings/action.yaml
@@ -1,12 +1,13 @@
 name: Check for new event warnings
 description: >
-  Fail when a cluster is still emitting Warning events after a successful
-  deployment. Records a marker timestamp, lets the cluster settle, then
-  inspects Warning events whose most-recent occurrence is at/after the marker —
-  i.e. warnings still firing at steady state (crash loops, repeated probe
-  failures, image back-off, etc.). Transient one-shot warnings emitted during
-  bootstrap are ignored because they fired before the marker. The full Warning
-  history is always printed for context.
+  Fail when a cluster is still emitting Warning events at steady state after a
+  deployment. Lets the cluster settle FIRST (so one-shot warnings emitted while
+  the deploy rolls out — e.g. a controller's readiness probe blipping while it
+  restarts to pick up a config change — drain away), THEN records a marker and
+  observes for a short window, flagging only Warning events whose most-recent
+  occurrence is at/after the marker — i.e. warnings still actively firing past
+  the settle (crash loops, repeated probe failures, image back-off). The full
+  Warning history is always printed for context.
 
 inputs:
   context:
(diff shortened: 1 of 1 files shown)


devantler:
Superseded by #1678: rather than fix the event-warnings gate's marker timing, we're removing the gate entirely and relying on Flux reconcile status (`ksail workload reconcile` already waits per-Kustomization; all Kustomizations are `wait: true` so Ready encodes Flux's health checks). Per maintainer preference for trusting Flux/ksail over bespoke CI heuristics.

PR #1690, opened 2026-06-02
feat(security): add opt-in Tetragon enforcement for system-file integrity

devantler:
> 📚 **Stacked on #1689** (base branch: `claude/tetragon-tracingpolicies`), which is stacked on #1688. Merge in order: #1688 → #1689 → this.

## What

PR 3 (final) of the Tetragon rollout — adds **runtime enforcement**, designed to be safe by construction.

`k8s/bases/infrastructure/tracing-policies/enforce-protect-system-files.yaml` — a `TracingPolicy` that **`Sigkill`s** any process writing to system / binary directories (`/etc`, `/usr/bin`, `/lib`, `/lib64`, `/boot`, `/usr/local/*`, `/root/.ssh`, …), a classic tampering / persistence pattern.

## Safe by construction

- **Opt-in, zero default blast radius:** `spec.podSelector` matches only pods labelled `tetragon.devantler.tech/enforce-file-integrity: "true"`. With nothing labelled, this policy enforces on **nothing**. (Works via the agent policy filter enabled in PR 1.)
- **Writes only** — reads are left to PR 2's observe policy, because system daemons legitimately read `/etc/shadow`, `/etc/sudoers`, etc. Killing on read would be catastrophic.
- **Excludes `/var/log` and `/dev/log`** to avoid killing legitimate logging even for opted-in pods.

## Rollout path

1. Merge; nothing changes (no pods labelled).
2. Label one workload's pods, watch PR 2's observe events in Loki to confirm no legitimate writes would be killed.
3. Widen the label as confidence grows.

## Sigkill vs Override

`Sigkill` is a post-hook signal and needs **no special kernel config**. The stronger `Override` action (fail the syscall *before* it runs) requires `CONFIG_BPF_KPROBE_OVERRIDE` — verify that on the pinned Talos kernel before switching. This is documented inline in the policy.

## Validation

| Check | Local | Prod |
|---|---|---|
| `ksail workload validate` | ✅ 266 files | ✅ 266 files |
| `kubectl kustomize` provider infra overlay | ✅ 2 TracingPolicies | ✅ 2 TracingPolicies |

### DIFF
--- a/k8s/bases/infrastructure/tracing-policies/enforce-protect-system-files.yaml
+++ b/k8s/bases/infrastructure/tracing-policies/enforce-protect-system-files.yaml
@@ -1,0 +1,65 @@
+---
+# ENFORCING TracingPolicy: SIGKILLs any process that attempts to WRITE to
+# system / binary directories (a classic tampering / persistence pattern).
+#
+# OPT-IN, ZERO DEFAULT BLAST RADIUS:
+#   spec.podSelector matches only pods carrying the label
+#   `tetragon.devantler.tech/enforce-file-integrity: "true"`. With no pod
+#   labelled, this policy enforces on nothing. Requires the policy filter,
+#   which is enabled on the agent (tetragon.enablePolicyFilter: true).
+#
+# This deliberately does NOT enforce on /var/log or /dev/log (legitimate
+# log writes) and does NOT enforce on file *reads* -- system daemons read
+# /etc/shadow etc. legitimately. Reads are covered observe-only by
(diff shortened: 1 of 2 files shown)


devantler:
Closing — enforcement **consolidated into #1710**, rebased off current `main` (so it drops this stack's stale-base / force-push problems). The opt-in SIGKILL file-integrity policy lives there, with the MAY_WRITE matcher hardened from `Equal` to a bitmask `Mask` so `O_RDWR` opens (mask 0x6) can't evade enforcement.

PR #1697, opened 2026-06-02
docs(security): plan enforcing cosign verification of the platform OCI artifact

devantler:
## What

The verify-enforcement spike from the supply-chain work. Adds [`docs/supply-chain-verification.md`](docs/supply-chain-verification.md). **No manifests changed.**

## Why this is a doc, not a manifest

The goal was to wire Flux `OCIRepository` `spec.verify` so the cluster *rejects* a platform artifact not signed by our CI. Investigating it surfaced a hard blocker that makes a live config change unsafe right now:

- **KSail owns the root `flux-system` `OCIRepository`** — it's created from `ksail.prod.yaml`'s `localRegistry`, not declared in this repo. Its full spec (interval, `layerSelector`, and the **pull `secretRef`** KSail generates) lives only in-cluster. A hand-written override must reproduce all of it exactly; a wrong `secretRef` means the source can't pull.
- **`FluxInstance.spec.sync` has no `verify` field** (schema-confirmed) — so verification can't be added through the resource we already own.
- **Freeze risk + nowhere to validate.** A bad `verify` makes the bootstrap source `NotReady` → every Kustomization stalls (a GitOps freeze the cluster can't self-heal from). And it's prod-only: local/CI artifacts are pushed **unsigned**, so there's no safe cluster to test on first.

So shipping a live override would be a prod-freeze landmine. The responsible spike output is the exact config + a safe path.

## What the doc captures

- The precise `OCIRepository.spec.verify` block, incl. the `matchOIDCIdentity` regexes for **both** `cd.yaml` (tags) and `ci.yaml` (merge-queue) — tight on repo+workflow, broad on git ref.
- The blocker analysis above.
- **Recommended fix:** a small KSail feature to pass `verify` through to the `OCIRepository` it generates (KSail owns the resource, so it should own its verify config). No `devantler-tech/ksail` issue exists yet — filing it is the next action (happy to open it on request).
- A supervised interim in-repo-override rollout + rollback plan, if verification is wanted before KSail ships the feature.
- The cleanup: drop the kubescape `C-0237` exception once enforced.

## Depends on

[#1694](https://github.com/devantler-tech/platform/pull/1694) — both deploy paths must sign before `matchOIDCIdentity` can require the `(cd|ci)` identities, else merge-queue deploys would freeze the cluster.

## Validation

Docs-only; no kustomization references `docs/`, so both overlays are unaffected. Relative links checked against the tree.

### DIFF
--- a/docs/supply-chain-verification.md
+++ b/docs/supply-chain-verification.md
@@ -1,0 +1,158 @@
+# Supply-chain verification (platform OCI artifact)
+
+This documents the **consumer-side verification** half of the platform's
+supply-chain story: making the cluster *reject* a platform OCI artifact that
+was not signed by our own CI. It is the planned follow-up that
+[#1628](https://github.com/devantler-tech/platform/pull/1628) explicitly left
+out of scope, and the reason the kubescape `C-0237` control is still excepted in
+[`security-exceptions/image-verification.yaml`](../k8s/bases/infrastructure/security-exceptions/image-verification.yaml).
+
+> **TL;DR.** Signing is done; enforcement is **blocked on KSail**, which owns the
+> root `flux-system` `OCIRepository` and doesn't yet expose Flux's `spec.verify`.
+> The clean fix is a small KSail feature; the interim in-repo override is
+> possible but carries a real prod-GitOps-freeze risk and can't be validated
+> anywhere but prod. This doc records the exact config and a safe path so it can
+> be executed deliberately rather than rediscovered.
+
+---
+
+## Where we are
+
+| Stage | Status |
+| --- | --- |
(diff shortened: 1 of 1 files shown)


devantler:
Superseded by an upstream feature request — **devantler-tech/ksail#4987** — per the decision to file this as a KSail issue rather than carry the plan as a platform doc.

The verify-enforcement work is blocked on KSail exposing Flux's `spec.verify` on the `OCIRepository` it generates (the resource is KSail-owned; a hand-written override risks a prod GitOps freeze). Once that ships, wiring it here is a small follow-up: set the `matchOIDCIdentity` regexes in `ksail.prod.yaml` and drop the kubescape `C-0237` exception. Closing this doc PR; the analysis now lives in the issue.

PR #1704, opened 2026-06-02
fix(cluster-policies): vendor patched policies the sync deletes

devantler:
## Problem

The **🧪 System Test** / **CI - Required Checks** on the daily sync PR [#1589](https://github.com/devantler-tech/platform/pull/1589) fail at the `ksail workload validate` step:

```
✗ bases/infrastructure/cluster-policies failed
accumulating resources from 'samples/best-practices/add-ns-quota/add-ns-quota.yaml':
  ... no such file or directory
✔ 72 kustomizations validated  ✗ 4 failed
```

(cascades to `bases/infrastructure`, `providers/docker/infrastructure`, `providers/hetzner/infrastructure`.)

## Root cause

`cluster-policies/kustomization.yaml` references three upstream kyverno policies under `samples/` — `add-ns-quota`, `create-pod-antiaffinity`, `spread-pods-across-topology` — that the **Sync Cluster Policies** workflow *deletes*.

`.policyignore` tries to whitelist exactly those three with gitignore-style `!` negation:

```
best-practices*
!best-practices/add-ns-quota      # ← intended whitelist
other*
!other/create-pod-antiaffinity    # ← intended whitelist
!other/spread-pods-across-topology # ← intended whitelist
```

But the reusable workflow (`devantler-tech/reusable-workflows`) applies it with a plain loop that **does not understand `!` negation**:

```bash
while IFS= read -r pattern; do
  find "$TMP" -path "$TMP/$pattern" -exec rm -rf {} +
done < .policyignore
```

`find -path "$TMP/!best-practices/add-ns-quota"` matches nothing (there is no literal `!` in any path), so each `!` line is a silent no-op. The broad `best-practices*` / `other*` lines then delete the whole subtrees — including the three policies the build references. Result: dangling `resources:` → `kustomize build` fails.

I verified the filter logic is **identical across every pinned version (v3.2.0 → v5.4.1)**, so this is a longstanding latent bug, not a recent regression. All three policies still exist upstream at their original paths — they are not being moved or removed upstream; the sync is simply wiping them locally.

These three policies are also heavily customized here with **positional JSON6902 patches** (`/spec/rules/0/...`, deep `patchStrategicMerge` paths). Blind upstream auto-sync would silently break those patches on any rule reorder — so auto-syncing them was never actually safe.

## Fix

Treat the three policies as what they already are in practice — **vendored, pinned, platform-owned** — and decouple them from the destructive sync:

- Move them out of the sync-managed `samples/` into an owned `vendored/` directory (siblings of the existing owned `best-practices/` and `flux/` policy sets).
- Repoint the three `resources:` entries in `kustomization.yaml`; all patches are unchanged (they target by `kind`+`name`, not path).
- Drop the three dead `!` lines from `.policyignore`.

After this lands, the daily sync's filtered output (`samples/_step-templates/` only) matches `main`, so the sync produces an empty diff and [#1589](https://github.com/devantler-tech/platform/pull/1589) auto-closes — the build can no longer be broken by the sync.

## Validation

- `ksail workload validate` (local) — **264 files validated, exit 0**
- `ksail --config ksail.prod.yaml workload validate` — **264 files validated, exit 0**
- `kubectl kustomize` of `clusters/local` and `clusters/prod` both build; all three policies render with their patches applied (`DoNotSchedule`, terminating-namespace precondition, `StatefulSet` match).
- **Rendered prod manifests are byte-identical to `origin/main`** — a pure relocation with zero behavioral change to the cluster.

## Recommended follow-up (separate repo)

The underlying bug lives in `devantler-tech/reusable-workflows` `sync-cluster-policies.yaml`: the `find -path` filter should honor gitignore-style `!` negation (e.g. process includes/excludes in order, or use `rsync --include/--exclude`). This PR makes the platform robust regardless, but fixing it upstream would restore working whitelists for any other consumer.


devantler:
Superseding this in favor of fixing the root cause upstream, per maintainer direction.

The real bug is in the sync workflow itself: `devantler-tech/reusable-workflows`'s `sync-cluster-policies.yaml` filtered `.policyignore` with a `find -path … -exec rm -rf` loop that has no notion of gitignore-style `!` negation, so the whitelist lines were silent no-ops and the three referenced policies got deleted every run. Fixed there in devantler-tech/reusable-workflows#277 (per-file, last-match-wins, honors `!`) — so no vendoring is needed and the whitelist works for every consumer.

Closing this. Platform follow-up once #277 releases:
1. Bump the `sync-cluster-policies.yaml` workflow pin to the released version.
2. Switch the three `.policyignore` negations from directory-level to **file-scoped** (e.g. `!best-practices/add-ns-quota/add-ns-quota.yaml`) so the synced output stays clean — with working negation, a bare `!dir` now re-includes that dir's `artifacthub-pkg.yml` / `.chainsaw-test/` / `.kyverno-test/` siblings too. Verified end-to-end that file-scoped negations reproduce exactly today's `samples/` contents, so the sync will then be a zero-diff no-op.

PR #1745, opened 2026-06-04
fix(cluster-policies): exempt monitoring from the default ResourceQuota

devantler:
## Symptom

Live prod warning events, all cascading from one cause:

```
monitoring   FailedCreate     daemonset/alloy ... forbidden: exceeded quota: default-resourcequota
monitoring   FailedCreate     daemonset/kube-prometheus-stack-prometheus-node-exporter ... exceeded quota
monitoring   FailedCreate     job/kube-prometheus-stack-admission-create ... exceeded quota (158x)
monitoring   RollbackFailed   helmrelease/kube-prometheus-stack
flux-system  HealthCheckFailed kustomization/infrastructure-controllers
```

The autoscaler node `autoscale-cx23` had **no** `node-exporter`/`alloy` pod
(every other node did): `alloy` DaemonSet `DESIRED 4 / CURRENT 3`,
`node-exporter` `DESIRED 7 / CURRENT 6`.

## Root cause

The Kyverno `add-ns-quota` policy generates a hard `requests.memory=16Gi`
ResourceQuota into every namespace **except** the platform/system ones
(`kube-system`, `kube-public`, `kube-node-lease`, `flux-system`). The
`monitoring` namespace was wrongly treated as a tenant, but it:

- runs the `node-exporter` and `alloy` **DaemonSets**, which must place a pod on
  **every** node — including autoscaler-provisioned ones. DaemonSet pods get no
  exemption from ResourceQuota admission, so a full namespace quota blocks them
  cluster-wide on scale-up;
- carries a **VPA-governed** observability footprint that legitimately exceeds
  the generic tenant cap.

When the cluster autoscaled a new node, the quota rejected the DaemonSet pods
**and** the kube-prometheus-stack Helm pre-upgrade hook Job for it. That stalled
the HelmRelease (`UpgradeFailed` → `RollbackFailed`, because the rollback waits
on the node-exporter DaemonSet) and the `infrastructure-controllers` Flux
Kustomization (`HealthCheckFailed`).

The fault was **amplified** by the quota controller's `used` value wedging at
`17124312638` (15.95/16Gi) while the real sum of all 19 running pods'
`requests.memory` was only **4.78 GiB** — i.e. the namespace was nowhere near its
true footprint, yet admission was blocked.

## Fix

Add `monitoring` to the `generate-resourcequota` exclude list, exempting it like
the other platform-owned namespaces. Its sizing is VPA-owned, not quota-bounded.

- Surgical: only the **ResourceQuota** (rule 0) is changed. The **LimitRange**
  (rule 1) still applies, so default container requests/limits are unchanged.
- With `generateExisting: true` + `synchronize: true`, Kyverno should garbage-
  collect the now-orphaned `default-resourcequota` in `monitoring`, which also
  clears the wedged `used` accounting and unblocks the DaemonSets + Helm hook.

> **Operational note (post-merge):** if Kyverno does not auto-remove the existing
> object, a one-time `kubectl delete resourcequota -n monitoring default-resourcequota`
> resets the stale accounting and lets Flux finish reconciling.

## Validation

- `kubectl kustomize k8s/clusters/{local,prod}/` — both build.
- `ksail workload validate` (local) and `ksail --config ksail.prod.yaml workload validate` — **279 files validated** each.
- Rendered policy confirms `monitoring` exempt from rule 0 only; rule 1 unchanged.

### DIFF
--- a/k8s/bases/infrastructure/cluster-policies/kustomization.yaml
+++ b/k8s/bases/infrastructure/cluster-policies/kustomization.yaml
@@ -35,6 +35,18 @@ patches:
                   - kube-public
                   - kube-node-lease
                   - flux-system
+                  # Platform observability stack. monitoring runs the
+                  # node-exporter and alloy DaemonSets, which must schedule a pod
+                  # on EVERY node -- including autoscaler-provisioned ones -- and
+                  # carries a VPA-governed footprint that legitimately exceeds the
+                  # generic tenant quota. A hard requests.memory ResourceQuota
+                  # here rejects those DaemonSet pods (and the kube-prometheus-stack
+                  # Helm pre-upgrade hook Job) on any new node, which stalls the
+                  # HelmRelease and the infrastructure-controllers Flux
+                  # Kustomization cluster-wide. Exempt it like the other
+                  # platform-owned namespaces; its sizing is VPA-owned, not
+                  # quota-bounded. The LimitRange (rule 1) still applies.
+                  - monitoring
       - op: add
         path: /spec/rules/0/generate/generateExisting
         value: true


devantler:
As we are currently switching to Coroot, I am not sure we need this. We need to verify this, and if I am right, this PR should be closed.

devantler:

**Verification: the exemption is still needed — recommend keeping (your call on close).**

I checked the live state against the Coroot switch:

1. **Coroot has not landed yet.** No `coroot` manifests exist on `main`, and the kube-prometheus-stack is still actively maintained — `v86.0.1` merged in #1741 the same day as your comment. The `monitoring` namespace with the `node-exporter` and `alloy` **DaemonSets** is live right now.
2. **The bug is still unfixed on `main`.** The `add-ns-quota` exclude patch in `k8s/bases/infrastructure/cluster-policies/kustomization.yaml` does not include `monitoring`, so the prod cascade this PR describes (DaemonSet `FailedCreate` on autoscale → `kube-prometheus-stack` `RollbackFailed` → `infrastructure-controllers` `HealthCheckFailed`) is current, not historical.
3. **Coroot does not remove the need — it relocates it.** Coroot CE deploys `coroot-node-agent` as a **DaemonSet** (eBPF, one pod per node, autoscaler nodes included). A hard `requests.memory` ResourceQuota in Coroot's namespace would reject those pods on scale-up exactly as it does for `node-exporter`/`alloy` today. So the underlying rule — *an observability namespace that runs a DaemonSet must be quota-exempt* — survives the migration; only the namespace name changes (`monitoring` → whatever Coroot lands in).

**Recommendation:** keep this and merge it to clear the live prod autoscale break now; at the Coroot cutover, **repoint** the exclude entry from `monitoring` to Coroot's namespace (or add it) rather than dropping the exemption. If you would still rather close — e.g. you intend to tear down `monitoring` imminently and tolerate the cascade until then — that's your call and I'll close on your word; I'm leaving it open since it fixes a live prod issue. CI is green (6 checks SUCCESS; only the infra-gated System Test pending).

PR #1755, opened 2026-06-05
fix(coroot): apply the Coroot CR in the infrastructure layer to break a CRD-ordering deadlock

devantler:

## Symptom

On the prod cluster, the whole Flux dependency chain is wedged and the replaced
`monitoring/kube-prometheus-stack` (plus `loki`/`alloy`) never clears:

```
infrastructure-controllers   False   Coroot/coroot/coroot dry-run failed: no matches for kind "Coroot" in version "coroot.com/v1"
infrastructure               False   dependency 'flux-system/infrastructure-controllers' is not ready
apps                         False   dependency 'flux-system/infrastructure' is not ready
```

## Root cause — a CRD-ordering deadlock introduced in #1742

The `coroot-operator` HelmRelease (which **ships** the `coroot.com/v1` CRD) and
the `Coroot` **custom resource** were placed in the **same** Flux Kustomization
(`infrastructure-controllers`). `kustomize-controller` server-side dry-runs the
entire resource set in one shot before applying. The `Coroot` CR fails that
dry-run because its CRD does not exist yet, so **the whole Kustomization aborts
— including the operator HelmRelease that would have installed the CRD.** The
CRD never appears, the dry-run never passes: a permanent deadlock.

Live confirmation: the `coroot` namespace exists, but there is **no**
coroot-operator HelmRelease and **no** `coroot.com/v1` CRD on the cluster.

Because `infrastructure-controllers` can never go Ready, `infrastructure` and
`apps` (which `dependsOn` it, `wait: true`) are frozen too. The
kube-prometheus-stack / loki / alloy HelmReleases that #1742 removed from the
repo all lived under `controllers/`, so they are pinned in that frozen
Kustomization's inventory and **never get pruned** — exactly the "monitoring not
clearing" symptom. (The stuck `kube-prometheus-stack` rollback — node-exporter
`DaemonSet` blocked by the near-full `monitoring` ResourceQuota — is downstream
of the same freeze: #1745's quota exemption lives in `infrastructure` and can't
apply either.)

## The fix

Split the operator (CRD provider) from the CR across the existing Flux layers —
**no new layer** (matching the prior `bootstrap`/PriorityClass decision):

- **`infrastructure-controllers`** keeps the `coroot-operator` HelmRelease, the
  namespace, network policy, SSO `HTTPRoute` and heartbeat `CronJob`. None of
  these need the `Coroot` CRD.
- **`infrastructure`** (which `dependsOn` + `wait: true`s for
  `infrastructure-controllers`) now owns the `Coroot` CR. By the time it runs,
  the operator is installed and the CRD is established, so the CR applies
  cleanly. The prod-only hcloud-storage/Slack patch moves with it into the
  hetzner `infrastructure` overlay.

Once merged and deployed, `infrastructure-controllers` reconciles, installs the
operator **and prunes the stale kube-prometheus-stack/loki/alloy releases**,
then `infrastructure` applies the `Coroot` CR → the chain unblocks and
`apps` follows.

## Validation (`kubectl kustomize`)

| Build | Result |
|---|---|
| `clusters/local` + `clusters/prod` overlays | build OK |
| `providers/hetzner/infrastructure` | exactly **1** `Coroot` CR, hetzner patch applied (`className: hcloud`, Slack webhook, `retention: 14d`) |
| `providers/hetzner/infrastructure/controllers` | **0** `Coroot` CRs, `coroot-operator` present |
| `providers/docker/infrastructure` | exactly **1** `Coroot` CR, base values (no hcloud) |
| `providers/docker/infrastructure/controllers` | **0** `Coroot` CRs, `coroot-operator` present |

## Notes

- **CI System Test** validation (`ksail workload validate`) is a *separate*
  concern — its client-side schema check still trips on the unknown `Coroot`
  kind, which is what #1753's `skipKinds: [Coroot]` addresses. This PR fixes the
  **live reconcile** deadlock; the two are complementary.
- **Post-deploy watch-points:** the stuck `kube-prometheus-stack` HelmRelease
  should `helm uninstall` cleanly on prune; if the `monitoring` namespace hangs
  in `Terminating`, check for leftover PVCs/finalizers.


devantler:

**CI triage — the System Test red is the known Coroot schema gate, not a defect in this change.**

The only failing kustomization is `providers/hetzner/infrastructure`:

```
✗ providers/hetzner/infrastructure: build failed: validate manifests:
  jsonschema validation failed with '.../CRDs-catalog/.../coroot.com/coroot_v1.json#'
  - at '/spec/projects/0': additional properties 'notificationIntegrations' not allowed
```

- The **base** Coroot CR validates fine (`✔ bases/infrastructure/coroot validated`). Only the **prod** build fails, because the hetzner patch adds `spec.projects[].notificationIntegrations` (the Slack webhook wiring from #1742) and the datreeio CRDs-catalog schema for `coroot.com/v1` is stale (`additionalProperties: false`, missing newer fields).
- This is the **same** validation failure #1742 was merged RED with — this PR only **relocates** it from the `providers/hetzner/infrastructure/controllers` group to `providers/hetzner/infrastructure` (still exactly `✗ 1 failed`, `✔ 79 validated`). No new validation surface, no regression.
- The designated fix is **#1753** (`skipKinds: [Coroot]`), which is kind-scoped, so it covers the CR wherever it lives. But #1753 requires **ksail ≥ 7.26.0** (`--skip-kinds` / `spec.workload.validation.skipKinds`), and 7.26.0 is not published yet (latest is v7.25.0) — so **#1753's own System Test is red too**. Both PRs are blocked from green System Test by the same external gate.

**There is no in-repo code change that makes this green now** — the field is a real prod feature, and the schema is upstream (datreeio/CRDs-catalog).

**Unblock paths:**
1. **Immediate prod recovery:** add this PR to the merge queue — the `merge_group` event runs *Deploy to Prod* (which does not run `workload validate`), so it pushes the reconcile fix and Flux self-heals, exactly as the queue did for #1742.
2. **Clean green CI:** once ksail 7.26.0 publishes → merge #1753 (skipKinds) → rebase this PR → System Test skips the Coroot kind → green → merge.

The reconcile fix itself is validated: all 4 provider builds + both cluster overlays build; `providers/hetzner/infrastructure` has exactly 1 Coroot CR (hetzner-patched), `.../controllers` has 0.

Issue #1807, opened 2026-06-05
k8s 1.36 readiness: clean on removals; pilot user namespaces, watch VolumeGroupSnapshot


devantler:
> 🤖 Generated by Claude Code at the maintainer's request (interactive session).

## k8s 1.36 ("Haru", GA ~22 Apr 2026) — platform readiness & follow-ups

A review of Kubernetes 1.36 against this platform. **Headline: nothing in 1.36 forces a change — the platform is clean against every removal and deprecation.** The version bump itself is now *possible* (Talos 1.13.3 raised the ceiling) but not urgent, and a few GA features are worth adopting opportunistically. This issue tracks the follow-ups.

### ✅ Clean bill of health — 1.36 removals/deprecations that do **not** affect us

| 1.36 change | Why we're unaffected |
| --- | --- |
| **IPVS kube-proxy deprecated** | Cilium runs `kubeProxyReplacement: true` (`k8s/bases/infrastructure/controllers/cilium/helm-release.yaml`) — no kube-proxy, no IPVS at all |
| **`gitRepo` volume removed** | Not used anywhere (GitOps via Flux) |
| **Service `externalIPs` deprecated** (removal ~1.43) | Not used — ingress is the Cilium Gateway → Hetzner LB |
| **Legacy AppArmor annotations removed** | `talos/cluster/apparmor.yaml` is a Talos kernel LSM arg, not the removed `container.apparmor.security.beta.kubernetes.io/*` pod annotation |
| **SELinux volume labeling GA** (breaking in 1.37) | Talos uses AppArmor, not SELinux |
| **Endpoints API deprecation** | No raw `kind: Endpoints`; EndpointSlice via Cilium |

### 🔼 The version bump (k8s 1.35.5 → 1.36) — optional, not urgent

- **Unlocked:** Talos 1.13.3 (already pinned, `ksail.prod.yaml`) supports k8s 1.31–1.36. The lockstep comment in `ksail.prod.yaml` makes "bump `kubernetesVersion` when Talos's range moves" the maintainer's manual step (protected file).
- **No forcing function:** the Talos [#13350](https://github.com/siderolabs/talos/issues/13350) scheduler-config bug (1.36 plugin fields leaking into a 1.35 kube-scheduler config) was a **1.13.2 regression fixed in 1.13.3**, which we already pin — so 1.35.5-on-1.13.3 is healthy.
[...]

PR #1808, opened 2026-06-05
feat(whoami): pilot user namespaces (hostUsers: false) in prod

devantler:
Part of #1807 (k8s 1.36 readiness).

## What

Run the stateless **whoami** pod inside a private user namespace by setting `hostUsers: false` on its pod spec — the first, canary step of adopting user namespaces across the platform.

User Namespaces (KEP-127) **graduated to GA in Kubernetes 1.36** and has been beta/on-by-default since 1.30, so this already takes effect on the cluster's current **v1.35.5** — no version bump required. The container's UID/GID (65534) is mapped to an unprivileged subordinate range on the host, so a container-runtime escape lands as a **non-root, capability-less host user** — defence in depth on top of the restricted `securityContext` already set in the base HelmRelease.

## Why whoami, why prod-only

- **whoami is the ideal canary:** stateless, **no volumes**, and `automountServiceAccountToken` is already `false` — so there are none of the idmap-volume constraints that make stateful workloads (Longhorn/CNPG) the *last* thing to migrate.
- **Scoped to the hetzner (prod) overlay** because prod nodes run Talos on real Hetzner kernels where user namespaces + idmap mounts are supported, whereas the local Docker provider runs Talos *nested in Docker* where userns support is not guaranteed. This matches the repo's convention for environment-capability-dependent settings (cf. the prod-only `require-mutual-auth` policy).

## How

An RFC 6902 `add` that **appends** (`/-`) to whoami's existing `postRenderers[0].kustomize.patches` list, so the base's strategy / ServiceAccount post-render patches are **preserved** rather than replaced (a strategic-merge patch on the list would replace it). The chart (`cowboysysop/whoami` 6.0.0) doesn't expose `hostUsers` in values, so a post-render patch is the mechanism.

## Validation (static only — no cluster)

- `kubectl kustomize` builds clean for `providers/hetzner/apps`, `clusters/prod`, and `clusters/local`.
- The rendered whoami HelmRelease carries **all three** post-render patches, with `hostUsers: false` (boolean) on the Deployment pod template.
- `providers/docker/apps` renders **0 resources** → the pilot is genuinely prod-only.
- `ksail --config ksail.prod.yaml workload validate` → **`providers/hetzner/apps validated ✔`**. The only failure is the pre-existing stale Coroot CRD-catalog schema (`notificationIntegrations`) in the *infrastructure* layer, unrelated to this change.

## How to verify once deployed to prod

```
kubectl -n whoami get pods                 # whoami pods Running, not CrashLoop
kubectl -n whoami exec deploy/whoami -- id # UID 65534 inside; remapped on host
```
Watch for: pod fails to start / `unable to set up user namespace` events. Rollback = revert this PR (removes the patch; whoami returns to host userns).

## Next (tracked in #1807)

homepage (apex; has read-only configMap mounts to verify under userns) → then stateful workloads, last.


devantler:
> 🤖 Analysis by Claude Code (at the maintainer's request)

**The System Test failure is the known infra-only reconcile flake — not caused by this PR.** No code change is warranted; a re-run should clear it.

The cluster came up healthy (etcd, CNI, cert-manager, CSI, metrics-server, policy-engine, Flux all `installed`), then `infrastructure`, `infrastructure-controllers`, and `apps` failed to reconcile within the timeout with transient infrastructure signatures:

- `client rate limiter Wait returned an error: context deadline exceeded` — API/etcd overload on the single-control-plane Docker runner
- `minio-create-bucket … dial tcp …:9000: connect: no route to host` — Cilium networking-convergence race
- `helper-pod-create-pvc-… not found in namespace "local-path-storage"` — local-path-storage race

### Why it isn't this PR

1. **The identical code passed System Test yesterday** — run on this branch at `2026-06-05T22:08` (commit `a89cf446`) was green. The only delta since is `dbd74219`, a **one-line YAML comment trim** — causally incapable of breaking a cluster reconcile.
2. The failing Kustomizations are `infrastructure` / `infrastructure-controllers`, which this PR **does not touch**. `apps` fails only as a cascade (it `dependsOn` infrastructure).
3. This change is **prod-only** (hetzner overlay); the local Docker `apps` overlay renders **0 resources**, so `whoami` / `hostUsers` is never deployed in the System Test at all.
4. The **same flake is hitting unrelated runs right now** — the `main` merge queue failed repeatedly today (02:10, 02:11, 06:31, 07:49) while other PRs intermittently pass.

### Recommendation

Re-run the **🧪 System Test** job (or let the merge queue retry). This matches the documented infra-only-flake handling. The flake recurrence across the merge queue suggests CI-runner capacity / the single-CP etcd-overload issue is acting up today and may warrant separate attention.

devantler:
> 🤖 Update

Full root-cause of the stuck merge queue is now tracked in #1816 — a prod wedge (spire-server's Longhorn volume stranded on a deleted autoscaler node → SPIRE down → ExternalSecrets cascade), **not** this PR. This PR stays BLOCKED until the queue drains (see #1816 §1 for the un-wedge steps); no change needed here.

PR #1844, opened 2026-06-07
fix(flux): enable HelmRelease drift detection platform-wide

devantler:
## Problem

Prod looked healthy at the Flux layer (every Kustomization/HelmRelease `Ready=True`), but two Helm-rendered Deployments had been **deleted out-of-band and never restored**:

- `keda-add-ons-http-external-scaler` (keda ns) → `external-scaler` Service had **no endpoints**, every HTTP `ScaledObject` was `READY=False` (`got empty metric spec`), and HTTP scale-from-zero was dead for `actual-budget` / `headlamp` / `hubble-ui` / `whoami`. The `hubble-ui` HPA `FailedGetResourceMetric` warning was downstream of this.
- `homepage` (Flagger source Deployment) → Flagger logged `deployment "homepage" not found` every ~44s.

**Root cause:** Flux's helm-controller only re-applies a chart's rendered resources on a chart/values change. With drift detection disabled (the default, and **none** of the 33 HelmReleases had it set), a deleted/mutated Helm-rendered resource is never restored — the release keeps reporting `install succeeded` while the workload is silently broken.

## Fix

Enable `spec.driftDetection.mode: enabled` on **every** HelmRelease via a shared Kustomize Component (`k8s/bases/components/helmrelease-drift-detection`), referenced from each provider build root that renders HelmReleases (hetzner + docker × controllers/apps/infrastructure).

Field-level `ignore` rules exclude values that **other controllers legitimately own at runtime**, so Flux won't fight them in a hot loop:

| Ignored path | Owner |
|---|---|
| `/spec/replicas` (Deployment/StatefulSet) | HPA / KEDA / Flagger / cluster-autoscaler |
| Validating/Mutating webhook `caBundle` | cert-manager injection + operators (cnpg, external-secrets, keda, kubescape, vpa) |
| `APIService .spec.caBundle` | aggregated-apiserver cert injection |
| CRD conversion `.caBundle` | cert-manager CA injection |

These are **field-level** ignores only — a wholesale DELETE of any such resource is still drift and is restored.

## Live remediation already applied

The two deleted Deployments were restored live (`flux reconcile hr … --force`) ahead of this PR, so prod is already healthy:
- KEDA: all 5 ScaledObjects now `READY=True`, scaler has endpoints, interceptor active, hubble-ui HPA back to its External metric.
- homepage: source Deployment restored, `homepage-primary` healthy, canary intact.

This PR makes that **self-healing and permanent**.

## Validation

- `kubectl kustomize` builds clean for every affected provider path; rendered HelmReleases = `driftDetection` blocks injected (27/27 hetzner controllers, 5/5 apps, 1/1 infra-opencost, 21/21 docker controllers, 1/1 docker infra).
- `ksail --config ksail.prod.yaml workload validate`: **86/86 groups pass**, including every HelmRelease group. The one unrelated failure (`providers/hetzner/infrastructure`) is the pre-existing Coroot CR `notificationIntegrations` CRDs-catalog false-positive — present on `main`, untouched here (this change only patches HelmReleases). CI's docker System Test uses `ksail.yaml` (`skipKinds: [Coroot]`) and does not hit it.

## Watch on first reconcile

After merge, watch the first helm-controller reconcile of the webhook/cert-managing HRs (cert-manager, trust-manager, cloudnative-pg, external-secrets, keda, kubescape, vertical-pod-autoscaler) for any unexpected drift correction; the `caBundle`/`webhooks` ignores are designed to prevent flapping, and the docker System Test exercises them.

### DIFF
--- a/k8s/bases/components/helmrelease-drift-detection/kustomization.yaml
+++ b/k8s/bases/components/helmrelease-drift-detection/kustomization.yaml
@@ -1,0 +1,67 @@
+---
+# Platform-wide HelmRelease drift detection.
+#
+# WHY: Flux's helm-controller only re-applies a chart's rendered resources on a
+# chart/values change (a Helm upgrade). If a Helm-rendered resource is deleted or
+# mutated out-of-band, nothing restores it — the HelmRelease still reports
+# "install succeeded" while the workload is silently broken. This happened twice:
+# the keda-add-ons-http `external-scaler` Deployment and the homepage (Flagger
+# source) Deployment were both deleted and never came back, which silently broke
+# all HTTP autoscaling and spammed Flagger "deployment not found" warnings for
+# days while Flux still reported everything Ready.
+#
+# WHAT: turn on `driftDetection.mode: enabled` for every HelmRelease so the
+# helm-controller continuously reconciles the cluster back to the rendered
+# release — recreating deleted resources and reverting out-of-band edits.
+#
+# IGNORES: a few fields are legitimately owned by OTHER controllers at runtime;
+# enforcing the chart's value would make Flux fight them in a hot loop:
+#   * /spec/replicas (Deployment/StatefulSet) — HPA / KEDA / Flagger / autoscaler
+#   * Validating/Mutating webhook `.webhooks[].clientConfig.caBundle` — injected
+#     by cert-manager and patched at runtime by operators (cloudnative-pg,
+#     external-secrets, keda, kubescape, vertical-pod-autoscaler)
+#   * APIService `.spec.caBundle` — aggregated-apiserver cert injection
+#   * CRD conversion `.caBundle` — cert-manager CA injection
+# These are field-level ignores only: a wholesale DELETE of any such resource is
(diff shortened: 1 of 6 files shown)


devantler:
**CI note:** the red 🧪 System Test is the pre-existing `pull-registry-image` local-action infra break in the ksail-cluster action (`Can't find 'action.yml' ... under .github/actions/pull-registry-image`) — it fails on every PR right now and is unrelated to this change (which only adds `driftDetection` to HelmReleases). Validated instead with:

- `ksail --config ksail.prod.yaml workload validate` → all 86 HelmRelease groups pass (the one unrelated failure is the known Coroot `notificationIntegrations` CRDs-catalog false-positive on `main`).
- `kubectl kustomize` builds clean for every affected provider path; `driftDetection` is injected on every rendered HelmRelease (27/27 hetzner controllers, 5/5 apps, 1/1 infra-opencost, 21/21 + 1/1 docker).

The merge queue skips the System Test, so this can still land once reviewed.

PR #1853, opened 2026-06-07
feat(ksail-operator): add operator with web UI + native Dex OIDC

devantler:
## What

Adds the **KSail operator** (`charts/ksail-operator`) as an infrastructure controller with the operator, REST API + embedded web UI, and **native (app-driven) OIDC against Dex**.

New component: `k8s/bases/infrastructure/controllers/ksail-operator/` (namespace, HelmRepository, HelmRelease, HTTPRoute, CiliumNetworkPolicy), wired into both the base controller list (local/CI) and the explicit hetzner list (prod). Consumes the published OCI chart `oci://ghcr.io/devantler-tech/charts/ksail-operator` **v7.36.0**.

## Feature config

| Feature | Setting | Notes |
|---|---|---|
| Operator + REST API + Web UI | on (`api.bindPort: 8080`) | UI embedded, same origin as API |
| `ui.readOnly` | **`true`** | GitOps-safe: Git/Flux stays source of truth; only writes locked |
| Native OIDC (Dex) | **prod-only** (`enabled: true` on hetzner) | see below |
| `operator.dockerSocket.enabled` | `false` | privileged; no `/var/run/docker.sock` on Talos → would crash-loop |
| `rbac.clusterAdmin` | `false` | least-privilege ClusterRole; no cluster-admin wildcard |

## Native OIDC is prod-only — and why

The operator does **eager, fatal OIDC discovery at manager startup**: the issuer must be resolvable + TLS-trusted *from inside the cluster*, or the manager fails to start and the pod crash-loops. The first full System Test run (which got past a transient kubescape flake and actually built the cluster) surfaced this on local:

```
run operator: start manager: set up OIDC authentication: discover OIDC provider:
Get "https://dex.platform.lan/.well-known/openid-configuration":
lookup dex.platform.lan on 10.96.0.10:53: no such host
```

On **local/CI**, `*.platform.lan` resolves only via the host's `/etc/hosts` (not in-cluster CoreDNS), so discovery fails → crash-loop → HelmRelease never Ready → System Test red. oauth2-proxy / Headlamp / Vault avoid this by discovering OIDC lazily or tolerating an unreachable Dex; the ksail operator is uniquely strict.

**Fix:** OIDC config lives in a **hetzner (prod) overlay patch**, not the base — matching the repo's "base = CI-safe minimal, hetzner patch adds prod-only config" pattern (same as Coroot's prod-only `notificationIntegrations`). On prod, `dex.platform.devantler.tech` resolves via public DNS and serves a trusted cert (the NetworkPolicy already allows egress `:443`), so OIDC works; local/CI runs the operator without OIDC (UI/API still served, read-only, host-only). Verified the kustomize merge: docker renders no `auth.oidc`; hetzner retains all base values and adds `auth.oidc.enabled: true`.

OIDC reuses the shared `public-client` confidential client + `${dex_client_secret}` (no new SOPS secret); callback `https://ksail.${domain}/api/v1/auth/callback` registered on it in the dex HelmRelease.

> **Upstream idea:** if the operator made OIDC discovery non-fatal / retrying at startup (like oauth2-proxy), OIDC could run on local too and the operator would be more robust to Dex bootstrap ordering on prod.

## Image-tag pin (intentional)

`operator.image.tag: v7.36.0` is pinned because the chart's `appVersion` is `7.36.0` (no `v`) while the released image is tagged **`v7.36.0`** — the no-`v` tag doesn't exist, so the chart default would `ImagePullBackOff`. Renovate tracks `chart.spec.version` (flux manager) but **not** this values image tag, so bump it in lockstep on chart bumps.

## Validation

- Rendered the published 7.36.0 chart with the PR values → image `…/ksail:v7.36.0`, `--read-only` + `--oidc-*` flags, Service `:8080`, OIDC Secret + scoped RBAC.
- `ksail workload validate` (local): ✔ **300 files**, exit 0.
- Prod validate: ksail-operator (base + hetzner OIDC patch) ✔; only failing group is the **pre-existing** Coroot stale-catalog false-positive (`coroot_v1.json … 'notificationIntegrations' not allowed`), unrelated.
- `readOnlyRootFilesystem` (Kyverno-enforced) did **not** break the operator — it started cleanly under the restricted PSS (the earlier-flagged risk did not materialize).

### DIFF
--- a/k8s/bases/infrastructure/controllers/ksail-operator/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/ksail-operator/helm-release.yaml
@@ -57,20 +57,15 @@ spec:
       # the API) so the Git repository stays the single source of truth. OIDC,
       # UI and API remain fully enabled -- only write actions are locked.
       readOnly: true
-    auth:
-      # Native, app-driven OIDC against Dex (Headlamp-style): the operator API
-      # owns the login/callback as a confidential client and establishes a
-      # session cookie. Reuses the shared `public-client` confidential client
-      # (same as oauth2-proxy / Vault / Headlamp) so no new SOPS secret is
-      # needed -- the redirect URI is registered on that client in the dex
-      # HelmRelease.
-      oidc:
-        enabled: true
-        issuerURL: https://dex.${domain}
-        clientID: public-client
-        clientSecret: ${dex_client_secret}
-        redirectURL: https://ksail.${domain}/api/v1/auth/callback
-        scopes: "openid email profile groups"
+    # Native OIDC is intentionally NOT enabled in the base -- it is layered on
+    # in the hetzner (prod) overlay (see ksail-operator/patches/). The operator
+    # performs *eager, fatal* OIDC discovery at startup: the issuer
+    # (https://dex.${domain}) must be resolvable AND TLS-trusted from inside the
+    # cluster, or the manager fails to start and the pod crash-loops. On
+    # local/CI the *.platform.lan names resolve only via the host's /etc/hosts
+    # (not in-cluster CoreDNS), so discovery would fail there. On prod,
+    # dex.platform.devantler.tech resolves via public DNS (netpol allows egress
+    # 443), so OIDC runs there. (auth.oidc.enabled defaults to false.)
     rbac:
       # Use the chart's least-privilege ClusterRole; clusterAdmin (a full
       # cluster-admin wildcard) stays off -- it is a privilege-escalation
--- a/k8s/bases/infrastructure/controllers/ksail-operator/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/ksail-operator/helm-release.yaml
@@ -37,13 +37,11 @@ spec:
     # default "<release>-ksail-operator").
     fullnameOverride: ksail-operator
     operator:
-      # The chart defaults the image tag to appVersion (7.36.0, no "v"), but the
-      # released ksail image is tagged "v7.36.0" -- the no-"v" tag does not
-      # exist, so the chart default would ImagePullBackOff. Pin the "v"-prefixed
-      # tag explicitly. Renovate tracks chart.spec.version (flux manager) but
-      # NOT this values image tag, so bump it in lockstep when the chart bumps.
-      image:
-        tag: v7.36.0
+      # Image tag is left to the chart default (v<appVersion>, per ksail #5168);
+      # chart.spec.version above selects the operator image version. Requires a
+      # chart release carrying #5168 AND a cosign-signed image (ksail #5165, for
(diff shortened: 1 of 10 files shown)


k8s/bases/infrastructure/controllers/ksail-operator/helm-release.yaml
```diff
@@ -0,0 +1,73 @@
...
+    # default "<release>-ksail-operator").
+    fullnameOverride: ksail-operator
+    operator:
+      # The chart defaults the image tag to appVersion (7.36.0, no "v"), but the
+      # released ksail image is tagged "v7.36.0" -- the no-"v" tag does not
+      # exist, so the chart default would ImagePullBackOff. Pin the "v"-prefixed
+      # tag explicitly. Renovate tracks chart.spec.version (flux manager) but
+      # NOT this values image tag, so bump it in lockstep when the chart bumps.
+      image:
+        tag: v7.36.0
```
devantler: Remove and rely on chart default.

devantler: Holding the pin for now — relying on the chart default is currently broken on two fronts (both verified against the registry):

1. **Wrong tag format.** The chart defaults the image to `ksail:{{ .Chart.AppVersion }}` (no `v`), but goreleaser only publishes `v`-prefixed tags (`v7.41.0`, …). Every no-`v` tag (`7.36.0`, `7.41.0`, …) 404s, so `image.tag: ""` → `ImagePullBackOff`.
2. **Unsigned image.** Even once the tag resolves, prod Talos `ImageVerificationConfig` (#1889) fails closed on the unsigned image. ksail #5165 (cosign signing) is merged, but every release through `v7.41.0` predates it and is still unsigned.

Fix for (1) is chart-side: **devantler-tech/ksail#5168** defaults the tag to `v<appVersion>`. Once a ksail release carrying that fix **and** a signed image (post-#5165) is published, I'll bump `chart.spec.version` here and drop this pin so the default resolves on its own. Keeping CI green until then.

devantler: Done — pin removed in 68a52140's follow-up (relying on the chart default now). Heads-up: the System Test will ImagePullBackOff until a ksail release carrying #5168 (v-prefixed default tag) + a signed image (#5165) is published and `chart.spec.version` here is bumped to it. Static validation is green.

devantler: Fully resolved & shipped in v1.42.0: pin removed and `chart.spec.version` bumped to 7.41.2 (first release carrying ksail#5168). Verified the chart default now resolves to `ghcr.io/devantler-tech/ksail:v7.41.2`, and `cosign verify` passes against the ksail CD identity, so prod Talos image verification (#1889) accepts it.

PR #1884, opened 2026-06-08
fix(secrets): seed R2 credentials from variables-cluster, not the stale base placeholder

devantler:
## Context — verifying the R2 fix on `main`

Following up on the R2 backup outage. The fix on `main` (`fcbf7db9`) rotated `r2_secret_access_key` in **`variables-cluster`**, but the live `BackupStorageLocation` is still `Unavailable` with the unchanged error:

```
api error InvalidArgument: Credential access key has length 27, should be 32
```

## Root cause — the PushSecret reads the wrong secret

There are two SOPS secrets in `flux-system`, both carrying R2 keys:

| Secret | `r2_access_key_id` length | Maintained? |
|--------|---------------------------|-------------|
| `variables-base` (`k8s/bases/bootstrap`) | **27 chars** (truncated placeholder) | stale — last touched by the `variables/→bootstrap/` rename (#1684) |
| `variables-cluster` (`k8s/clusters/<env>/bootstrap`) | **32 chars** (correct) | yes — holds the real per-env creds + the freshly-rotated secret |

`seed-r2-credentials` seeds OpenBao (`infrastructure/backup/r2`) from **`variables-base`** — so the truncated 27-char `access_key_id` is what flows to Velero (and CNPG), and that's the `length 27` the BSL rejects. The error is about the **access key ID**, which the `main` fix didn't touch; the correct one has been sitting in `variables-cluster` all along.

## Fix

Point the `seed-r2-credentials` PushSecret selector at **`variables-cluster`**, where the real per-environment R2 credentials live. `seed-cloudflare` still reads `variables-base` (`cloudflare_api_token` is a genuinely shared infra credential — unaffected).

Once deployed, ESO re-pushes the correct creds to OpenBao → `velero-r2-credentials` / `cnpg-r2-credentials` refresh → the BSL re-validates. This also **unblocks CNPG-native R2 backups** (separate PR), which read the same OpenBao path.

## Validation

- `kubectl kustomize k8s/clusters/{local,prod}/` — both build.
- `ksail --config ksail.prod.yaml workload validate` — `providers/hetzner/infrastructure/vault-seed validated ✔` (the one failure, `providers/hetzner/infrastructure`, is the pre-existing Coroot CRD catalog-schema issue, unrelated).
- Rendered selector confirmed: `seed-r2-credentials → variables-cluster`; `seed-cloudflare → variables-base` (unchanged).
- Both `clusters/local` and `clusters/prod` `variables-cluster` carry `r2_access_key_id` + `r2_secret_access_key`, so the repoint is safe for CI and prod.

## Follow-up

The now-unused R2 keys in `variables-base-secret.enc.yaml` are a harmless redundant copy; they can be dropped in a future re-encrypt (requires the Age key).

### DIFF
--- a/k8s/bases/apps/umami/db-backup-external-secret.yaml
+++ b/k8s/bases/apps/umami/db-backup-external-secret.yaml
@@ -18,6 +18,15 @@ spec:
   target:
     name: umami-db-backup-r2
     creationPolicy: Owner
+    template:
+      # REGION is a constant ("auto") rather than a fetched secret — CNPG's
+      # s3Credentials.region must point at a Secret key, and barman-cloud/boto3
+      # need a region for SigV4 signing. "auto" is R2's environment-independent
+      # convention (the same value Velero's BackupStorageLocation uses).
+      data:
+        ACCESS_KEY_ID: "{{ .ACCESS_KEY_ID }}"
+        SECRET_ACCESS_KEY: "{{ .SECRET_ACCESS_KEY }}"
+        REGION: "auto"
   data:
     - secretKey: ACCESS_KEY_ID
       remoteRef:
--- a/k8s/bases/apps/umami/db-backup-external-secret.yaml
+++ b/k8s/bases/apps/umami/db-backup-external-secret.yaml
@@ -1,0 +1,29 @@
+# R2 credentials for umami-db's CNPG barmanObjectStore backup (postgres-cluster.yaml).
+# Projects the shared backup R2 key/secret from OpenBao (infrastructure/backup/r2 —
+# the same path Velero's velero-r2-credentials reads) into a Secret in the umami
+# namespace, where CNPG's backup s3Credentials must live (it can only reference a
+# Secret in the Cluster's own namespace). Replaces the previously-orphaned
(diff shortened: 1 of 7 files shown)


devantler:
Follow-up from a code-review pass on the stacked backup work (#1885 was merged into this branch).

Added one commit — `fix(umami): set R2 region (auto) for CNPG barman S3 signing` (e7986af7): the `barmanObjectStore` `s3Credentials` was missing `region`. barman-cloud/boto3 need a region for SigV4 signing against R2, and CNPG's `region` field must reference a Secret key — so `REGION: auto` is added to the `umami-db-backup-r2` ExternalSecret (R2's environment-independent convention, matching Velero's working BSL) and wired via `s3Credentials.region`. Strictly safe: R2 accepts `auto` regardless; omitting it risked a runtime `NoRegionError` on the backup path.

Validated: `kubectl kustomize k8s/clusters/{local,prod}/` build; `ksail --config ksail.prod.yaml workload validate` → `bases/apps/umami ✔`, `providers/hetzner/apps ✔` (only the pre-existing Coroot CRD catalog failure remains, unrelated).

PR #1886, opened 2026-06-08
chore(config): remove dead Flux substitution variables and secret keys

devantler:
## What

Remove unused Flux post-build substitution **variables** and **secret keys** across the bootstrap config. Each was verified to have zero consumers — no `${name}` reference and (for secrets) no `vault-seed` PushSecret reading it.

### ConfigMap variables removed

| Variable | Defined in | Why dead |
|---|---|---|
| `homepage_replicas` | prod | homepage is a Flagger canary; chart `replicas` stripped by the postRenderer |
| `umami_replicas` | prod | same — umami is a Flagger canary (`replicas` stripped, `replicaCount` omitted) |
| `domain_regex` | local + prod | orphaned; no `${domain_regex}` consumer |
| `cloudflare_account_id` | base | no consumer; value is also hard-coded into `r2_endpoint` |
| `r2_prefix_cnpg` | base + local + prod | unreferenced sibling of the live `r2_prefix_velero` |

### SOPS secret keys removed (via `sops unset`; `.sops.yaml` + recipients unchanged, files re-verified to decrypt)

| Secret key | Defined in | Why dead |
|---|---|---|
| `cloudflared_tunnel_secret` | prod | no cloudflared/tunnel deployment exists |
| `fleetdm_mysql_password` | local + prod | superseded by an External Secrets `Password` generator → OpenBao `apps/fleetdm/mysql` |
| `fleetdm_mysql_root_password` | local + prod | same |
| `fleetdm_mysql_replication_password` | local + prod | same |
| `fleetdm_redis_password` | local + prod | generated → `apps/fleetdm/redis` |
| `fleetdm_server_private_key` | local + prod | generated → `apps/fleetdm/server`; the SOPS copy was never read, so removing it does not affect enrolled devices |

`fleetdm_license_key` **kept** (live — seeded to OpenBao by the `seed-fleetdm` PushSecret). [`fleetdm/README.md`](https://github.com/devantler-tech/platform/blob/main/k8s/bases/apps/fleetdm/README.md) updated to document the generated-secret flow and drop the obsolete `sops set` rotation commands.

## Method

Built the full set of *referenced* variable names repo-wide and subtracted it from the *defined* set; cross-checked secret keys against `vault-seed` PushSecret `secretKey`s (their non-`${}` consumption path) to avoid false positives. All live vars/keys (every other `*_replicas`, `r2_prefix_velero`, `cloudflare_api_token`, `hcloud_token`, the OIDC/alertmanager keys, etc.) were left untouched.

## Validation

- `kubectl kustomize k8s/clusters/local/` & `k8s/clusters/prod/` — build ✅
- `ksail workload validate` — 295 files validated ✅
- `ksail --config ksail.prod.yaml workload validate` — `clusters/prod/bootstrap validated` ✅; only failure is the **pre-existing, unrelated** Coroot CRDs-catalog `notificationIntegrations` schema error.
- Both edited `*.enc.yaml` re-verified: `sops -d` decrypts cleanly (MAC valid), only the intended keys removed.


devantler:
## Full dead-variable audit — items left for your call

This PR removes the 3 clearly-safe, editable **ConfigMap** variables. Below are the remaining candidates I found but did **not** touch, with reasoning. The secret ones live in protected `*.enc.yaml` and need `sops` re-encryption (I can't edit them).

### ConfigMap vars — dead now, but I'd hold off

| Variable | Defined in | Note |
|---|---|---|
| `r2_prefix_cnpg` | base + local + prod CM | Sibling of the live `r2_prefix_velero`, but currently unreferenced. It looks **provisioned ahead** of a consumer — and CNPG-native R2 backups are being built right now on the `claude/cnpg-native-r2-backups` branch. Removing it would likely just be re-added / cause a conflict. **Recommend: keep.** |
| `cloudflare_account_id` | base CM | No `${cloudflare_account_id}` consumer; the value is also hard-coded into `r2_endpoint`. Pairs with the unused `cloudflared_tunnel_secret` (no cloudflared/tunnel deployment exists), so likely an abandoned-tunnel remnant. **Safe to remove, but confirm there's no planned Cloudflare Tunnel.** |

### Secret keys — appear vestigial, but protected + sensitive (need `sops`)

These are keys in `variables-cluster`/`variables-base` **Secrets**. Their only legitimate consumption path is either `${...}` substitution or a `vault-seed` PushSecret `secretKey`. The following have **neither**:

| Secret key | Defined in | Why it looks dead |
|---|---|---|
| `cloudflared_tunnel_secret` | prod | No cloudflared deployment anywhere. |
| `fleetdm_mysql_password` | local + prod | Superseded by a **generated** secret — `generators.yaml` creates `fleetdm-mysql-password`, pushed to OpenBao `apps/fleetdm/mysql` by `push-generated-secrets.yaml`. |
| `fleetdm_mysql_root_password` | local + prod | Same — generated. |
| `fleetdm_mysql_replication_password` | local + prod | Same — generated. |
| `fleetdm_redis_password` | local + prod | Same — generated (`apps/fleetdm/redis`). |
| `fleetdm_server_private_key` | local + prod | Generated (`apps/fleetdm/server`). ⚠️ **Verify before removing** — `k8s/bases/apps/fleetdm/README.md` still documents this as the rotation source and warns it "must be stable forever — losing it invalidates every enrolled device". If the generated key is now authoritative, the README is stale; if Fleet was enrolled against the SOPS value, do not drop it. There's also active work on `feat/fleetdm-mysql-static-role-rotation`. |

If you confirm the generated-secret path is authoritative, these can be dropped from the SOPS secrets with `sops --set`/key deletion + re-encrypt, and the stale rows in `k8s/bases/apps/fleetdm/README.md` updated. I can prep that as a follow-up once you give the go-ahead (and since `*.enc.yaml` is protected, you'd run the re-encryption).

devantler:
Delete all of those. If not used now, they should be removed!

devantler:
Done — deleted all of them.

**ConfigMap vars:** `cloudflare_account_id`, `r2_prefix_cnpg` (in addition to the earlier `homepage_replicas` / `umami_replicas` / `domain_regex`).

**Secret keys** (`sops unset`, `.sops.yaml` + recipients unchanged, both files re-verified to decrypt cleanly):
- prod: `cloudflared_tunnel_secret` + the 5 `fleetdm_*` mysql/redis/server-key
- local: the same 5 `fleetdm_*` keys

`fleetdm_license_key` kept (live). I confirmed the removed fleetdm keys have **zero** runtime consumers — Fleet's real creds come from the generated secrets via OpenBao (`apps/fleetdm/{mysql,redis,server}`), so dropping the SOPS copies doesn't touch enrolled devices. README updated to match.

Note on `r2_prefix_cnpg`: removed per your call; if the in-flight CNPG-native R2 backup work needs a prefix var, it'll just re-add one.

The 3 new commits re-trigger CI, so the earlier `actual-budget` System Test flake re-runs automatically — no manual re-run needed.

PR #1902, opened 2026-06-08
fix(apps): run whoami and fleetdm always-on HA (KEDA min 2)

devantler:
Stacked on #1901. The KEDA scale-to-zero apps — made HA where the app actually supports multiple replicas.

## Always-on HA (min 2)
| App | Why it's safe |
|---|---|
| **whoami** | stateless echo |
| **fleetdm** | app tier is stateless (state in MySQL/Redis) |

`HTTPScaledObject` `replicas.min: 0 → 2`, initial replicas `1 → 2`. KEDA still owns the count; these stay exempt from the floor (the floor governs *statically-declared* replicas, not autoscaled ones).

## Genuinely single-pod (stay as-is, documented)
| App | Hard constraint |
|---|---|
| **hubble-ui** | per-pod connection state — streaming sessions bounce across pods |
| **headlamp** | OIDC login state is in-memory per pod |
| **actual-budget** | SQLite on an RWO PVC (single writer) |
| **longhorn-ui** | max-1 admin dashboard |

## ⚠️ Trade-off
This **abandons scale-to-zero** for whoami/fleetdm (always 2 pods, ~idle cost). The maintainer deliberately built scale-to-zero for these; revert per-app by setting the `HTTPScaledObject` `min` back to `0` if the idle cost outweighs the availability.

**Validation:** `ksail workload validate` passes for the changed files.

### DIFF
--- a/k8s/bases/apps/fleetdm/http-scaled-object.yaml
+++ b/k8s/bases/apps/fleetdm/http-scaled-object.yaml
@@ -12,10 +12,13 @@ spec:
     service: fleetdm-service
     port: 8080
   replicas:
-    min: 0
+    # Always-on HA: min 2 keeps the Fleet server replicated at all times
+    # (survives a pod/node failure) rather than scaling to zero. The app tier is
+    # stateless — state lives in MySQL/Redis.
+    min: 2
     max: 2
-  # Longer cooldown for FleetDM — enrolled devices check in periodically.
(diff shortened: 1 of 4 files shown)


devantler:
Closing — KEDA scale-to-zero apps should keep min 0 (they're already drain-safe). Dropped from the stack.

PR #1903, opened 2026-06-08
fix(apps): run homepage and umami Flagger primaries at 3 replicas for HA

devantler:
Stacked on #1902. The Flagger-managed apps — make their **primaries** run HA.

## Mechanism
Flagger copies the source Deployment's `spec.replicas` onto the `-primary`. The postRenderers previously **removed** `/spec/replicas` (so Flux wouldn't fight Flagger's between-rollout scale-to-zero of the source), which left each primary at the chart default of **1** — the maintainer documented this as "a minor HA reduction".

This **pins the source replicas to 3** (`op: replace` instead of `op: remove`), so Flagger runs the primary at 3. Safe because the platform's **Flux drift detection already ignores `/spec/replicas`** — Flagger's scale-to-zero of the source isn't corrected, which was the whole reason for the original strip.

| App | Change |
|---|---|
| **homepage** (apex) | primary 1 → 3 |
| **umami** (analytics) | primary 1 → 3 |
| **opencost** | left at 1 — cost *exporter*, singleton by nature (3 = duplicate compute/metrics), already observability-exempt |

## ⚠️ Watch
homepage is the platform front page (high blast radius). The change is to the replica-pinning only (probes/affinity/topology-spread untouched), but **watch the first Flagger rollout** after this lands.

**Validation:** `ksail workload validate` passes for the changed files. (Flagger's runtime replica copy is exercised in-cluster, not by `kubectl kustomize`.)

### DIFF
--- a/k8s/bases/apps/homepage/canary.yaml
+++ b/k8s/bases/apps/homepage/canary.yaml
@@ -6,10 +6,10 @@
 # new version. The public path (httproute.yaml -> oauth2-proxy -> auth-proxy ->
 # homepage apex Service) is untouched and always resolves to the primary.
 #
-# ⚠️ High blast radius (platform front page) — watch the first rollout. Flagger
-# owns replicas (stripped in the HelmRelease), so the primary runs the chart
-# default (1) instead of 2: a minor HA reduction. Latency threshold is generous
-# because Homepage is a single-threaded Next.js SSR app that bursts under load.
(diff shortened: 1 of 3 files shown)


devantler:
Superseded by a clean-branch re-open (the dropped KEDA commit is removed from history). See the new Flagger PR.

PR #1904, opened 2026-06-08
docs(dr): OpenBao single-node → 3-node Raft HA migration runbook

devantler:
Stacked on #1903. The final "everything HA" item — **OpenBao** — delivered as a runbook, because it's the one thing that can't be a blind manifest flip.

## Why this isn't a values-only PR
OpenBao runs **standalone + file storage** today. Raft is a different storage backend, so flipping the HelmRelease to `ha.raft` without a data migration brings up **empty** raft pods and **wipes every secret cluster-wide** (all ExternalSecrets/PushSecrets, the CNPG/FleetDM DB-engine roles). The cutover is a coordinated **offline migration on the secrets backbone** — not something Flux should auto-apply.

This PR is **doc-only** (safe to merge); the dangerous values flip stays manual, executed via the runbook when you're ready.

## What the runbook covers
- Exact 3-node Raft HA HelmRelease values (`standalone: false` + `ha.raft` + `retry_join` auto-join + re-declared audit device).
- The **`:8201` cluster-port** netpol + Talos-firewall prerequisite (the only piece safe to land ahead of time).
- Two migration paths: **snapshot + fresh-init + restore** (preferred on k8s, exercises DR) vs `bao operator migrate` (file→raft, preserves seal).
- Unseal automation (`retry_join` + the existing postStart Shamir hook covers all 3).
- Data-loss **risks + mitigations** (empty-cluster flip, double-init split-brain, unseal-before-join, even-node quorum, blocked 8201, audit fail-closed ×3, no-backup).
- **transit auto-unseal** hardening as a follow-up.

`openbao` stays exempt from `require-three-replicas` until this runs (already noted in the policy).

### DIFF
--- a/docs/dr/openbao-raft-ha-migration.md
+++ b/docs/dr/openbao-raft-ha-migration.md
@@ -1,0 +1,168 @@
+# OpenBao single-node → 3-node Raft HA migration
+
+> **Status: runbook / not yet executed.** This is the one HA item that is **not**
+> a GitOps auto-apply. OpenBao currently runs `standalone` mode with **file**
+> storage; flipping the HelmRelease to Raft without a data migration brings up
+> empty raft pods and **wipes every secret cluster-wide** (all ExternalSecrets,
+> PushSecrets, the CNPG/FleetDM DB-engine roles). Execute the steps below
+> deliberately; do not merge the values change ahead of the migration.
+
+## Why this can't be a values flip
+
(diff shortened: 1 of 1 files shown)


devantler:
Superseded: the OpenBao item is now the actual Raft HA switch (nothing critical stored yet), not just a runbook. See the new OpenBao PR.

PR #1929, opened 2026-06-09
refactor(infrastructure): use path-only kustomize patches where self-identifying

devantler:
## What

Kustomize patch entries mixed two equivalent styles — an explicit `target:` selector next to `path:`, and `path:` alone — sometimes in the same file. Since every **strategic-merge** patch file already carries full identity (`apiVersion`/`kind`/`name`/`namespace`), the `target:` block merely restated it and was redundant.

## Change

Drop the redundant `target:` selectors, keeping `path:` alone — the path-only style the hetzner overlays already use and that the coroot-patch comment documents. Converted 11 entries across 3 files:

- `providers/docker/infrastructure/controllers` — 7 HelmRelease/KubeVirt patches
- `providers/hetzner/apps` — 3 (ascoachingogvaner, headlamp, wedding-app)
- `providers/hetzner/infrastructure` — 1 (gateway-patch)

**Principled exception:** `whoami`'s patch is **JSON6902** (a list of ops, no embedded identity) so it genuinely requires a `target:` — kept, now with a comment explaining why. The remaining mix is intentional and documented, not accidental.

## Validation (behavior-preserving)

- `kubectl kustomize` for all three affected provider paths is **byte-identical** to before.
- `ksail workload validate`: **local 0 failures**; **prod** unchanged (only the pre-existing coroot catalog-schema failure).

### DIFF
--- a/k8s/providers/docker/infrastructure/controllers/kustomization.yaml
+++ b/k8s/providers/docker/infrastructure/controllers/kustomization.yaml
@@ -12,39 +12,14 @@ components:
   # Platform-wide HelmRelease drift detection (mode: enabled) — see
   # bases/components/helmrelease-drift-detection.
   - ../../../../bases/components/helmrelease-drift-detection
+# Each patch file self-identifies (apiVersion/kind/name/namespace), so a path is
+# enough — no redundant `target:` selector. Matches the path-only style used by
+# the hetzner overlays.
 patches:
-  - target:
-      kind: HelmRelease
-      name: cilium
-      namespace: kube-system
-    path: cilium/patches/helm-release-patch.yaml
-  - target:
-      kind: HelmRelease
-      name: metrics-server
-      namespace: kube-system
-    path: metrics-server/patches/helm-release-patch.yaml
-  - target:
-      kind: HelmRelease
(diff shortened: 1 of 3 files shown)


devantler:
The failing **🧪 System Test** here is **not caused by this PR** — it's a pre-existing regression on `main`.

This PR's change is render-neutral (path-only patch style); its flagger HelmRelease is byte-identical to `main` (`git diff origin/main -- .../flagger/helm-release.yaml` is empty). The System Test fails at the flagger Helm install:

```
error while running post render on manifests: replace operation does not apply:
doc is missing path: /spec/strategy/type: missing value
```

Root cause: #1900 enabled flagger `leaderElection`, and the chart only renders `strategy.type: Recreate` when leaderElection is disabled — so the rendered Deployment no longer has a `strategy` block, and the post-renderer's `op: replace /spec/strategy/type` hits a missing path. This wedges `infrastructure-controllers` and fails the System Test on every k8s PR.

Fix in #1938 (changes the post-renderer to `op: add /spec/strategy`). Once that merges, re-running this PR's System Test should go green. No change needed here.

PR #1938, opened 2026-06-09
fix(flagger): make strategy post-renderer robust to the missing chart strategy block

devantler:
## Problem

The System Test fails on **every** k8s PR (surfaced on #1929, but not caused by it). The flagger HelmRelease install fails at post-render:

```
Helm install failed for release flagger-system/flagger with chart flagger@1.43.0:
error while running post render on manifests: replace operation does not apply:
doc is missing path: /spec/strategy/type: missing value
```

…which wedges `infrastructure-controllers → infrastructure → apps`.

## Root cause (regression from #1900)

[#1900](https://github.com/devantler-tech/platform/pull/1900) enabled flagger `leaderElection` for HA. The flagger chart only renders `strategy.type: Recreate` when leaderElection is **disabled**; with it **enabled** the chart **omits the `strategy` block entirely**. The existing post-renderer used `op: replace /spec/strategy/type`, and `replace` fails when the target path doesn't exist — so the install now dies deterministically.

(#1900's own System Test presumably passed before the leaderElection commit landed, or merged via the merge queue, which skips the docker System Test.)

## Fix

Replace the `replace`+`add` pair with a single **`op: add /spec/strategy`** for the whole strategy object. `/spec` always exists on a Deployment, so `add` creates-or-replaces deterministically whether or not the chart renders a strategy block — and still pins the explicit RollingUpdate defaults so drift detection (which doesn't ignore `/spec/strategy`) sees no diff. The comment is updated to reflect the leaderElection interaction.

## Validation

- `ksail workload validate`: **local 0 failures**; **prod** unchanged (only the pre-existing coroot CRD-catalog schema flake). Both provider controller overlays build.
- The post-render itself is only exercisable in the **System Test** (it runs at Helm install time) — this PR's System Test is the real confirmation. `op: add` on `/spec` cannot raise the "missing path" error by construction.

## Impact

Unblocks the System Test for all open k8s PRs (#1927/#1928 already merged; this restores green for #1929, #1931, and the merge queue).

### DIFF
--- a/k8s/bases/infrastructure/controllers/flagger/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/flagger/helm-release.yaml
@@ -81,22 +81,23 @@ spec:
       limits:
         memory: 256Mi
 
-  # Keep the Deployment on a RollingUpdate strategy. The chart hardcodes
-  # `strategy.type: Recreate` on the flagger Deployment
-  # (charts/flagger/templates/deployment.yaml) with no values override. An
-  # earlier chart revision rendered RollingUpdate, so the live Deployment carries
-  # an API-defaulted `rollingUpdate{maxSurge,maxUnavailable}` block. Server-side
-  # apply does NOT prune that block when `type` flips to Recreate, so the merged
-  # object is rejected as invalid — `spec.strategy.rollingUpdate may not be
-  # specified when strategy type is 'Recreate'` — and every Helm upgrade fails
-  # and rolls back, stalling the flagger HelmRelease and wedging
-  # infrastructure-controllers (→ infrastructure → apps). Flagger runs leader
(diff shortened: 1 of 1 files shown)


devantler:
Closing as **redundant** — `main` already has this exact fix.

#1901 (the VPA/HA follow-up) independently caught the same #1900 regression and applied the **identical** post-renderer change — `op: add /spec/strategy` with the same `{type: RollingUpdate, rollingUpdate: {maxSurge: 25%, maxUnavailable: 25%}}` values (verified byte-identical patch ops). The "merge conflict" here is just both branches changing the same lines to the same thing.

Net: nothing to merge from this PR. The flagger install regression is already resolved on `main`, which should unblock the System Test for the remaining k8s PRs (#1929, #1931).

PR #1957, opened 2026-06-10
fix(infra): stop health-gating deploys on the over-provisioning buffer

devantler:

## Problem

Every prod deploy (merge-queue `deploy-prod` and tag `CD`) has been red for ~30h with:

```
health check failed after 20m: timeout waiting for: [ReplicaSet/overprovisioning/overprovisioning status: 'InProgress']
```

The `infrastructure` Flux Kustomization was stuck Progressing on this single resource, which also blocked its `apps` dependent — the apps layer was pinned to a 30h-old revision.

## Root cause

Two stacked causes:

1. **Abandoned cordon (live state, already fixed):** a `ksail cluster update` cordoned `autoscale-cx33-3b1bb19008b35538` on 2026-06-08T21:46Z (field manager `ksail` on `spec.unschedulable`) and never uncordoned it, wasting a quarter of the autoscale pool for ~32h. I uncordoned it live.
2. **Design defect (this PR):** even after the uncordon, the buffer's 2 CPU / 4Gi pause pod cannot fit anywhere once the Cluster Autoscaler hits `maxNodesTotal` (`NotTriggerScaleUp: max total nodes in cluster reached`). That is the buffer working **as designed** — it exists to be the first thing squeezed out at the capacity ceiling — but it lived in the `wait: true` infrastructure layer, and kstatus reports a ReplicaSet with Pending pods as InProgress forever. Flux has no per-resource health-check opt-out, so an expected-Pending placeholder health-gated the entire pipeline.

## Fix

- Move `overprovisioning/` out of `providers/hetzner/infrastructure/` into a sibling `providers/hetzner/infrastructure-overprovisioning/` (manifests unchanged).
- Apply it via a new prod-only `infrastructure-overprovisioning` Flux Kustomization with `wait: false` (apply-only; pod schedulability is explicitly not a health signal for a preemption buffer), `dependsOn: infrastructure`, `prune: true`.
- Exclude that one Kustomization by name from the `enforce-flux-best-practices` Kyverno rule that hard-requires `wait: true` (documented as the one sanctioned exception).

Note: on first reconcile after merge there can be a one-cycle admission race (the cluster-level Kustomization CR is admitted by the *old* policy until `infrastructure` applies the updated one); Flux retries make it self-heal within one retry interval.

## Validation

- `kubectl kustomize k8s/clusters/local/` ✅ and `k8s/clusters/prod/` ✅ (rendered CR verified: `wait: false`, correct path).
- `ksail workload validate` ✅ (314 files).
- `ksail --config ksail.prod.yaml workload validate`: only failure is the **pre-existing** stale `coroot.com/v1` CRDs-catalog schema (`notificationIntegrations not allowed`), present on main and untouched here.

## Live remediation already applied (will be superseded by this PR)

- `kubectl uncordon autoscale-cx33-3b1bb19008b35538` — restored pool capacity; `infrastructure` + `apps` reconciled to the latest revision.
- `kubectl scale rs -n overprovisioning overprovisioning --replicas=0` — temporary; Flux re-asserts `replicas: 1` from git (under the new non-gating Kustomization once this merges).


devantler:

Heads-up on urgency: every `infrastructure` re-apply (artifact change or 60m interval) re-asserts `replicas: 1` on the buffer ReplicaSet, after which the health-check wedge recurs and the merge queue goes red again (~20m health-check timeout per attempt). I am re-applying the temporary `scale --replicas=0` as needed, but that lasts only until the next re-apply — merging this PR is what stops the cycle.

This PR's own merge-queue run should pass even with the buffer Pending: its artifact moves the ReplicaSet out of the health-gated layer before health checks run (modulo the one-cycle Kyverno admission race documented in the description, which self-heals within a Flux retry).

PR #1965, opened 2026-06-10
fix(openbao): make init/unseal deterministic per-pod and retry the postStart unseal

devantler:

## Problem

The 2026-06-10 Raft-HA cutover (#1907) wedged on prod: `openbao-1/2` came up **sealed and stayed sealed**, while `openbao-0` (not yet rolled — `OnDelete` update strategy) kept serving from the old file backend. Root causes, both structural:

1. **The vault-config Job does every status/init/unseal call through the round-robin `openbao` Service.** With 3 HA pods (sealed ones deliberately Ready via `sealedcode=204`), consecutive requests land on *different* pods: the "is it initialized?" check, the `operator init` and the `operator unseal` can each hit a different node. These are exactly the **double-init split-brain** and **unseal-before-join** risks the runbook (`docs/dr/openbao-raft-ha-migration.md`) warns about.
2. **The server postStart unseal hook was one-shot**: it waited for one `bao status` to answer, and a single transient failure of the *next* status call (or an unseal against a follower that hadn't `retry_join`ed yet) was swallowed by `|| true` — pod sealed until manual intervention, forever.

## Fix

**vault-init (Job)** — per-pod, deterministic:
- addresses pods individually via `openbao-<i>.openbao-internal` headless DNS (`publishNotReadyAddresses: true`, verified live)
- `operator init` runs **exactly once, on openbao-0**, and only when **no** pod reports an existing cluster (a fresh openbao-0 can report `initialized: false` while peers hold raft data — initializing it then would split-brain)
- unseals **every** pod with retries (36×5s): followers join via `retry_join` once the leader is unsealed, then their unseal succeeds; a wrong-key attempt fails harmlessly and is logged
- replica count via the same `${openbao_replicas:=1}` Flux post-build variable the HelmRelease uses (prod = 3, local/CI = 1 → behavior there is unchanged)
- succeeds when ≥1 pod is unsealed (active Service has an endpoint); stragglers retry on the next Flux-forced Job recreation

**vault-config main container** — `BAO_ADDR` → `openbao-active` (configuration writes never hit sealed standbys).

**postStart hook (HelmRelease)** — retry loop (~2 min), exits 0 immediately on `initialized: false` (cold bootstrap is the Job's job), never fails the hook (sealed pods stay Ready by design).

Net effect: a rescheduled/restarted pod self-unseals (hook), and even if the hook window is missed, the next vault-config run unseals all pods — today's wedge becomes self-healing.

## Validation

- `ksail workload validate`: ✅ 314 files
- `sh -n` syntax check on both rendered scripts (with `${openbao_replicas:=1}` substituted): ✅
- prod overlay: 1 pre-existing failure only (stale `coroot.com/v1` datreeio catalog schema, unrelated)

Sibling PR #1964 points all OpenBao *clients* (ESO stores, generators, backup job, UI route) at `openbao-active`.

## Note on the live wedge

This PR makes the flow deterministic going forward; the **in-flight cutover still needs the manual completion** per `docs/dr/openbao-raft-ha-migration.md` (unseal the raft followers with the post-init key, then roll openbao-0 onto the raft config). Details in the session handoff.

### DIFF
--- a/k8s/bases/infrastructure/controllers/openbao/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/openbao/helm-release.yaml
@@ -56,21 +56,43 @@ spec:
           readOnly: true
         - name: tmp
           mountPath: /tmp
+      # The hook retries for ~2 min and always exits 0. The previous one-shot
+      # version raced the server: it waited for ONE `bao status` to answer,
+      # then a SINGLE transient failure of the next status call (or an unseal
+      # against a raft follower that had not joined yet) was swallowed by
+      # `|| true` and the pod stayed sealed until manual intervention
+      # (2026-06-10: openbao-1/2 rescheduled mid-cutover and never unsealed).
+      # An uninitialized server exits immediately — on cold bootstrap the
+      # vault-config Job owns init and unseals every pod afterwards.
       postStart:
         - /bin/sh
         - -ec
         - |
-          export BAO_ADDR="${BAO_ADDR:-http://127.0.0.1:8200}"
-          sleep 5
+          export BAO_ADDR="http://127.0.0.1:8200"
           UNSEAL_KEY=$(cat /openbao/unseal/unseal-key 2>/dev/null || true)
           if [ -n "$UNSEAL_KEY" ] && [ "$UNSEAL_KEY" != "REPLACE_AFTER_INIT" ]; then
-            until bao status -format=json 2>/dev/null | grep -q '"sealed":'; do
-              sleep 2
+            ATTEMPTS=24
+            while [ $ATTEMPTS -gt 0 ]; do
+              STATUS=$(bao status -format=json 2>/dev/null || true)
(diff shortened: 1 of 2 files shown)


devantler:

Live evidence captured after this PR was opened — the round-robin failure mode is actively crashlooping the current vault-config Job on prod (pod `vault-config-2wgww`, 9 restarts): the `vault-init` init container happened to land on openbao-0 twice ("Vault is already initialized. Vault is already unsealed.") and passed, then the main `vault-config` container's writes round-robined onto a sealed standby and died:

```
GET  http://openbao.openbao.svc.cluster.local:8200/v1/sys/mounts        -> 503 * Vault is sealed
POST http://openbao.openbao.svc.cluster.local:8200/v1/sys/mounts/secret -> 503 * Vault is sealed
```

Exactly the two call sites this PR repoints (per-pod init/unseal + `openbao-active` for the config writes).

PR #1981, opened 2026-06-10
fix(vault-seed): cache generated secrets in K8s Secrets and push from the cache

devantler:

## Root cause

The generated secrets (dex / flux-web client secrets, oauth2-proxy cookie secret, fleetdm bootstrap passwords, umami app/admin/tenant passwords) were pushed **straight from ESO Password generators**. A `generatorRef`-selector PushSecret produces a *new random value on every sync*, so they had to be push-once (`refreshInterval: "0"`) — which made OpenBao the only place the values existed. When the 2026-06-10 incident re-initialized the vault with an empty KV store (context in #1979/#1980), they were unrecoverable, and every consumer ExternalSecret (`vault-config-oidc`, `headlamp-oidc`, `actual-budget-oidc`, `umami-admin`, …) wedged in `SecretSyncedError` with no self-healing path.

## Fix

Each generated secret becomes two steps (the proven `push-umami-db-superuser` pattern):

1. a `generated-*` **ExternalSecret** runs the Password generator **exactly once** (`refreshInterval: "0"`) and persists the value in a durable cache Secret in the `openbao` namespace
2. the existing **PushSecret** mirrors that cache Secret into OpenBao hourly (`refreshInterval: 1h`, selector switched from `generatorRef` to the cache Secret)

Healthy vault → hourly push is a no-op; values never rotate. Wiped vault → KV re-seeds within the hour from the cache.

## ⚠️ One-time effect on the current incident

Because the old generated values died with the wiped vault, the caches will generate **fresh values** on first reconcile. All consumers converge automatically via their ExternalSecrets (SSO clients, cookie secret → existing sessions reset; fleetdm has no pods yet, harmless), **except**:

- **umami admin password**: umami stores the hash in its DB, and the provision job can only rotate from the default `umami` password. After this merges, the admin password must be updated once in the umami UI (log in with the *old* value from the current `umami/umami-admin` Secret, set it to the new value from `openbao/generated-umami-admin-password`). The provision-tenants CronJob then goes green again.

## Validation

- `ksail workload validate` → 314 files validated ✅
- `kubectl apply --dry-run=client` on the changed file against live CRDs ✅

### DIFF
--- a/k8s/bases/infrastructure/vault-seed/generators.yaml
+++ b/k8s/bases/infrastructure/vault-seed/generators.yaml
@@ -1,7 +1,8 @@
 # ESO Password generators for secrets that can be randomly generated.
-# Each generator defines the password spec; PushSecrets in
-# push-generated-secrets.yaml reference these via generatorRef to seed
-# OpenBao on first reconciliation.
+# Each generator defines the password spec; the `generated-*`
+# ExternalSecrets in push-generated-secrets.yaml run them exactly once
+# (refreshInterval "0") to materialize a durable cache Secret, which the
+# PushSecrets then mirror into OpenBao hourly.
 ---
 apiVersion: generators.external-secrets.io/v1alpha1
 kind: Password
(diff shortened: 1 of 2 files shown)


devantler:

PR #1982, opened 2026-06-10
fix(vault-config): refuse to auto-init when openbao-unseal already holds keys

devantler:

## Root cause

The `vault-init` init container auto-runs `bao operator init` whenever **no pod reports an initialized barrier**. That state is ambiguous:

- fresh install → auto-init is exactly right
- **data loss** (raft dir missing/unreadable, e.g. after a storage-backend cutover gone wrong) → auto-init silently brings up an **empty vault**, and `store-keys` then overwrites the previous unseal key + root token in `openbao-unseal`, making any surviving data unrecoverable

The second case is precisely how the 2026-06-10 incident destroyed the entire KV store: the standalone→raft deploy flip-flop (#1907 deployed pre-merge at 06:43 UTC, reverted by a main deploy at 13:08, re-applied at 14:57) left `openbao-0`'s data dir without a raft db; when the pods were rolled, the Job saw "uninitialized", initialized a blank vault over it at 17:48 UTC, and replaced the old keys.

## Fix

If no pod is initialized **but the `openbao-unseal` Secret already holds keys from a previous cluster**, the Job now fails loudly with recovery instructions instead of initializing. A fresh init requires explicit operator acknowledgement of the data loss:

```
kubectl delete secret openbao-unseal -n openbao
```

Unaffected flows: fresh installs (no Secret → init proceeds), Velero restores (Secret + data PVCs both restored → pods report initialized → init skipped), and the current prod cluster (pods report initialized → init skipped).

## Validation

Companion to #1979/#1980/#1981 (2026-06-10 OpenBao incident remediation).

### DIFF
--- a/k8s/bases/infrastructure/vault-config/job.yaml
+++ b/k8s/bases/infrastructure/vault-config/job.yaml
@@ -127,6 +127,27 @@ spec:
               done
 
               if [ "$ANY_INITIALIZED" = "false" ]; then
+                # Data-loss guard: keys in openbao-unseal mean a cluster was
+                # initialized before. If on top of that NO pod reports an
+                # initialized barrier, the raft data is missing or unreadable —
+                # auto-initializing here would silently bring up an EMPTY vault
+                # and overwrite the previous unseal key + root token (exactly
+                # how the 2026-06-10 incident destroyed the whole KV store).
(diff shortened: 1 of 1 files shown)


devantler:

PR #1983, opened 2026-06-10
feat(openbao): make the vault-snapshot CronJob take real raft snapshots

devantler:

## Root cause / gap

Despite its name, the `vault-snapshot` CronJob **never snapshots anything** — its entire script is a `bao status` health check. When the 2026-06-10 incident wiped the OpenBao KV store (#1979 has the timeline), there was **no snapshot to restore from**: the only off-pod copy would have been Velero's namespace backup, which has been failing for weeks on broken R2 credentials.

## Fix

After the existing health check, the job now runs

```
bao operator raft snapshot save /snapshots/openbao-<utc-timestamp>.snap
```

against the `openbao-active` Service (leader-only, as before), onto a new dedicated `vault-snapshots` PVC (5Gi, default StorageClass), keeping the newest 14 snapshots. The `vault-snapshot-readonly` policy in the vault-config Job gains `read` on `sys/storage/raft/snapshot` (the snapshot export endpoint). ServiceAccount, auth role, schedule (03:30 daily) and CiliumNetworkPolicy are unchanged.

Restore path: `bao operator raft snapshot restore <file>` + unseal with the `openbao-unseal` key that was current when the snapshot was taken — which is also why #1982 refuses to overwrite those keys on silent re-init. Velero's daily namespace backup carries the PVC off-cluster once its R2 credentials work again.

Note: this PR touches `vault-config/job.yaml` in a different section than #1982 — they merge cleanly in either order.

## Validation

- `ksail workload validate` → 315 files validated ✅

### DIFF
--- a/k8s/bases/infrastructure/vault-backup/cronjob.yaml
+++ b/k8s/bases/infrastructure/vault-backup/cronjob.yaml
@@ -1,6 +1,11 @@
-# Daily backup verification of OpenBao data. Vault data is persisted on a
-# PVC and backed up by Velero. This job logs vault health and seal status
-# for monitoring.
+# Daily OpenBao raft snapshot to the vault-snapshots PVC (newest 14 kept),
+# plus a health/seal-status check. Until 2026-06-10 this job ONLY did the
+# health check despite its name — when the vault was wiped that day there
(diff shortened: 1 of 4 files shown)


devantler:

PR #1989, opened 2026-06-10
fix(vault-seed): seed OIDC client/cookie secrets from the SOPS bootstrap source

devantler:

## Root cause

The same logical secret has **two unrelated values**:

- **dex** (staticClients) and **oauth2-proxy** consume `${dex_client_secret}` / `${flux_web_client_secret}` / `${oauth2_proxy_cookie_secret}` directly from the `variables-cluster` SOPS secret via Flux post-build substitution
- **headlamp**, **actual-budget** and the **OpenBao OIDC auth method** read the same logical secrets from the vault paths `infrastructure/oidc/{dex,flux-web,oauth2-proxy}` — which were seeded by an **independent Password generator**

Verified live on prod just now: `sha256(variables-cluster.dex_client_secret) ≠ sha256(vault-synced dex-client-secret)` → dex rejects every vault-sourced consumer. This predates the 2026-06-10 incident (the generator never knew about the SOPS value), and the post-incident regeneration (#1981) re-rolled the dice.

## Fix

Replace the three generator-based pushes with `seed-*` PushSecrets that mirror the `variables-cluster` keys into the same vault paths (hourly, like every other SOPS-sourced seed). Drop the now-unused Password generators and cache ExternalSecrets (Flux prunes them; the cache Secrets are GC'd with their owners). The truly generated secrets (umami / fleetdm) keep the #1981 cache pattern — they have no external system of record.

After merge I'll re-run the vault-config Job once so the OpenBao OIDC auth method picks up the corrected `oidc_client_secret`.

## Validation

### DIFF
--- a/k8s/bases/infrastructure/vault-seed/generators.yaml
+++ b/k8s/bases/infrastructure/vault-seed/generators.yaml
@@ -7,42 +7,6 @@
 apiVersion: generators.external-secrets.io/v1alpha1
 kind: Password
 metadata:
-  name: dex-client-secret
-  namespace: openbao
-spec:
-  length: 32
-  digits: 8
-  symbols: 0
-  noUpper: false
-  allowRepeat: true
----
-apiVersion: generators.external-secrets.io/v1alpha1
-kind: Password
-metadata:
-  name: flux-web-client-secret
-  namespace: openbao
-spec:
-  length: 32
-  digits: 8
-  symbols: 0
-  noUpper: false
-  allowRepeat: true
----
-apiVersion: generators.external-secrets.io/v1alpha1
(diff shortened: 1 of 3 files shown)


devantler:

ℹ️ The 🧪 System Test failure here is the **pre-existing repo-wide CI breakage** (OpenBao service-registration label updates rejected with HTTP 422 → `openbao-active` Service never gains endpoints), not caused by this PR — this run also served as the control proving #1985 alone doesn't cure CI. Root-causing continues in #1990, whose diagnostics probe prints the apiserver's full 422 reason. This PR itself is static-validated against both overlays.

Issue #1880, opened 2026-06-08
PDBs with minAvailable: 1 over ≤1-replica workloads make nodes undrainable (blocks ksail recycles & rolling reboots)


devantler:
A prod `ksail cluster update` failed when ksail's autoscaler-node recycle hung
draining `cert-manager/origin-ca-issuer` for the full 2-minute eviction timeout:

```
evicting pod cert-manager/origin-ca-issuer-7c6f64969c-m4m59
error when evicting pods/"origin-ca-issuer-..." -n "cert-manager" (will retry after 0s): Cannot evict pod as it would violate the pod's disruption budget.
...
✗ failed to apply updates: ... drain node autoscale-cx33-...: global timeout reached: 2m0s
```

## Root cause

`origin-ca-issuer`'s PodDisruptionBudget sets `minAvailable: 1` and is applied
**unconditionally**, but its Deployment replica count is variable
(`replicaCount: ${origin_ca_issuer_replicas:=2}`) and is set to **`0`** in prod
(intentionally disabled — *"Non-essential services scaled to 0 … Disabled until
we actually need Cloudflare Origin certs."*).

- `k8s/providers/hetzner/infrastructure/controllers/origin-ca-issuer/pod-disruption-budget.yaml` → `minAvailable: 1`
- `k8s/providers/hetzner/infrastructure/controllers/origin-ca-issuer/helm-release.yaml` → `replicaCount: ${origin_ca_issuer_replicas:=2}`
- `k8s/clusters/prod/bootstrap/variables-cluster-config-map.yaml` → `origin_ca_issuer_replicas: "0"`

A `minAvailable: 1` PDB over a workload running fewer than 2 ready replicas
permits **zero** voluntary disruptions, so any node hosting its pod can never be
drained. The eviction API loops indefinitely until the drain timeout aborts —
deadlocking node recycles and rolling reboots.

The correct pattern already exists in this repo:
`k8s/bases/apps/fleetdm/pod-disruption-budget.yaml` uses `minAvailable: 0`
explicitly *"to allow KEDA to scale the deployment to zero replicas."* That same
reasoning just wasn't applied to `origin-ca-issuer` or the single-replica
controllers.

## Fix options (pick one per workload)

1. Scale the PDB with the Deployment — template `minAvailable` (or drop the PDB)
   when replicas ≤ 1, mirroring the existing `fleetdm` `minAvailable: 0`
   precedent for scale-to-zero services.
[...]

PR #1991, opened 2026-06-10
feat(ha): add drain-safe PDBs to the remaining critical controllers and apps

devantler:

## Summary

Completes the #1880/#1882 drain-safe PDB sweep so node drains, Talos upgrades and autoscaler scale-down never take out a whole platform component at once — and never deadlock. Every change uses the house pattern `maxUnavailable: 1`, which the disruption controller evaluates against the live replica count, so it stays drainable even if a replica floor drops.

### Chart-native values (chart supports the maxUnavailable shape)

| Component | What gets a PDB |
|---|---|
| keda 2.20.0 | operator, metrics-apiserver, admission-webhooks |
| external-secrets 2.5.0 | controller, webhook, cert-controller (`minAvailable: null` removes the chart's default minAvailable shape that `validate-pdb-drain-safe` flags) |
| vpa 4.11.0 | recommender + updater (mirrors the existing admissionController PDB) |
| kyverno 3.8.1 | background/reports/cleanup controllers — the chart **auto-enables** their PDBs at replicas > 1 and defaults them to `minAvailable: 1`, so prod (3 replicas each) has been running the flagged shape; this pins `maxUnavailable: 1` instead |

### Raw PDB manifests (no usable chart knob)

| Component | Why raw |
|---|---|
| flagger | chart only offers `minAvailable` |
| keda-add-ons-http external scaler | chart ships a PDB only for the interceptor; the scaler is on the scale-from-zero decision path |
| snapshot-controller (hetzner) | piraeus chart has no PDB value; keeps Velero CSI snapshots from hanging mid-drain |
| umami | Flagger target — selector keys on `app.kubernetes.io/instance` because Flagger rewrites `/name` to `umami-umami-primary` on the primary (see canary.yaml); provision-tenants Job pods carry different labels and are not matched |
| actual-budget | KEDA scale-to-zero, inert at 0 replicas, never blocks a drain at 1 |

### Explicitly unchanged

- **OpenBao**: at `openbao_replicas: 3` the chart's default `server.ha.disruptionBudget` already renders `maxUnavailable: 1` (verified against the 0.28.3 template math).
- Velero, flux-operator, external-dns, spire-server, coroot: intentional singletons / operator-managed, per `validate-replica-floor` exemptions.

## Validation

- `ksail workload validate`: **green** (320 files)
- `ksail --config ksail.prod.yaml workload validate`: only failure is the **pre-existing** coroot schema gap in the upstream datreeio CRDs-catalog (fixed upstream by CRDs-catalog#896, unrelated)
- Every chart-value shape verified by rendering the exact chart version with `helm template` and inspecting the produced PDBs (all render `maxUnavailable: 1`; kyverno verified via its `kyverno.pdb.spec` helper which forbids setting both fields)


devantler:

System-test status: the first run failed in ~3 min on a transient schema-fetch error during `ksail workload validate` (`bases/apps ... validation failed: EOF` — also hit #1995, which touches no manifests). On rerun, **validate passed (✔ 320 files validated — including every PDB added here)** and the job ran the full ~53 min, failing at reconcile with `infrastructure` Progressing / `apps` blocked — the known repo-wide OpenBao label-patch 422 wedge being probed in #1990 (same signature and duration as #1987/#1989, pre-existing on main). Nothing in this PR participates in the OpenBao seeding chain; it can go green once the 422 regression is fixed.

devantler:

Fixed the reported kyverno failure and merged `origin/main` into the branch. Root cause: the kyverno chart defaults each controller's `podDisruptionBudget` to `minAvailable: 1`, and Helm **deep-merges** user values onto chart defaults — so setting only `maxUnavailable: 1` left both fields present, and the chart's `kyverno.pdb.spec` helper hard-fails the whole upgrade at render time wherever a controller runs >1 replica (prod runs 3). The fix adds an explicit `minAvailable: null` to all three controller PDB blocks, which deletes the chart-default key at value coalescing — the same mechanism the external-secrets values in this PR already use.

Verified by rendering kyverno 3.8.1 with all three controllers at 3 replicas: without the null the render reproduces the exact error; with it each controller gets a clean `maxUnavailable: 1` PDB. keda and vpa are unaffected (their chart PDB defaults are `{}`, nothing to merge against). `ksail workload validate` green (323 files).

devantler:

System Test run [27361178852](https://github.com/devantler-tech/platform/actions/runs/27361178852) (the first run after the kyverno `minAvailable: null` fix) failed on the **known local-cluster infra flake**, not on this PR's content: the single-control-plane Docker cluster's API server became unresponsive mid-reconcile — `helm-controller`/`source-controller` stuck in `CreateContainerError` on local-worker-2, lease-lock renewals timing out (`the server was unable to return a response`), `client rate limiter ... context deadline exceeded` across the board — which cascaded into `infrastructure-controllers/infrastructure/apps` reconcile failures. Same etcd-overload signature documented in #1779.

The kyverno values themselves are sound — independently verified by rendering chart 3.8.1 with the fixed values (`minAvailable: null` + `maxUnavailable: 1` at 3/2/2 replicas renders the three PDBs correctly), and prod's kyverno helm upgrade succeeded once main-based artifacts re-deployed at ~16:45 (`kyverno.v8`).

FYI: the wider incident this PR's merge-queue attempt caused is written up in #2026 (merge-group deploys leave prod on unmerged code when the group fails).

devantler:

The 16:18 run on `399f924c` is an infra flake, not a regression: cluster create needed a retry on a slow etcd (`service etcd not in expected state Running`), and the reconcile then cascaded with apiserver rate-limiter timeouts (`failed to list pods in flux-system: client rate limiter Wait returned an error: context deadline exceeded`) taking down 3/5 Kustomizations including infrastructure-controllers — the known single-CP Docker etcd-overload flake. No `Cannot set both` anywhere in the log and no kyverno HelmRelease failure, and the 15:35 run on `6be628a1` (same PDB content, pre-null-fix merge) **passed the full system test**. Re-running the failed job.

devantler:

The 17:36 run on `94a3980f` is the same infra flake as the 16:18 one, with clearer evidence this time: from early in the run the Flux controllers could not even renew their leader-election leases (`Error retrieving lease lock ... Client.Timeout exceeded` / `the server was unable to return a response in the time allotted`), source-controller and helm-controller sat 0/1 Ready, CoreDNS answered 503, and every HelmChart reported 'latest generation has not been reconciled' — cluster-wide API-server starvation on the single-CP Docker cluster, the documented etcd-overload flake. No manifest in this PR is implicated (identical content passed the full system test at 15:35). Re-running.

devantler:

Third occurrence of the runner-starvation flake on this branch (this time visible at the lowest level: **kube-scheduler itself** timing out lease renewal against KubePrism at `127.0.0.1:7445` — below anything Flux deploys). Cross-checking the rest of the fleet this afternoon: six merge-queue and PR runs passed in the same window (including the PDB-policy PR #2025 and the velero-group fix #2031), one unrelated PR flaked the same way. Identical content passed the full system test at 15:35. PDB objects are inert at admission and cannot load an apiserver; this branch is just rolling badly on the runner lottery. Re-running (attempt 4).

PR #1995, opened 2026-06-10
ci: implement the documented Velero restore drill in the system-test job

devantler:
> 🤖 Generated by the Daily AI Assistant

## Summary

[docs/dr/restore-drill.md](docs/dr/restore-drill.md) has documented a CI restore drill since it was written — but **no workflow ever implemented it**. The backup → data-loss → restore path was never regression-tested, so a Velero chart bump, RBAC drift, or MinIO credential break would only surface during a real disaster (as 2026-06-10 demonstrated for the adjacent vault-snapshot path, which turned out to be a health check in a trench coat).

This implements the drill as steps **inside the existing `system-test` job**, reusing the Talos+Docker cluster it just reconciled — a separate job (as the doc originally described) would pay a second ~10-minute cluster bootstrap for no extra signal. Added wall-clock: ~2-3 minutes.

### The drill

1. Wait for `BackupStorageLocation/default` → `Available` (Velero → in-cluster MinIO, the local R2 stand-in — same S3 code path as prod)
2. Create `dr-drill` namespace + marker ConfigMap carrying `run-id`/`sha`
3. `Backup` CR scoped to the namespace; wait for `Completed`, failing fast on `Failed`/`PartiallyFailed`/`FailedValidation`
4. Delete the namespace, assert it's gone
5. `Restore` CR from the backup; wait for `Completed`
6. Assert the restored ConfigMap's `run-id` matches `GITHUB_RUN_ID`

Velero CRs are created with `kubectl` (no velero CLI install, can't drift from the deployed Velero version). On failure the step dumps BSL/Backup/Restore state + the Velero server log.

### Doc truth-ups in the same change

- `restore-drill.md` described a standalone job with its own cluster and a `timeout-minutes: 240` budget that never existed → now describes the implemented steps.
- `runbook.md` claimed the drill asserts etcd encryption-at-rest; `restore-drill.md` itself explicitly scopes that out → corrected.

## Validation

- Workflow YAML parses (yq); the 90-line drill script passes `bash -n`; the heredoc Backup/Restore payloads verified as valid YAML at column 0.
- ⚠️ **Expected CI status**: system-test is currently red repo-wide from the pre-existing OpenBao label-patch 422 regression (being probed by #1990); the drill steps run after reconcile, so this PR can't go green until that's resolved — the drill itself is unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1997, opened 2026-06-10
feat(dr): one-button prod rebuild workflow (runbook scenario 4, executable)

devantler:

## Summary

Makes runbook Scenario 4 (full cluster rebuild from zero) a **single `workflow_dispatch` run** instead of a sequence of local manual commands — completing the "GitHub Actions can recover the platform, including persisted data and the vault" goal.

**Depends on #1996** (the `vault-snapshot-r2` Secret, the R2 snapshot mirror, and the vault-config auto-restore path).

### What the workflow does

| Phase | Detail |
|---|---|
| Confirm | Requires typing `REBUILD-PROD`; shares the `prod-deploy` concurrency group with ci/cd so it can never race a deploy |
| Rebuild | `ksail cluster create` → `workload push` → `reconcile` → wait for all four Flux Kustomizations. Fresh kubeconfig/talosconfig are written **on the runner**, so the workflow is immune to the stale `KUBE_CONFIG`/`TALOS_CONFIG` problem that breaks the regular pipeline after a rebuild |
| Velero restore | Newest `Completed` backup synced from R2; `PartiallyFailed` accepted (most resources already exist on the freshly-converged cluster — the restore's job is what Flux doesn't own) |
| OpenBao recovery | In-cluster pod fetches the newest raft snapshot from the R2 `openbao-snapshots/` mirror onto the PVC (credentials never leave the cluster — reuses the `vault-snapshot-r2` Secret and the CronJob's NetworkPolicy via its `app` label) → pre-incident `openbao-unseal` restored from the Velero backup → fresh vault reset (HR suspended, sts→0, data PVCs deleted) → vault-config's automated snapshot-restore brings the old vault back → ExternalSecrets/PushSecrets force-synced |
| Credentials | `KUBE_CONFIG`/`TALOS_CONFIG` refreshed automatically when a `DR_GH_ADMIN_TOKEN` secret (fine-grained PAT, environment-secrets write) is configured; loud warning otherwise |
| Summary | LB IP, external-dns note, CNPG + per-app data pointers |

### Safety properties

- Aborts the OpenBao surgery **before touching data PVCs** if the old unseal Secret didn't come back from the backup.
- The snapshot-fetch pod and `dr-r2-target` ConfigMap are cleaned up either way.
- `restore=false` input gives a bare rebuild (fresh vault, SOPS-seeded) with no restore steps.

### Runbook truth-ups

- Scenario 4 leads with the workflow; the manual path stays as the GitHub-Actions-is-down fallback.
- The manual **DNS step is gone**: external-dns (`policy: sync`, `gateway-httproute` source) repoints the Cloudflare records automatically after a rebuild — the runbook step predated external-dns. Now it's "verify, intervene only if external-dns is broken".
- Scenario 9 documents the automatic secret refresh; the OpenBao artifacts row points at the snapshot mirror + automated restore.

### Honest limits (stated in the workflow header + runbook)

- Per-app PVC data (headlamp, actual-budget) isn't rehydrated into already-running pods — Velero skips existing resources; per-app reset per Scenario 5.
- CNPG (umami-db) recovers from its own barman/R2 backups (Scenario 5's `kubectl cnpg restore`).
- The workflow can't be end-to-end tested without sacrificing a prod cluster; every step mirrors the runbook commands and all step scripts pass `bash -n`, but treat the first real run as supervised.

## Validation

- Workflow YAML parses (yq); all 7 step scripts pass `bash -n`; embedded Pod manifest parses as valid YAML.
- No k8s manifests changed → no overlay rebuild needed (ci.yaml path filter doesn't trigger system-test for this file).

### DIFF
--- a/.github/workflows/dr-rebuild.yaml
+++ b/.github/workflows/dr-rebuild.yaml
@@ -1,0 +1,391 @@
+# DR — Rebuild Prod: the executable form of docs/dr/runbook.md Scenario 4
+# ("full cluster rebuild from zero"), runnable as one button press when the
+# production cluster is gone.
+#
+# What it does, in order:
+#   1. ksail cluster create  — fresh Hetzner servers, Talos, CCM, CSI
+#   2. workload push + reconcile — Flux converges the platform (fresh, empty
+#      OpenBao; SOPS-sourced secrets re-seed automatically)
+#   3. (restore=true) Velero resource restore from the newest Completed
+#      backup synced from R2
+#   4. (restore=true) OpenBao data recovery: fetch the newest raft snapshot
+#      from the R2 openbao-snapshots/ mirror onto the vault-snapshots PVC,
+#      restore the pre-incident openbao-unseal Secret from the Velero backup,
+#      reset the fresh vault (scale down + delete data PVCs), and let the
+#      vault-config Job's automated snapshot-restore path bring the old vault
+#      back (docs/dr/openbao.md scenario 2/3)
+#   5. Optionally refresh the KUBE_CONFIG / TALOS_CONFIG environment secrets
+#      (requires a DR_GH_ADMIN_TOKEN secret with environment-secrets write;
+#      without it the step prints the manual Scenario 9 instructions)
+#
+# Known limits (documented in the runbook):
+#   * Per-app PVC data (headlamp, actual-budget) is NOT rehydrated into
+#     already-running pods — Velero skips existing resources. Recover an app's
+#     data with the per-app reset dance (runbook Scenario 5).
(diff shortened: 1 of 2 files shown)


devantler:

Applied the lesson from the restore drill's first real run on #1995: `kubectl get backups` resolves to **CNPG's** `backups.postgresql.cnpg.io` on this cluster, not Velero's. In a real DR, this workflow's backup-discovery loop would have listed CNPG Backups, found no `Completed` ones, and aborted the entire restore with 'No Completed Velero backup found'. All velero `get`/`describe`/`wait` calls now use fully-qualified `*.velero.io` names. All step scripts still pass `bash -n`.

PR #1999, opened 2026-06-10
fix(cluster-policies): mutate pod security contexts on CREATE only

devantler:

## Root cause of the repo-wide CI failure — finally pinned

#1990's patch probe printed the apiserver's full 422:

```
The Pod "openbao-0" is invalid: spec: Forbidden: pod updates may not change fields
other than `spec.containers[*].image`, ... (only additions to existing tolerations) ...
```

…and the accompanying diff shows the rejected change is **not the label patch at all** — it's `capabilities.drop: [ALL]`, `runAsNonRoot: true`, `readOnlyRootFilesystem: true`, `seccompProfile` being injected into the pod spec. That is the `add-security-context` ClusterPolicy: it matches `Pod` with **no `operations:` scope**, so Kyverno mutates **UPDATEs too**.

Pod spec is immutable. So any pod created while the policy/webhook wasn't active yet — exactly what fresh-cluster bring-up ordering produces for `openbao-0` — becomes **permanently un-updatable**: every subsequent update gets the mutation bolted on and 422s. OpenBao's service-registration label updates (`openbao-active/sealed/initialized`) therefore failed forever, `openbao-active` kept zero endpoints, the vault seeding chain timed out, and **every system test since the 2026-06-10 active-service cutover (#1964) failed**. Prod was unaffected only because its pods happened to be recreated while the policy was live (mutation already in the spec → no-op on update).

## Fix

`operations: [CREATE]` on both rules. Mutating immutable pod spec on UPDATE can never succeed — it can only brick pods. Pods created before the policy stay unmutated until their next natural recreation (kubescape may flag them transiently), which is strictly better than being permanently un-updatable.

**This PR's own system test is the verification** — it should be the first green run since 2026-06-10 ~18:30 UTC.

## Validation

- `kubectl apply --dry-run=server` of the policy against prod ✅

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -42,7 +42,7 @@ jobs:
         # cluster-update fixes; renovate keeps this pin current.
         env:
           # renovate: datasource=github-releases depName=devantler-tech/ksail extractVersion=^v(?<version>.+)$
-          KSAIL_VERSION: "7.54.0"
+          KSAIL_VERSION: "7.55.0"
         run: |
           curl -fsSL "https://github.com/devantler-tech/ksail/releases/download/v${KSAIL_VERSION}/ksail_${KSAIL_VERSION}_linux_amd64.tar.gz" -o /tmp/ksail.tar.gz
           tar -xzf /tmp/ksail.tar.gz -C /tmp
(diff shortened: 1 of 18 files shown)


devantler:

✅ **Verified end-to-end**: #2002's system test (this fix + the PVC init Job stacked) just passed — the first green run since 2026-06-10 18:30 UTC. This PR's own red run is expected (it lacks #2002's PVC fix); merging this then #2002 restores CI for every PR.

PR #2071, opened 2026-06-16
fix(longhorn): keep replicas off ephemeral autoscaler nodes

devantler:
## Problem (root cause of the 2026-06-16 volume faulting)

`ksail.prod.yaml` provisions the `autoscale-cx*` pools as **compute-only** burst capacity and states the intent plainly: *"the static workers are the guaranteed baseline and the only Longhorn storage nodes (autoscaler nodes are compute-only)."*

But KSail boots autoscaler nodes from the **same Talos worker config** as the static baseline workers, so they inherit `node.longhorn.io/create-default-disk=true` (`talos/workers/node-labels.yaml`) and Longhorn treats them as storage nodes. Live confirmation:

| Node | Longhorn `create-default-disk` | Ephemeral? |
|---|---|---|
| `prod-worker-1/2/3` (cx43) | `true` | no — static baseline |
| `autoscale-cx33-…`, `autoscale-cx53-…` | `true` | **yes — autoscaler-managed** |

When the cluster-autoscaler scales an `autoscale-*` node down, every Longhorn replica it held is destroyed. A volume whose replicas all sat on now-removed autoscaler nodes goes **`faulted` — data loss, not just degraded.**

**Observed on prod 2026-06-16** (3 `autoscale-cx33-*` nodes scaled down ~06:26):
- `wedding-app/wedding-db` CNPG cluster **0/3 ready** (all 3 instance PVCs faulted) — full DB outage
- `umami/umami-db` degraded (1/3), `kubescape` storage PVC faulted, `headlamp` + `actual-budget` volumes unknown
- `openbao/vault-snapshots` volume stuck → `vault-snapshot-init` job blocked → **`flux-system/infrastructure` Kustomization stuck → `apps` + `infrastructure-overprovisioning` blocked**

`replicaAutoBalance: best-effort` (added the same day) keeps replicas spread across the *live* nodes but cannot save a volume whose only replicas leave together on a scale-down — **the autoscaler nodes must not host replicas in the first place.**

## Fix

A Kyverno mutate `ClusterPolicy` (`restrict-longhorn-to-baseline-nodes`, Hetzner overlay alongside `prefer-baseline-nodes`) that sets `spec.allowScheduling: false` on the **Longhorn `Node` CR** of every `autoscale-*` node. Longhorn then keeps replicas on the three static workers only (matching `defaultReplicaCount: 3`), so each RWO volume retains full 3-way redundancy on stable nodes. Baseline workers (`prod-worker-*`) and control planes never match.

- Admission-time mutate on **CREATE + UPDATE** — safe because `allowScheduling` is a *mutable* Longhorn field (unlike the immutable Pod `spec.affinity` that forces `prefer-baseline-nodes` to CREATE-only).
- Node-name prefix is the only discriminator available today; the proper long-term fix is for KSail to stamp `ksail.io/autoscaled=true` on its pools (the same label `prefer-baseline-nodes` is waiting on). Filed as a follow-up.

## Validation

- **kyverno CLI v1.18.1 unit tests:** `autoscale-cx33-…` → `allowScheduling: false` (pass); `prod-worker-1` → skipped, unchanged.
- `kubectl kustomize k8s/clusters/prod/` and `k8s/clusters/local/` both build.
- `ksail --config ksail.prod.yaml workload validate` → 324 files validated (incl. this policy).

## Not covered here (operational / separate)

- **Longhorn Node CRs that already exist** when this lands are not retroactively patched (admission-only; a `mutateExisting` rule would need extra background-controller RBAC on `longhorn.io/nodes`). The 2 current autoscaler nodes should be set `allowScheduling: false` once by hand; they then recycle into policy-covered nodes.
- **Recovering the already-faulted volumes** (wedding-db restore from CNPG/Velero backup, clearing the stuck `vault-snapshots` PVC, pruning 3 stale Longhorn node objects) is live remediation, tracked separately.

### DIFF
--- a/k8s/providers/hetzner/infrastructure/cluster-policies/restrict-longhorn-to-baseline-nodes.yaml
+++ b/k8s/providers/hetzner/infrastructure/cluster-policies/restrict-longhorn-to-baseline-nodes.yaml
@@ -1,0 +1,84 @@
+# Keep Longhorn replicas off the ephemeral cluster-autoscaler nodes.
+#
+# Why: ksail.prod.yaml provisions the `autoscale-cx*` pools as compute-only
+# burst capacity that scales back to zero — but KSail boots them from the SAME
+# Talos worker config as the static baseline workers, so they inherit
+# `node.longhorn.io/create-default-disk=true` (talos/workers/node-labels.yaml)
+# and Longhorn treats them as storage nodes. When the autoscaler scales such a
+# node down, every replica it held is destroyed; a volume whose replicas all
+# sat on now-removed autoscaler nodes goes `faulted` — data loss, not just
+# degraded. Observed 2026-06-16: wedding-db (0/3 instances), umami-db, the
+# kubescape storage PVC and the openbao vault-snapshots volume all faulted
+# after the `autoscale-cx33-*` nodes holding their replicas were scaled down.
+# `replicaAutoBalance: best-effort` (longhorn/helm-release.yaml) keeps replicas
+# spread across the *live* nodes but cannot help once a volume's only replicas
+# leave together on a scale-down — the autoscaler nodes must not host replicas
+# in the first place.
+#
+# The intent in ksail.prod.yaml is explicit: "the static workers are the
+# guaranteed baseline and the only Longhorn storage nodes (autoscaler nodes are
+# compute-only)". This policy enforces that intent. The proper long-term fix is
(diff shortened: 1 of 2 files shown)


devantler:
Keeping the autoscaler nodes as Longhorn storage nodes is intended, so excluding them is not the desired approach. Pivoting to recovering the faulted volumes / unblocking the infrastructure→apps chain instead.

PR #2086, opened 2026-06-16
feat(autoscaler): replace balloon-pod overprovisioning with native CapacityBuffer

devantler:
## What & why

KSail now ships a **native** Cluster Autoscaler over-provisioning mechanism — [ksail#5277](https://github.com/devantler-tech/ksail/pull/5277) added a `spec.cluster.autoscaler.node.capacityBuffers` knob (merged 2026-06-16). When enabled, KSail installs the `CapacityBuffer` CRD (`capacitybuffers.autoscaling.x-k8s.io`) plus the buffer controller and pod-injection flags. `CapacityBuffer` resources then reserve scale-up headroom as **virtual, pod-less chunks** simulated in autoscaler memory.

This replaces the platform's hand-rolled **low-priority balloon-pod** buffer (a `pause`-pod `ReplicaSet` at priority −1) with the upstream-native equivalent.

## The key win: no more `wait: false` special-case

The balloon buffer's pause pods are *expected* to sit `Pending` at the autoscaler ceiling. Under Flux `wait: true`, kstatus reports a ReplicaSet with Pending pods as `InProgress` forever — which held `infrastructure` (+ its `apps` dependent + every CI/CD deploy) **red for ~30h on 2026-06-09/10**. That forced a dedicated, non-health-gated `infrastructure-overprovisioning` Flux Kustomization and a Kyverno carve-out.

The native buffer creates **no real pods**, so nothing sits Pending. The CapacityBuffer exposes no Pending-pod / `Reconciling` / `Stalled` signal for kstatus to hang on, so the buffer **folds back into the health-gated `infrastructure` layer** and the special-case Kustomization + Kyverno exclusion are deleted.

## Changes

| | |
|---|---|
| **Add** | `providers/hetzner/infrastructure/overprovisioning/` — core/v1 `PodTemplate` (one chunk = 2 CPU/4Gi), `CapacityBuffer` (`replicas: 1`, `autoscaling.x-k8s.io/v1beta1`, `podTemplateRef`), and the `overprovisioning` namespace |
| **Enable** | `capacityBuffers: true` under `autoscaler.node` in `ksail.prod.yaml` ⚠️ *protected file — see below* |
| **Remove** | the balloon `ReplicaSet` + `PriorityClass`, the dedicated `infrastructure-overprovisioning` Flux Kustomization, and its Kyverno `enforce-flux-best-practices` `wait:false` exclusion |

Net **−78 lines** (172 insertions / 250 deletions). Sizing is preserved (2 CPU/4Gi × 1 chunk ≈ one warm spare node, bounded by `maxNodesTotal: 10`); the `-1` `PriorityClass` is gone (the native buffer doesn't use the balloon-preemption trick).

## ⚠️ Sequencing — do not tag a deploy until KSail catches up

`capacityBuffers` is **not in a KSail release yet** — it's 4 commits ahead of the latest tag (`v7.58.3`). The platform pins KSail via a renovate-managed `KSAIL_VERSION` (currently **7.55.0**) in `ci.yaml` / `cd.yaml` / `dr-rebuild.yaml`.

**This PR must not be promoted+tagged for deploy until** ksail cuts a release with #5277 **and** renovate bumps `KSAIL_VERSION` to it. Without that, `cluster update` won't install the CapacityBuffer CRD, and Flux's apply of the `CapacityBuffer` CR would fail.

Good news: `ksail.prod.yaml` validates cleanly on the *current* pinned KSail — older KSail silently ignores the unknown `capacityBuffers` field (no hard config-parse failure), so the only consequence of a premature deploy is "CRD missing", not a crash. I intentionally did **not** bump `KSAIL_VERSION` here (renovate owns it, and the target release doesn't exist yet).

## Protected-file note

`ksail.prod.yaml` is marked *"Protected — Do Not Modify"* in `AGENTS.md`. The one-line `capacityBuffers: true` toggle is the documented way to enable the feature and was made at your explicit direction; it's flagged here for visibility. As a draft, nothing deploys until you promote + tag.

## Topology note (residual uncertainty)

I folded the CapacityBuffer into the `infrastructure` (`wait: true`) layer per your call. My reading of kstatus says it won't hang on a CapacityBuffer (no `Reconciling`/`Stalled` conditions, no top-level `observedGeneration`), but I **couldn't verify on a live cluster** (no cluster spun up). If the live cluster ever shows the `infrastructure` Kustomization gated on the CapacityBuffer, the fallback is to give it back its own dedicated Kustomization (now safely `wait: true`, no Kyverno carve-out needed).

## Validation

- `kubectl kustomize` — `clusters/prod`, `clusters/local`, `providers/hetzner/infrastructure`, `providers/docker/infrastructure` all build green.
- `ksail --config ksail.prod.yaml workload validate` — **303 files validated, exit 0**, including all three new manifests.
- Full system test runs in CI on this PR (local Talos+Docker). Note: the buffer is Hetzner/prod-only and `capacityBuffers` is ignored without the node autoscaler, so the local system test exercises the manifests' validity, not the live buffer.


devantler:
Superseded by #2087. This PR was branched from the monorepo's pinned platform submodule commit, which was ~50 commits behind `main` and predated #2035 — so it missed the cluster-proportional-autoscaler now part of the buffer stack. #2087 redoes the migration on current `main`, retiring the CPA HelmRelease/HelmRepository/NetworkPolicy together with the balloon ReplicaSet (as the CPA HelmRelease header anticipated) and builds conflict-free.

Issue #2149, opened 2026-06-18
Reduce reachable base-image CVEs: pin/refresh operator-managed images flagged by Kubescape


devantler:
> 🤖 Filed via Claude Code (interactive prod survey)

## Problem

Kubescape's reachable-CVE scan flags several workloads on prod running stale base images with **reachable critical** CVEs. The highest-signal hotspots:

| Workload | Reachable criticals | Notes |
|---|---|---|
| `flagger-system/flagger-loadtester` | ~42C / 134H | test-only loadtester, very stale base |
| `dex/dex` | ~25C reachable | internet-exposed (Gateway API) |
| `observability/coroot-db` (CNPG Postgres `18.3-system-trixie`) | 22C / 119H, **5 fixable** (gnutls/openssl) | CNPG operator-default image |
| `observability/coroot*` (coroot, cluster-agent, node-agent) | 9–13C each | **pinned to `:latest`** |
| `velero`, `flux-system/kustomize-controller`, `openbao/vault-snapshot`, `cert-manager/origin-ca-issuer` | 8–12C each | base-image age |

## Why it's not "just Renovate"

Some of these **are** Renovate's job (HelmRelease-pinned tags: dex, flagger-loadtester) and should already flow — worth confirming coverage. But two gaps are not auto-handled:

1. **coroot uses `:latest`** for its own images (coroot / cluster-agent / node-agent). `:latest` defeats CVE tracking entirely (no version to bump, `disallow-latest-tag` can't help) and makes the running image non-deterministic.
2. **CNPG Postgres image is operator-default** (`18.3-system-trixie`), not pinned per-cluster — it only refreshes when the cloudnative-pg operator bumps, so the 5 fixable criticals linger until the operator moves.

## Proposed direction

1. **Pin coroot images** to digests/tags via the Coroot CR (`spec.communityEdition.image` etc., per `k8s/bases/infrastructure/coroot/coroot.yaml`) so they're versioned and Renovate-trackable; drop `:latest`.
2. **Confirm Renovate coverage** for the HelmRelease-pinned images (dex, flagger-loadtester) and close any gaps; consider whether flagger-loadtester is even needed on prod (test helper).
3. **Track the CNPG Postgres base** — either keep the operator current (Renovate on the cloudnative-pg operator) or pin `imageName` per CNPG `Cluster` to a patched tag; this also covers `umami-db`/`wedding-db`.

## Acceptance criteria

[...]

PR #2161, opened 2026-06-19
fix(coroot): pin server/agent images to explicit versions for CVE tracking

devantler:

## Why

Kubescape's reachable-CVE scan flags the coroot stack as running stale base images on a floating tag (platform #2149). Root cause: the **coroot-operator** (v1.9.5, chart `coroot-operator` 0.9.7) leaves the **coroot server / node-agent / cluster-agent** images *unset*, so on every reconcile it lists `ghcr.io/coroot/<component>` and resolves the **highest semver tag** (falling back to `:latest` if that listing fails). The running versions therefore **float** — non-deterministic, and impossible to track for CVEs (no version to bump, `disallow-latest-tag` can't help). This is exactly the gap #2149 calls out for `observability`.

## What

Pin the three component images to their current latest releases on the `Coroot` CR, and make them Renovate-trackable:

| Component | Pinned to |
|---|---|
| `ghcr.io/coroot/coroot` (server / CE) | `v1.22.2` |
| `ghcr.io/coroot/coroot-node-agent` | `v1.33.3` |
| `ghcr.io/coroot/coroot-cluster-agent` | `v1.7.1` |

- Set on `spec.communityEdition.image.name`, `spec.nodeAgent.image.name`, `spec.clusterAgent.image.name`. The operator's `ImageSpec` has only a **`name`** field (full `registry/component:tag` ref) — there is **no** separate `tag`/`repository` subfield, and a bare string under `image:` is silently ignored; pinning `.name` short-circuits the auto-resolution (the operator early-returns on a non-empty `image.name`).
- Added a regex **Renovate `customManager`** (`.github/renovate.json`) scoped to `coroot.yaml`. Renovate's built-in `kubernetes` manager only reads a bare `image:` string, not the `image.name` mapping the CRD uses, so without this the pins would go stale. The manager tracks each `name: ghcr.io/coroot/<component>:<tag>` via the docker datasource → normal bump PRs (minor/patch automerge per existing rules).

This pins to the **current latest** of each, so it's behaviour-preserving versus the operator's own "newest semver" resolution — it just freezes + tracks it instead of letting it drift.

## Validation

- `kubectl kustomize k8s/clusters/local/` and `k8s/clusters/prod/` — both build clean.
- `kubectl kustomize k8s/bases/infrastructure/coroot/` — builds; all three pins present in the output.
- `kubectl apply --dry-run=client -f .../coroot.yaml` — validates against the live `coroot.com/v1` CRD schema (`coroot.coroot.com/coroot configured (dry run)`), confirming the `image.name` placement is accepted.
- `.github/renovate.json` is valid JSON; the regex matches all three pins.

## Notes / scope

- Component versions are released independently (server v1.22.x, node-agent v1.33.x, cluster-agent v1.7.x) — there's no operator-dictated matched set; pinning each to its own latest is the operator's own per-app resolution model. Verified against `coroot-operator@v1.9.5` source (`api/v1/common.go` `ImageSpec`, `controller/versions.go`, `controller/registry.go`).
- Bases hold universal config (this pin applies to every cluster), so it lives in the base, consistent with the recent ClickHouse-HA edit to this same file.
- Remaining `#2149` scope (not in this PR): confirming Renovate coverage for the HelmRelease-pinned `dex`/`flagger-loadtester` images, and a documented stance on the CNPG Postgres base-image refresh cadence.

Part of #2149.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -78,6 +78,18 @@
       "depNameTemplate": "kubernetes/kubernetes",
       "versioningTemplate": "semver",
       "extractVersionTemplate": "^v(?<version>.+)$"
+    },
+    {
+      "customType": "regex",
+      "description": "Track the pinned Coroot CE component images (server, node-agent, cluster-agent) in the Coroot CR. The coroot-operator's `Coroot` CRD nests each image reference under `image.name` (an ImageSpec object), NOT a bare `image:` string, so Renovate's built-in `kubernetes` manager does not detect it — and when the field is left unset the operator auto-resolves the highest semver tag from ghcr on every reconcile (floating, untrackable; the source of the stale base-image CVEs in platform#2149). This regex manager captures each `name: ghcr.io/coroot/<component>:<tag>` and tracks it via the docker datasource so the pins get the normal Renovate bump PRs (minor/patch automerge per the rules below).",
+      "managerFilePatterns": [
+        "/^k8s/bases/infrastructure/coroot/coroot\\.ya?ml$/"
+      ],
+      "matchStrings": [
+        "name:\\s*(?<depName>ghcr\\.io/coroot/[a-z-]+):(?<currentValue>\\S+)"
+      ],
+      "datasourceTemplate": "docker",
+      "versioningTemplate": "semver"
     }
   ],
   "minor": {
--- a/k8s/bases/infrastructure/coroot/coroot.yaml
+++ b/k8s/bases/infrastructure/coroot/coroot.yaml
@@ -30,14 +30,19 @@ spec:
   # auto-resolution (it early-returns on a non-empty `image.name`). NB the
   # ImageSpec has only a `name` field (full ref incl. tag) — there is NO separate
   # `tag`/`repository` subfield, and a bare string under `image:` is silently
-  # ignored; the value MUST be `image.name`. The three pins below are kept
+  # ignored; the value MUST be `image.name`. CRITICAL: Coroot publishes its
+  # container tags as BARE semver (`1.22.2`), NOT the `v`-prefixed GitHub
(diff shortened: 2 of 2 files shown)


devantler:

## ⚠️ Corrected a tag bug that would have frozen prod GitOps

While investigating live-platform health I found the `infrastructure` Kustomization **wedged in a health-check loop**, blocking `apps`, because the three Coroot component pods were in `ErrImagePull`/`ImagePullBackOff`:

```
ghcr.io/coroot/coroot:v1.22.2              -> NotFound (404)
ghcr.io/coroot/coroot-node-agent:v1.33.3  -> NotFound (404)
ghcr.io/coroot/coroot-cluster-agent:v1.7.1 -> NotFound (404)
```

**Root cause:** Coroot publishes its **container** tags as *bare* semver (`1.22.2`), but this PR pinned them to the **`v`-prefixed GitHub _release_ tag** (`v1.22.2`). Verified against ghcr.io:

| ref | HTTP |
|---|---|
| `coroot:v1.22.2` | 404 |
| `coroot:1.22.2` | **200** |
| `coroot-node-agent:v1.33.3` | 404 |
| `coroot-node-agent:1.33.3` | **200** |
| `coroot-cluster-agent:v1.7.1` | 404 |
| `coroot-cluster-agent:1.7.1` | **200** |

The versions chosen were correct (they match `releases/latest`); only the tag *format* was wrong. As-was, **merging this PR would re-freeze prod** the same way.

**Fix (pushed):** dropped the `v` prefix on all three pins → `1.22.2` / `1.33.3` / `1.7.1`, plus a comment documenting the pitfall. `kubectl kustomize` builds clean and the Renovate regex still matches.

**Live incident:** the pins had been applied to the cluster out-of-band (they were never on `main`), so Flux self-healed by reverting the drift; I deleted the one stuck `coroot-coroot-1` straggler to clear the StatefulSet rollout. `infrastructure` + `apps` are Ready again.

PR #2173, opened 2026-06-19
fix(auth-proxy): allow egress to crossview so the SSO upstream resolves

devantler:

## Symptom

`https://crossview.platform.devantler.tech/` is unreachable: after Dex SSO completes, the authenticated user gets an **HTTP 504 after exactly 30 s**. Reported as "crossview is still not responding."

## Root cause

Request path: `Gateway → oauth2-proxy (auth) → auth-proxy (Traefik, host-routes) → crossview-service:80 → crossview pod :3001`.

The crossview **pod is healthy** — it serves its frontend on `127.0.0.1:3001`, the Service (80→3001) and endpoints are correct, the HTTPRoute is `Accepted`/`ResolvedRefs`, and crossview's own ingress policy already allows `oauth2-proxy → 3001`.

The break is one hop earlier. The `oauth2-proxy` namespace runs under `default-deny`, and `allow-auth-proxy` (the CiliumNetworkPolicy that enumerates auth-proxy's permitted **egress** upstreams) lists homepage, the KEDA interceptor, coroot, opencost and longhorn — **but not crossview**. So every `auth-proxy → crossview` SYN is dropped at egress.

Hubble on the source node, confirming the drop (not a mutual-auth failure — `authtype=-`):

```
oauth2-proxy/auth-proxy → crossview/crossview:3001  DROPPED  drop=POLICY_DENIED  (SYN)
```

And the empirical proof from inside the auth-proxy pod (coroot, which *is* in the allow-list, works from the same pod):

```
auth-proxy → crossview-service:80   → wget: download timed out   (→ Traefik 504)
auth-proxy → coroot:8080            → 200, serves HTML
```

crossview has been broken since it was added — its egress allowance was simply never created.

## Fix

Add the missing egress rule to `allow-auth-proxy`: `oauth2-proxy` (auth-proxy) → `crossview` namespace, **TCP 3001** (the backend pod port Cilium enforces on, matching the existing homepage:3000 / coroot:8080 / opencost:9090 entries). Inert on local/CI where the `crossview` namespace is absent.

## Validation

- `kubectl apply --dry-run=client` on the manifest: OK
- `kubectl kustomize k8s/clusters/local` and `k8s/clusters/prod`: build OK
- `kubectl kustomize k8s/providers/hetzner/infrastructure/controllers`: builds (6395 lines); rendered `allow-auth-proxy` contains the `namespace: crossview` egress on port `3001`

Not applied to the live cluster — GitOps reconciles it on merge.


devantler:

The **🧪 System Test** failure here looks like a flake, not a regression from this change — I've re-run the failed jobs.

Reasoning:
- This PR only **adds** one egress allow-rule to `allow-auth-proxy`, scoped to `toEndpoints` namespace `crossview` on TCP 3001. `crossview` is a **prod-only** app (the docker/local apps overlay is `resources: []`), so that namespace doesn't exist in the system test's local Talos+Docker cluster. A Cilium egress selector that matches no endpoint is inert — it can't deny traffic or fail policy validation.
- The run reached the live-reconcile phase (~34 min) before the step timed out, i.e. it passed all static/build validation; and the **same System Test passed on the sibling PR #2174** and other recent PRs, so the harness itself is healthy.
- Both overlays build locally (`kubectl kustomize k8s/clusters/{local,prod}`) and the rule renders in `k8s/providers/hetzner/infrastructure/controllers`.

If the re-run fails again deterministically, I'll dig into the Flux diagnosis from that run.

devantler:

**Confirmed: the two System Test failures were an etcd/runner infra flake, not this change.** I pulled the full job log — the local cluster's API server was overloaded and etcd requests timed out, which cascaded into KubeVirt and Kyverno control-plane failures and ultimately timed out the Flux health checks:

```
virt-operator … "reason":"etcdserver: request timed out"
error retrieving resource lock kubevirt/virt-operator: … context deadline exceeded
kyverno … Failed to update lease optimistically … etcdserver: request timed out
kyverno … failed to wait for cache sync
```

This change only **adds** one Cilium egress *allow* rule (`allow-auth-proxy` → `crossview` ns, TCP 3001) — it cannot cause etcd request timeouts, and `crossview` doesn't even exist in the local/docker overlay so the rule is inert there. Both failed runs also re-ran the *same* stale merge base; meanwhile the merge queue and other PRs pass on current `main`.

I've merged current `main` into this branch to rebuild on a fresh runner against the known-good tree. If the System Test flakes again here, it's the same KubeVirt/etcd resource pressure and is safe to re-run or admin-merge — the diff is build-clean (`kubectl kustomize` of both overlays + the hetzner controllers render the rule).

Issue #2208, opened 2026-06-22
incident(prod): Longhorn replica faults freezing the merge queue — every deploy-prod ejecting since ~08:55Z 06-22


devantler:
> 🤖 Generated by the Daily AI Assistant

## Summary

Since **~08:55Z on 2026-06-22**, the merge-queue **`🚀 Deploy to Prod`** health check has been failing
for **every** PR and ejecting it from the queue. The root signature is a **Longhorn storage-layer
incident** (faulted replicas + snapshot-purge i/o timeouts) that cascades into openbao and wedding-db,
timing out the `apps`/`wedding-app` Flux Kustomization health checks. **No PR can merge through the
queue until prod storage recovers.**

This is independent of any one PR's content — confirmed below — so I'm surfacing it as an incident
(I can't reach the prod cluster from the autonomous run to remediate).

## Evidence it's cluster-wide, not PR-specific

Merge-group `deploy-prod` results today (`ci.yaml`, `merge_group`):

| Time (UTC) | PR | Result |
|---|---|---|
| 08:24 | #2205 (dep bump) | ✅ success |
| 08:52 | #2206 (dep bump) | ✅ success |
| **08:55** | #2202 (renovate velero) | ❌ failure |
| 10:19 | #2202 | ❌ failure |
| 11:55 | #2197 (Tenant RGD) | ❌ failure |

**#2202 is a pure `renovate` Helm-release bump (velero v12.0.3) — it touches no storage/infra** — yet it
fails `deploy-prod` with the same storage errors as #2197. So the incident began ~08:55Z (after #2206
merged cleanly) and is environmental, not introduced by a PR.

## Failure signature (Longhorn → openbao → wedding-db)

From the failed `🚀 Deploy to Prod` logs (e.g. run 27945812861 for #2202, run 27950800649 for #2197):

```
longhorn-system  Warning FailedSnapshotPurge  volume/pvc-…  failed snapshot purge: … rpc error: code = Unavailable …
                 "transport: Error while dialing: dial tcp 10.244.21.245:10736: i/o timeout"
longhorn-system  Warning Faulted  engine/pvc-4bfb55df-…-e-0  Detected replica …-r-… (10.244.23.240:10513) in error
openbao          Warning FailedMount  pod/vault-config-…  MountVolume.SetUp failed for "pvc-4a2f858e-…":
                 rpc error: code = Internal desc = failed to publish volume: mount failed: exit status 255
[...]

PR #2197, opened 2026-06-21
feat(kro): add Tenant archetype ResourceGraphDefinition

devantler:
> 🤖 Generated by the Daily AI Assistant

## What

Defines the **KRO `Tenant` archetype** as a `ResourceGraphDefinition` — **step 2** of the maintainer's incremental plan for #1932 (step 1, the KRO controller install, landed in #2190). Onboarding a tenant becomes a ~5-line typed `Tenant` CR instead of the ~7-file control-plane skeleton each tenant hand-copies today (and which already drifts).

```yaml
apiVersion: kro.run/v1alpha1
kind: Tenant
metadata: { name: ascoachingogvaner }
spec:
  name: ascoachingogvaner
  externalDns: true   # tenant runs its own external-dns
  # sops: true        # tenant's Kustomization decrypts SOPS (e.g. wedding-app)
```

## Design (matches the ratified ADR — KRO decided 2026-06-19)

The RGD (`k8s/bases/infrastructure/archetypes/tenant/`) renders the full skeleton verbatim from today's bases:

- **Core (always):** namespace (PSA `restricted`), Flux-impersonated `edit` ServiceAccount + RoleBinding, `ghcr-auth` ExternalSecret (OpenBao `ClusterSecretStore`), cosign-verified `OCIRepository`, impersonating `Kustomization` (interval/timeout/retries per `enforce-flux-best-practices`), default-deny `NetworkPolicy`.
- **`includeWhen` toggles — the only real variations across today's tenants:**
  - `externalDns` → the three external-dns RBAC bindings (ascoaching's `simply.com` zone).
  - `sops` → emits the Kustomization's `decryption` block (wedding-app). Two Kustomization variants because `decryption` is a Kustomization-spec field, not a separable resource, and a `secretRef: sops-age` on a tenant with no `sops-age` Secret would wedge reconciliation. **The encrypted `sops-age` Secret stays applied separately — secret material is never inlined into a `Tenant` spec.**
- **Platform invariants templated/fixed**, not re-declared per tenant: the OCI URL (`oci://ghcr.io/devantler-tech/<name>/manifests`) and the cosign OIDC subject (the shared `publish-app` reusable workflow). Add an override field only if a future tenant diverges.

### RBAC (KRO runs `rbac.mode: aggregation`)

[...]

Issue #1933, opened 2026-06-09
Generalize the WebApp abstraction (ascoachingogvaner, wedding-app)


devantler:
## Summary

Generalize a **WebApp** archetype — a reusable definition of a containerized web application (workload + route + TLS + scaling + secrets) — so the tenant web apps (`ascoachingogvaner`, `wedding-app`) and future ones become a small per-app declaration instead of copied workload boilerplate.

Sibling to the **Tenant** archetype issue: a *Tenant* sets up the namespace/isolation/sync; a *WebApp* defines the application that runs inside it.

## Context

`ascoachingogvaner` (coaching site) and `wedding-app` are simple web apps. Their workload manifests (Deployment/container or chart, Service, Gateway-API `HTTPRoute`, optional scaling, optional secrets) live in each tenant's **own OCI artifact** (synced via the platform-provisioned `OCIRepository` + Flux `Kustomization`), not in this repo — but they follow the same web-app shape and will multiply as more tenants onboard. The shape is near-identical: image/chart + host + TLS + a route to the central gateway + (optionally) auth and a secret.

## Goal & constraint

One **WebApp** abstraction → a per-app instance specifying just `{image|chart, host, tls, scaling, auth?, secretRefs?}`. **No more Kustomize overlay/component stacking** (the readability reason we're choosing an archetype).

## Options to evaluate

1. **Helm app/library chart — recommended start.** A `webapp` chart (named templates for Namespace?/Deployment/Service/HTTPRoute/optional `HTTPScaledObject`/optional ExternalSecret) consumed via a HelmRelease or referenced from the tenant artifact. `values.schema.json` gives light typing; stays inside the Flux→HelmRelease model; zero new controllers; no pre-1.0 risk.
2. **KRO `WebApp` RGD.** Typed, `kubectl`-native, conditional children via `includeWhen` (KEDA/OIDC). But KRO is **pre-1.0 (`v1alpha1`)** and adds an alpha controller at the `infrastructure-controllers` chokepoint — see the KRO assessment. Better revisited once KRO hits beta/1.0.

## Open design question — placement

Tenant workloads live in **tenant-owned OCI artifacts**. Decide whether the WebApp archetype is:
- a **platform-published chart/RGD** that tenants reference from their artifact (platform owns the template, tenant owns the values), or
- defined per-tenant.

The platform-published option keeps the archetype DRY across tenants and lets us evolve it centrally; it should be the default unless there's a reason to fork per tenant.

[...]

Issue #1932, opened 2026-06-09
Generalize the Tenant abstraction — evaluate Capsule vs KRO vs Helm library chart


devantler:
## Summary

Generalize the **Tenant** archetype — the platform-side control-plane resources we hand-copy for every tenant — into a single reusable abstraction, so onboarding a tenant becomes one small declaration instead of a ~7-file copy. **Explicitly evaluate [projectcapsule/Capsule](https://github.com/projectcapsule/capsule) (buy) before building our own (KRO/Helm).**

This is the structural follow-up to the incremental DRY pass (#1927–#1931). Those squeezed the copy-paste out of the infra layer; the per-tenant plumbing is the remaining large archetype.

## Current state (what a Tenant abstraction would own)

Each tenant duplicates a near-identical control-plane skeleton. For `ascoachingogvaner`:

- `k8s/bases/apps/ascoachingogvaner/namespace.yaml`
- `…/serviceaccount.yaml` (the Flux-impersonated, namespace-scoped `edit` SA)
- `…/rolebinding.yaml`
- `…/sync.yaml` (cosign-verified `OCIRepository` + Flux `Kustomization` pointing at the tenant's own artifact)
- `…/ghcr-auth-externalsecret.yaml` (ESO → dockerconfigjson)
- `…/networkpolicy.yaml`
- `k8s/providers/hetzner/apps/ascoachingogvaner/patches/kustomization-patch.yaml`

`wedding-app` adds `secretstore.yaml` + `sops-age-secret.enc.yaml`. The two skeletons differ only by name/namespace/OCI URL and the SOPS block — and they already drift (e.g. inconsistent `app.kubernetes.io/managed-by: ksail` labels).

## Goal & constraint

One typed/abstracted **Tenant** → a ~5-line declaration per tenant. **Must not add more Kustomize overlay/component stacking** — "overlays on overlays on overlays" is hard to read without a visualization, which is the whole reason we're reaching for an archetype rather than another component.

## Options to evaluate

1. **Capsule (CNCF multi-tenancy operator) — buy, recommended to assess first.** Purpose-built: a `Tenant` CRD groups namespaces and auto-inherits ResourceQuota, LimitRange, NetworkPolicy, and RBAC; supports namespace auto-provisioning and `additionalMetadata` for traceability. Much more mature than KRO. **Open questions to resolve:**
   - Does its *owner-based namespace-grouping* model fit our mostly *single-namespace + Flux-impersonation* tenants, or is it overkill?
[...]

PR #2212, opened 2026-06-22
fix(kro): grant the Tenant controller watch on the tenants instance kind

devantler:

## Problem

After #1932's Tenant RGD (#2197) merged, `rgd/tenant.kro.run` came up **`state: Inactive`** with:

```
ControllerReady=False  FailedToStart
  add parent handler kro.run/v1alpha1, Resource=tenants: cache sync timeout for kro.run/v1alpha1, Resource=tenants
```

`GraphAccepted=True` / `KindReady=True` — the schema is fine. The failure is that KRO's **dynamic per-RGD micro-controller for `tenants` can't sync its informer**, because the controller SA has no `list/watch` on the instance kind (verified live):

```
kubectl auth can-i watch tenants.kro.run --as=system:serviceaccount:kro-system:kro  → no
kubectl auth can-i list  tenants.kro.run --as=system:serviceaccount:kro-system:kro  → no
```

`rbac.mode: aggregation` grants KRO only what the aggregated archetype roles declare. `kro-tenant-archetype` declared verbs on the Tenant's **child** kinds (namespaces, serviceaccounts, rolebindings, externalsecrets, networkpolicies, ocirepositories, kustomizations) but **never a rule for `tenants` itself** — and `kro:controller:static` only watches `resourcegraphdefinitions`/`graphrevisions`. So nothing lets the controller watch the new instance kind → informer cache-sync timeout → RGD stuck `Inactive`.

This matters operationally: the `healthCheckExprs` RGD gate (also added in #2197) makes the `infrastructure` Kustomization **wait** for the RGD to reach Active, so a stuck RGD blocks `infrastructure` → `apps` → the merge queue. Tracked live on #2208.

## Fix

Add the missing **instance-kind self-watch** to the `kro-tenant-archetype` aggregated ClusterRole: `get/list/watch/update/patch` on `tenants` (+ `tenants/status`, so the controller can write reconciled status back). Header comment updated to document that aggregation mode does **not** grant the instance-kind watch automatically — so every future archetype (the **#1933 WebApp RGD** especially) must ship its own self-watch rule.

## Validation

- Pure additive RBAC rule; `kubectl apply --dry-run=client` parses clean.
- The equivalent grant was applied live as a temporary standalone ClusterRole to unwedge the queue (see #2208); this PR folds it into the archetype role so the temp role can be deleted.

## Note on merge ordering

The queue is recovering from the #2208 incident. If the temporary unblock ClusterRole from #2208 is already applied, this is a no-op convergence; once merged, delete `kro-tenant-instance-watch`. `Refs #1932`.

### DIFF
--- a/k8s/bases/infrastructure/archetypes/tenant/aggregated-cluster-role.yaml
+++ b/k8s/bases/infrastructure/archetypes/tenant/aggregated-cluster-role.yaml
@@ -7,9 +7,17 @@
 # the verbs to manage the child kinds the Tenant RGD renders — nothing more — so
 # each archetype's blast radius is its own declared resources.
 #
-# Keep in lockstep with resource-graph-definition.yaml: a child kind added there
-# without a rule here makes KRO fail to reconcile that resource (forbidden);
-# a rule here with no corresponding child is dead over-grant.
+# Two distinct grants live here:
+#   1. The instance kind ITSELF (`tenants`) — KRO's dynamic micro-controller
+#      opens an informer on the kind, so it needs get/list/watch on it (and
+#      update/patch to write status back). Aggregation mode does NOT grant this
+#      automatically; without it the controller fails to start ("cache sync
+#      timeout for kro.run, Resource=tenants") and the RGD stays Inactive.
+#   2. The child kinds the RGD renders — so KRO can create/manage them.
+# Keep BOTH in lockstep with resource-graph-definition.yaml: a child kind added
+# there without a rule here makes KRO fail to reconcile that resource (forbidden);
+# a rule here with no corresponding child is dead over-grant. Every future
+# archetype (e.g. the #1933 WebApp RGD) needs its own instance-kind self-watch.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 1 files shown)


devantler:
Confirmed this is the **active blocker** for `infrastructure` → `apps` reconciliation on prod right now, and that this fix resolves it:

```
$ kubectl auth can-i list tenants.kro.run --as=system:serviceaccount:kro-system:kro
no
$ kubectl auth can-i list namespaces      --as=system:serviceaccount:kro-system:kro
yes        # aggregation works — only the parent kind was missing
```

Live state: `ResourceGraphDefinition/tenant.kro.run` is `state: Inactive`, `ControllerReady=False` (`cache sync timeout for kro.run, Resource=tenants`); `kustomization/infrastructure` is stuck `Reconciliation in progress` and `kustomization/apps` reports `dependency 'flux-system/infrastructure' is not ready`.

**It self-heals on merge** — the corrected role ships in the same revision the infra health check gates on, so the RGD goes Active as soon as Flux applies it; no operator action needed. Worth promoting + merging promptly since it currently blocks the whole apps layer.

(Closing my duplicate #2214 in favour of this one — it came first and the header note here is clearer. Only delta there was an explicit `tenants/finalizers` rule, which `update`/`patch` on `tenants` already covers.)

PR #2233, opened 2026-06-23
fix(unifi): persist the tf-runner pod to stop recurring "restarted" alerts

devantler:
## Problem

Coroot/Alertmanager fire recurring `Instance restarts - unifi/unifi-tf-runner application restarted N times` alerts — even though the unifi Terraform is `Ready` with no drift.

## Root cause (not a crash — runner recycling)

The unifi `Terraform` CR runs **observe-first / plan-only** (`approvePlan: ""`) on a **15m** interval. It doesn't set `alwaysCleanupRunnerPod`, so it inherits the controller default (**`true`**) — tofu-controller **creates and deletes** the `unifi-tf-runner` pod on every reconcile. Confirmed from events (`Created` → `Killing` every ~15m). Because the pod has a stable name, Coroot reads each recreation as a restart.

## Fix

Set `alwaysCleanupRunnerPod: false` so a single runner pod persists across the 15m plan cycles. The Terraform behaviour is unchanged (still plan-only/observe-first); the only difference is one small persistent runner pod (holding the read-only UniFi API key) instead of one created+destroyed every 15 minutes. This removes the restart churn at its source rather than muting the alert.

## Validation

- `kubectl kustomize k8s/providers/hetzner/apps/unifi/` renders `alwaysCleanupRunnerPod: false`.
- `ksail --config ksail.prod.yaml workload validate` → **✔ 358 files validated**.
- `kubectl kustomize k8s/clusters/prod/` and `k8s/clusters/local/` both build.

> Part of an investigation into the cluster's hourly Slack alert flood. Draft for maintainer review.

### DIFF
--- a/k8s/providers/hetzner/apps/unifi/terraform.yaml
+++ b/k8s/providers/hetzner/apps/unifi/terraform.yaml
@@ -22,6 +22,14 @@ spec:
   varsFrom:
     - kind: Secret
       name: unifi-credentials
+  # Keep the plan-runner pod alive between reconciles instead of recreating it
+  # every `interval`. With the controller default (true) the observe-first,
+  # plan-only loop creates + deletes the `unifi-tf-runner` pod every 15m, which
+  # Coroot reports as recurring "application restarted N times" alerts even
+  # though nothing is wrong (the Terraform stays Ready / no-drift — the runner
+  # is just being recycled, not crashing). A single persistent runner removes
+  # that churn at the cost of one small pod holding the read-only UniFi API key.
+  alwaysCleanupRunnerPod: false
   # Restricted-PSS-compliant runner. NB (live-validate): the upstream tf-runner
   # image may default to a root user; if it cannot satisfy restricted PSS, the
   # platform Kyverno mutate policies should supply the container securityContext


devantler:
Closing — you were right that `alwaysCleanupRunnerPod: false` is a workaround, not a root-cause fix, and it gives up the resource savings.

A deep trace of Coroot 1.22.2 + coroot-node-agent shows **why** the alert fires, and that there is **no in-repo manifest fix** that keeps `alwaysCleanupRunnerPod: true` and stops it:

- Coroot does **not** count container `restartCount` (always 0 here — each cycle is a brand-new pod, never an in-place restart). It counts **pod recreation**: the node-agent keys containers by a name-derived id `/k8s/unifi/unifi-tf-runner/tf-runner`; the next cycle's same-named pod hits the registry "id conflict" branch, reuses the `Container` struct, and `c.restarts++`.
- The runner is an **ownerless bare pod** → Coroot's constructor assigns `ApplicationKindStaticPods` (incident id `…:unifi:StaticPods:unifi-tf-runner`). `PeriodicJob()` is true **only** for `Job`/`CronJob`, so StaticPods gets restart-checked (threshold 1).
- Classification depends **solely** on the controller `ownerReference`. tofu-controller v0.16.4 creates the runner with **no** ownerReference and offers no Job-mode/ownerRef knob; the Terraform `runnerPodTemplate` exposes only `metadata.{labels,annotations}` (no `restartPolicy`, no `ownerReferences`). Coroot honors only 5 annotations (`custom_application_name`, `application_category`, 3× `slo_*`) — none flip `PeriodicJob()` or suppress the check.

So the **real fix is upstream**: tofu-controller should give the runner pod an ownerReference (or run it as a Job) so Coroot classifies it as ephemeral. Keeping `alwaysCleanupRunnerPod: true` is correct; I'll track the upstream ask separately rather than ship the cleanup=false workaround here.

Issue #2235, opened 2026-06-23
github-config: maintainers-team Crossplane resources fail (teamId parse + 403) → recurring CannotObserve/CannotCreate alerts


devantler:
## Symptom
A large share of the hourly Coroot/Slack alert flood is `Kubernetes events - CannotObserveExternalResource / CannotCreateExternalResource / UpdateFailed`. It also intermittently flaps the `github-config` Flux Kustomization health check (`HealthCheckFailed: timeout waiting for ... TeamRepository ... InProgress`).

## Root cause
The `maintainers`-team Crossplane managed resources never reconcile. Verified from live events in `github-config`:

1. **`teamId` is the literal string `"maintainers"`, not a numeric ID** →
   ```
   CannotObserveExternalResource  teamrepository/maintainers-platform
     failed to observe: strconv.ParseInt: parsing "maintainers": invalid syntax
     GET https://api.github.com/orgs/devantler-tech/teams/maintainers: 404 Not Found
   ```
2. **The GitHub App lacks org-owner / member-management permission** →
   ```
   CannotCreateExternalResource  teammembership/maintainers-devantler
     403 You must be an organization owner or team maintainer to add a team membership
   ```
3. The `Team` itself isn't created (the 404 above), so every dependent `TeamRepository`/`TeamMembership` loops on observe/create.

Note: the `issuelabels` / `repository` MRs are all `Ready=True/Synced=True` — only the **team / team-membership / team-repository** resources are broken.

## Proper fix (not in this repo)
These MRs are delivered by the external `devantler-tech/.github` OCI artifact (pulled via `k8s/bases/apps/github-config/sync.yaml`), so the fix is **not** a manifest in this repo:
1. In `devantler-tech/.github`: reference the team by **numeric ID** or a Crossplane `teamIdRef`/`teamIdSelector` to the `Team` MR, instead of the literal name `"maintainers"`, so the `Team` is created first and `TeamRepository`/`TeamMembership` resolve a valid ID.
2. Grant the provider-upjet-github **GitHub App** the org permission to manage teams/memberships (org owner, or the `members`/`administration` permission), to clear the `403`.

Until then this is the dominant non-Longhorn source of the hourly alert flood.

---
🤖 Filed during an investigation of the prod hourly alert flood (Claude Code).

devantler:
[...]

PR #2240, opened 2026-06-23
fix(github-config): wait: false so external GitHub failures don't degrade apps health

devantler:
## Problem

The `apps` and `github-config` Flux Kustomizations intermittently flap into `HealthCheckFailed` / `Unknown (Reconciliation in progress)` — real GitOps health degradation on the prod cluster.

## Root cause

`github-config` applies **external GitHub state** as Crossplane managed resources (Team / TeamRepository / TeamMembership, from the `devantler-tech/.github` OCI artifact). Their readiness depends on the **GitHub API + App credentials**, outside the platform's control. With **`wait: true`**, Flux gates this Kustomization's health on that external reconciliation — and because the parent `apps` Kustomization health-checks this one, a GitHub-side failure **cascades into the whole apps layer's GitOps health**.

It's happening now (see #2235): the `maintainers`-team resources never reconcile —
```
CannotObserveExternalResource ... strconv.ParseInt: parsing "maintainers": invalid syntax  (teamId is a name, not an id)
CannotCreateExternalResource  ... 403 You must be an organization owner or team maintainer  (App lacks org perms)
```
— so `wait: true` times out at 5m and flaps `apps`/`github-config` unhealthy.

## Fix

Set **`wait: false`** on the `github-config` Kustomization to decouple external-dependency reconciliation from in-cluster GitOps health. The apply still runs and apply errors are still surfaced — only the readiness *wait* is dropped, so a broken external resource is tracked via its Crossplane status / alerts (#2235) instead of degrading unrelated in-cluster apps.

- **Failure isolation, not masking.** github-config is the only app whose health depends on external-API reconciliation; every other app's health is internal (pods). Coupling the platform's GitOps health to GitHub-side state conflates two domains.
- **Does not fix the root cause** (#2235, external) — it stops the blast radius from including the whole apps layer.
- `github-config` is a leaf (nothing in-cluster depends on the GitHub state reconciling), so dropping the wait has no dependents.

## Validation

- `kubectl kustomize k8s/clusters/prod/` and `k8s/clusters/local/` both build; the rendered `github-config` Kustomization now has `wait: false`.

> Found during a fresh sweep of the prod platform. Draft for maintainer review (it's a judgment call on coupling external-dependency health to GitOps health).

### DIFF
--- a/k8s/bases/apps/github-config/sync.yaml
+++ b/k8s/bases/apps/github-config/sync.yaml
@@ -47,7 +47,21 @@ spec:
   retryInterval: 2m
   path: .
   prune: true
-  wait: true
+  # wait: false (NOT true like other apps). This Kustomization applies external
+  # GitHub state as Crossplane managed resources, whose readiness depends on the
+  # GitHub API + App credentials -- OUTSIDE the platform's control. With
+  # wait: true, Flux gates this Kustomization's health on that external
+  # reconciliation, and because this Kustomization is itself health-checked by
+  # the parent `apps` Flux Kustomization, any GitHub-side failure cascades into
+  # the WHOLE apps layer's GitOps health. That is currently happening: the
+  # maintainers-team resources fail forever (teamId references the team by name
+  # not id -> strconv.ParseInt error; the App lacks org-owner perms -> 403 -- a
+  # devantler-tech/.github + GitHub-App problem, tracked in #2235), so wait: true
+  # times out at 5m and flaps `apps` into HealthCheckFailed. The apply still
+  # runs and apply errors are still surfaced; only the readiness WAIT is dropped,
+  # so a broken EXTERNAL resource is tracked via its Crossplane status / alerts
+  # instead of degrading unrelated in-cluster apps' GitOps health.
+  wait: false
   force: true
   # Runs as the app SA: the GitHub managed resources are applied with only the
   # namespace-scoped authority in rbac.yaml. This Kustomization is itself applied


devantler:
Closing — you're right, \`wait: false\` treats the symptom. The real root cause is a **GitHub App permission**, not the Flux health gate.

**Definitive diagnosis:** the \`provider-upjet-github\` GitHub App (\`app_id 4085939\`) has **\`members: none\`** (every other org App has at least \`members: read\`):
\`\`\`
provider-upjet-github: members=none administration=write org_admin=write
\`\`\`
Without the org **Members** permission it cannot **read** the \`maintainers\` team (→ \`observe failed: external resource does not exist\`, on team \`17046606\` which does exist) or **write** team membership / repo access (→ \`403 You must be an organization owner or team maintainer\`).

The other layers were already handled:
- The numeric-id manifest bug is **already fixed** in \`devantler-tech/.github\` (#63, released v1.6.0, deployed) — the live MRs carry \`external-name: 17046606\`.
- A secondary **stuck provider Terraform state** (old \`"maintainers"\` ID from before #63) was cleared by recreating the observe-only MRs.

**Fix (org admin, GitHub App settings):** grant \`provider-upjet-github\` **Organization permissions → Members: Read and write**, then approve the new permission on the org installation. All 12 \`maintainers-*\` MRs then reconcile (observe + create) and the github-config/apps Flux health recovers — making \`wait: false\` unnecessary. Tracked in #2235.

PR #2248, opened 2026-06-23
docs(github-management): document the required Organization Members App permission

devantler:
## Why

The GitHub App setup step in `docs/github-management.md` listed *Repository: Administration/Contents* and *Organization: Administration* — but **not Organization: Members**. That's exactly the permission whose absence silently breaks the `maintainers` team management (#2235): the `provider-upjet-github` App had `members: none`, so it couldn't **read** the team (observe → `external resource does not exist`) or **write** membership/team-repo access (`403 You must be an organization owner or team maintainer`), while repo/ruleset/label resources kept working.

## Change

- Add **Organization → Members (read/write)** to the documented App permissions, with the failure-mode it causes if missed (`Administration` is not enough).
- Note that **adding** a permission to an *existing* App requires **approving the new permission on the org installation** before it takes effect — otherwise the resources stay broken even after editing the App.

Docs-only. No manifest change.

> Follow-up to the #2235 investigation (root cause = the App's missing Members permission, not a manifest or in-cluster issue).

### DIFF
--- a/docs/github-management.md
+++ b/docs/github-management.md
@@ -54,10 +54,24 @@ Velero), so the manually-set values are durable without a GitOps source of truth
 ## Credential setup (one-time, maintainer)
 
 1. Create a **GitHub App** on the devantler-tech org (Settings → Developer
-   settings → GitHub Apps): permissions *Repository: Administration
-   (read/write), Contents (read)* and *Organization: Administration
-   (read/write)* to start — widen as more resource kinds come under
-   management. Install it on **all repositories** of the org. No webhook.
+   settings → GitHub Apps) with these permissions, then install it on **all
+   repositories** of the org (no webhook):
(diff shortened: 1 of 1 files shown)


devantler:
Closing — superseded and redundant. \`main\` already documents this: the App-setup step now lists *Organization: Administration (read/write), **Members (read/write)*** with the maintainers-team failure modes (\`CannotObserveExternalResource\` / \`403 must be an organization owner or team maintainer\` / github-config Flux health flap). This branch was cut from an older HEAD, hence the conflict. The docs gap is already closed.

One small thing main's version doesn't call out, relevant to the immediate fix: when you **add** \`Members\` to the *existing* \`provider-upjet-github\` App, you must **approve the new permission on the org installation** (Org settings → Installed GitHub Apps → the App → review request) before it takes effect.

PR #2253, opened 2026-06-24
fix(backstage): wire the backend to Postgres via a generated appConfig

devantler:

## Problem (live on prod)

The `backstage/backstage` HelmRelease is stuck in a perpetual **install → 10m timeout → uninstall remediation** loop. Flux reports `Ready=Unknown` ("Running 'install' action with timeout of 10m0s"), and this blocks the `flux-system/apps` Kustomization from going `Ready` (it sits in health-check wait).

The backend pod is `0/1 Running`:
- **liveness** `…/health/v1/liveness` → `200`
- **readiness** `…/health/v1/readiness` → `503` (forever)

Pod logs show the real cause:

```
Plugin 'app' threw an error during startup …
Error: The SQLite driver does not support plugin division mode 'schema'
```

## Root cause

The upstream demo image (`ghcr.io/backstage/backstage:1.52.0`) ships **two** config files:

- `app-config.yaml` — hard-codes `client: better-sqlite3`, `connection: ':memory:'`
- `app-config.production.yaml` — the Postgres wiring: `client: pg` + `connection.{host,port,user,password}` from `${POSTGRES_*}`

The chart's default command is `node packages/backend` with **no `--config` args**, so only the SQLite `app-config.yaml` is loaded. The `POSTGRES_*` env vars the HelmRelease sets (from the CNPG `backstage-db-app` Secret) are referenced **only** by `app-config.production.yaml`, which was never loaded — so they were dead. With SQLite active, the `APP_CONFIG_backend_database_pluginDivisionMode=schema` override (a **Postgres-only** feature) throws and kills the process before readiness can pass.

The HelmRelease comment assumed *"the demo image's app-config already references these [POSTGRES_*]"* — but only the **production** config does, and it wasn't being loaded.

## Fix

Set `backstage.args` so the backend loads both files (base first, production last so it wins):

```yaml
args:
  - "--config"
  - "app-config.yaml"
  - "--config"
  - "app-config.production.yaml"
```

`app-config.production.yaml` selects `client: pg` and reads the **already-present** `POSTGRES_*` env vars. The public-URL `APP_CONFIG_*` env overrides still take final precedence over both files (env > file), so the `localhost:7007` defaults in the production file are overridden as before. No new secrets or values are introduced — this only makes the existing Postgres wiring take effect.

## Validation

- `helm template backstage/backstage --version 2.8.2` with the new args renders the container as `node packages/backend --config app-config.yaml --config app-config.production.yaml`.
- `kubectl kustomize k8s/providers/hetzner/apps` (prod) → builds (212 resources), backstage args present.
- `kubectl kustomize k8s/clusters/prod/` and `k8s/clusters/local/` → build clean.
- Change is isolated to the backstage base HelmRelease (`+19/-3`, one file).

## Scope / notes

- Backstage remains the documented **scaffold** (demo image, guest auth behind forward-auth SSO at the gateway). This PR only makes the install succeed on Postgres; productionising via a custom-built image is unchanged future work (see `docs/backstage.md`).
- Draft per the engineering contract — ready for promotion; CI green + threads resolved before hand-off.

### DIFF
--- a/k8s/bases/apps/backstage/helm-release.yaml
+++ b/k8s/bases/apps/backstage/helm-release.yaml
@@ -65,9 +65,15 @@ spec:
       # Generated ConfigMap: the chart auto-mounts it and appends a second
       # `--config /app/app-config-from-configmap.yaml`. Overrides only the database
       # block — replaces app-config.yaml's in-memory SQLite (client:
-      # better-sqlite3, connection: ':memory:') with Postgres. ${POSTGRES_*} are
-      # substituted at load time from the extraEnvVars below (sourced from the CNPG
-      # backstage-db-app Secret), so no secret values land in the ConfigMap itself.
+      # better-sqlite3, connection: ':memory:') with Postgres. This establishes
+      # `backend.database.connection` as an OBJECT (so reading it never hits the
+      # in-memory string); the connection's secret values (host/port/user/password)
+      # are layered on top from the APP_CONFIG_* env vars below rather than written
+      # here. NB: we must NOT put `${POSTGRES_*}` here — Flux's Kustomize postBuild
+      # substitution runs on this manifest at deploy time and, since those are not
+      # Flux variables, would replace them with empty/null before Backstage ever
+      # sees them (the image's app-config.production.yaml gets away with ${…}
+      # because it lives inside the container image, which Flux never touches).
       appConfig:
         backend:
           database:
@@ -77,10 +83,6 @@ spec:
             # CREATEDB on the role.
             pluginDivisionMode: schema
             connection:
-              host: ${POSTGRES_HOST}
-              port: ${POSTGRES_PORT}
-              user: ${POSTGRES_USER}
-              password: ${POSTGRES_PASSWORD}
               database: backstage
 
       # Single pod: Backstage runs plugin DB migrations on startup, so two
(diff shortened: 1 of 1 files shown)


devantler:

## Follow-up: the first fix was necessary but not sufficient — live root-cause found

After this PR was promoted and queued, its `merge_group` **Deploy to Prod** run kept failing: the `apps` Kustomization timed out on `HelmRelease/backstage` staying `InProgress`/`Failed`. I diagnosed it on the live cluster (`admin@prod`):

- The original commit's `--config app-config.yaml --config app-config.production.yaml` **was applied** (deploy args confirmed live), and Postgres (`backstage-db-1/2`) is healthy — yet the backend stayed in `CrashLoopBackOff`.
- The crash, from the pod log:
  ```
  Unhandled rejection Invalid type in config for key 'backend.listen' in
  'app-config.production.yaml', got string, wanted object
  ```
- Root cause: the demo image's bundled `app-config.production.yaml` sets `backend.listen: ':7007'` (a **string**), which overrides `app-config.yaml`'s object form (`listen: {port: 7007}`). `readHttpListenOptions` wants an object → the backend dies on startup. So loading that whole file to get Postgres also imported its broken `listen`.

### Corrected approach (this push)
Stop loading `app-config.production.yaml`. Instead load `app-config.yaml` **plus a small chart-generated `appConfig`** (`backstage.appConfig`) that overrides **only** `backend.database` → Postgres (`client: pg`, connection from the CNPG `backstage-db-app` secret, `pluginDivisionMode: schema`). `app-config.yaml`'s correct `listen` object stands, so the backend starts. The now-redundant `APP_CONFIG_backend_database_*` env vars were folded into the appConfig so the whole `backend.database` block has a single source (no string/object merge ambiguity).

### Validation
- `helm template` of the real HelmRelease values renders `args: [--config app-config.yaml, --config /app/app-config-from-configmap.yaml]` and a ConfigMap carrying only the `database` block — no `backend.listen`.
- `ksail --config ksail.prod.yaml workload validate` → `✔ bases/apps/backstage validated`. (The unrelated `actual-budget`/`login/openid` schema failure pre-exists on `main` and is CI-tolerated.)
- CNPG secret keys verified live: `host=backstage-db-rw`, `user=backstage`, `dbname=backstage`.

This PR is deploy-gated through the merge queue, so it now only merges once backstage actually reaches Ready in prod.

devantler:

## Second live finding: Flux postBuild substitution was nullifying the connection

The previous push (appConfig) was **structurally correct** — verified live: the backend started cleanly (no more `backend.listen` crash), initialized all plugins, and served the liveness probe — but it then failed at **`Failed to connect to the database`**. The rendered ConfigMap on the cluster showed:

```yaml
backend:
  database:
    client: pg
    connection:
      database: backstage
      host: null   # ← ${POSTGRES_HOST} eaten by Flux
      password: null
      port: null
      user: null
    pluginDivisionMode: schema
```

Root cause: the `${POSTGRES_*}` tokens I put in the appConfig are processed by **Flux's Kustomize postBuild substitution** at deploy time. They aren't Flux substitution variables, so Flux replaced them with empty → `null`. (The image's bundled `app-config.production.yaml` gets away with `${…}` only because it lives *inside the container image*, which Flux never touches.)

**Fix (this push, 628a416):** source the connection's secret values as `APP_CONFIG_backend_database_connection_{host,port,user,password}` env vars (`valueFrom` the CNPG `backstage-db-app` secret). Env values aren't run through Flux substitution, and Backstage reads `APP_CONFIG_*` as highest-precedence config overrides that layer onto the `connection` object the appConfig still defines (`client: pg`, `connection.database`, `pluginDivisionMode: schema`). No `${}` tokens remain in the manifest.

Re-validated `ksail … workload validate` ✔ backstage. Re-queuing once PR CI is green; the merge-queue deploy gate is the final check against live prod.

PR #2309, opened 2026-06-27
fix(rbac): scope headlamp SA to cluster-reader, fix C-0187 wildcard exceptions

devantler:
## What

Reduce wildcard / cluster-admin RBAC flagged by Kubescape **CIS-5.1.3 (C-0187)** — *Minimize wildcard use in Roles and ClusterRoles*.

- **Headlamp (root-cause fix)** — replace the chart's default `cluster-admin` ClusterRoleBinding with a read-only binding to `cluster-reader`. Headlamp authenticates every request with the logged-in user's OIDC token (`config.unsafeUseServiceAccountToken: false`), so the pod SA is not used for API calls and does not need cluster-admin. Implemented as `clusterRoleBinding.create: false` + a separate `headlamp-cluster-reader` binding, **not** by setting `clusterRoleName` — a ClusterRoleBinding's `roleRef` is immutable, so retargeting the existing binding in place fails the Helm upgrade.
- **Exceptions (mechanism fix)** — the C-0187 entry in `controller-rbac.yaml` used a `namespaceSelector`, which does **not** suppress RBAC findings: Kubescape keys C-0187 on the Role/ClusterRole/binding object, so a namespace match never applies (proven — the namespaced `velero-server` Role kept failing C-0187 even though `velero` was in the selector). Removed that dead entry and added `wildcard-rbac.yaml`, which matches the by-design holders by **kind+name**:
  - **Flux** GitOps reconcilers (`cluster-reconciler-flux-system`, `flux-operator`) — cluster-admin is the GitOps trust model (applies arbitrary manifests cluster-wide).
  - **Velero** (cluster-admin ClusterRoleBinding + namespaced `velero-server` Role) — the chart's RBAC is all-or-nothing and restore must write arbitrary kinds.

  The match is deliberately scoped to those bindings so C-0187 still flags any **new** accidental cluster-admin grant. `longhorn-support-bundle` (a transient diagnostic SA) is intentionally left unexempted.

## Why

C-0187 is a CIS control — **not** part of the NSA CI gate, so this is non-gating — but reducing standing cluster-admin is good hygiene. Headlamp was the one avoidable god-role; Flux and Velero require broad RBAC by design and are now covered by an exception that actually works (the previous `namespaceSelector` one was a silent no-op).

## Validation (static, no cluster mutated)

- `kubectl kustomize` builds clean for `clusters/local`, `clusters/prod`, and the `providers/{hetzner,docker}/{apps,infrastructure}` roots.
- Headlamp: chart renders **0** ClusterRoleBindings with `create: false` (Helm deletes the old `headlamp-admin`), and the new `headlamp-cluster-reader` binding subjects the `headlamp` SA to `cluster-reader` — no `*/*/*`.
- New `ClusterSecurityException` is schema-valid against the live CRD (`kubectl apply --dry-run=server`).
- Scoped exemption match verified read-only via `kubescape scan --exceptions`: C-0187 drops from 7 to exactly the Flux/Velero relationships suppressed, others retained.

> **v1 rollout fix:** the first revision set `clusterRoleName: cluster-reader`, which retargets the existing binding's immutable `roleRef` (`cannot change roleRef`) and wedged the Helm upgrade in the merge queue (Flux rolled back to cluster-admin). Fixed by recreating the binding under a new name.

> Post-merge check: confirm the in-cluster Kubescape operator dashboard shows C-0187 cleared for Flux/Velero. The operator applies `ClusterSecurityException` CRs additively (unlike the CLI `--exceptions` proxy used here), so the built-in `system:masters` exemption is preserved.

### DIFF
--- a/k8s/bases/apps/headlamp/clusterrolebinding.yaml
+++ b/k8s/bases/apps/headlamp/clusterrolebinding.yaml
@@ -1,0 +1,27 @@
+---
+# Read-only cluster RBAC for the Headlamp backend ServiceAccount. Replaces the
+# chart's default cluster-admin ClusterRoleBinding, which is disabled in
+# helm-release.yaml (clusterRoleBinding.create: false). Headlamp authenticates
+# user requests with the logged-in user's OIDC token, so its ServiceAccount only
+# needs the platform's read-only cluster-reader role — not a `*/*/*` god-role
+# (Kubescape CIS-5.1.3 / C-0187).
+#
+# This is a separate, differently-named binding (not headlamp-admin) on purpose:
+# a ClusterRoleBinding's roleRef is immutable, so the existing cluster-admin
+# binding cannot be retargeted in place. Helm removes the old binding and this
+# one is created fresh, avoiding the "cannot change roleRef" upgrade failure.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: headlamp-cluster-reader
+  labels:
+    app.kubernetes.io/name: headlamp
+    app.kubernetes.io/managed-by: ksail
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-reader
+subjects:
+  - kind: ServiceAccount
+    name: headlamp
+    namespace: headlamp
--- a/k8s/bases/apps/headlamp/helm-release.yaml
+++ b/k8s/bases/apps/headlamp/helm-release.yaml
@@ -121,13 +121,18 @@ spec:
     # cluster-admin, but Headlamp runs with OIDC (config.oidc below;
     # unsafeUseServiceAccountToken defaults to false), so every Kubernetes API
     # request is authenticated with the logged-in USER's token — the pod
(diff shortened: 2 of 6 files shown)


devantler:
Diagnosed the failed merge-queue rollout. It wedged on headlamp's Helm upgrade — helm rev 31/33:

```
Upgrade "headlamp" failed: server-side apply failed for object /headlamp-admin
... ClusterRoleBinding "headlamp-admin" is invalid: roleRef: Invalid value:
{...Name:"cluster-reader"}: cannot change roleRef
```

A ClusterRoleBinding's `roleRef` is immutable, so setting `clusterRoleName: cluster-reader` tried to mutate the existing binding in place and failed; Flux remediation thrashed (upgrade→rollback ×2) and settled back on cluster-admin (rev 34).

Fixed in 44a5ebe1: `clusterRoleBinding.create: false` (Helm *deletes* the old binding — deletion is allowed, only `roleRef` updates are blocked) plus a separate, differently-named `headlamp-cluster-reader` binding to `cluster-reader`. Same least-privilege end state, no immutable conflict. Verified the chart renders 0 ClusterRoleBindings with `create: false` and all overlays build. Flux should self-heal headlamp on the next reconcile after merge.

PR #2344, opened 2026-06-29
fix(security-context): pair pod-level runAsUser with injected runAsGroup (#2292 regression)

devantler:
## Problem

`homepage-primary` pods fail to start with:

```
FailedCreatePodSandbox: ... failed to generate user string: user group "1000" is specified without user
```

The `add-security-context` Kyverno ClusterPolicy injects **pod-level** `runAsGroup: 1000` but no `runAsUser`. A pod-level group with no user produces the OCI user string `":1000"`, which containerd rejects when building the **pod sandbox** (the `pause` container, which uses the pod-level `securityContext`). homepage's chart sets `runAsUser` only at the **container** level, so its pod-level sandbox had a group and no user → every new pod is stuck in `ContainerCreating` (primary degraded to 1/3).

This is the #2292/#2296 design: `runAsGroup` was injected pod-level for C-0013/C-0211, the same sandbox failure was hit on tofu-controller's `tf-runner`, and **only tofu-controller was excluded**. `cert-manager`, `keda` and `opencost` are the same latent landmine — their currently-running pods predate the 2026-06-27 policy rollout and would fail identically on the next restart.

## Fix

- **Inject a matching pod-level `runAsUser: 1000`** alongside `runAsGroup: 1000` in `add-pod-security-context` (both via `+()` conditional anchors, so any chart/container that sets its own value still wins). The pod-level user+group pair is now always complete, so the sandbox builds. This is the more-hardened default.
- **Exempt images with a baked-in non-1000 UID and no chart `runAsUser`** — `cert-manager`/`trust-manager`/`origin-ca-issuer` (65532), `keda` (65532), `opencost` (1001) — from the forced uid/gid. Forcing them to 1000 would override their proven image UID on critical infra. They are handled by `add-imageuid-pod-security-context` (generalized from the former tofu-only `add-tofu-runner-pod-security-context` rule), which supplies pod-level `seccompProfile` + `runAsNonRoot` + `fsGroup` + `fsGroupChangePolicy` but **neither `runAsUser` nor `runAsGroup`** — so they keep their image UID and the sandbox still builds (no group without user).

Image UIDs confirmed via `crane config`. Namespace exemptions, not container changes, since these rely purely on the image `USER`.

## Validation

`kustomize build` of the cluster-policies base renders clean. `kyverno apply` (CLI 1.18.1) against pod fixtures:

| Pod | Pod-level result |
|---|---|
| homepage (`homepage` ns, container `runAsUser:1000`) | `runAsUser:1000` + `runAsGroup:1000` + fsGroup/seccomp/runAsNonRoot ✅ |
| cert-manager (`cert-manager` ns, no `runAsUser`) | fsGroup/seccomp/runAsNonRoot, **no runAsUser/runAsGroup** ✅ |
| generic (in-scope ns, no `runAsUser`) | `runAsUser:1000` + `runAsGroup:1000` ✅ |

Compliance: only *adds* securityContext fields (no kubescape NSA/CIS gate regression); the exempted namespaces lose only the pod-level non-root *group* control (C-0211, CIS — already covered by the `pod-security-mutations` ClusterSecurityException), not anything in the NSA CI gate.

### DIFF
--- a/k8s/bases/infrastructure/cluster-policies/best-practices/add-security-context.yaml
+++ b/k8s/bases/infrastructure/cluster-policies/best-practices/add-security-context.yaml
@@ -4,20 +4,32 @@
 # - readOnlyRootFilesystem: true (C-0017)
 # - drop ALL capabilities (C-0055)
 # - runAsNonRoot: true (C-0013) — pod + container
-# - runAsGroup: 1000 (C-0013/C-0211 non-root group) — pod level
+# - runAsUser: 1000 + runAsGroup: 1000 (C-0013/C-0211 non-root user+group) — pod level
 # - fsGroup: 1000 + fsGroupChangePolicy: OnRootMismatch (C-0211) — pod level
 # - seLinuxOptions: {level: s0} (C-0211 CIS-5.7.3) — containers + initContainers
 #
 # Together these complete the security context kubescape's CIS-5.7.3 (C-0211) checks
 # (non-root-containers, set-fsgroup-value, set-fsgroupchangepolicy-value,
 # set-seLinuxOptions). Why each is safe to inject cluster-wide:
-# - runAsGroup is injected at the POD level only (never the container level), so it
-#   sets the process primary GID without clobbering a container-level value a chart
-#   sets. Pod-level runAsGroup alone satisfies the non-root-containers rule — no
-#   runAsUser needed (verified with kubescape).
-# - runAsUser is deliberately NOT injected: it would override image-baked UIDs and
-#   break charts (cert-manager, distroless images, …). The non-root rule is already
-#   satisfied by runAsGroup, so runAsUser is unnecessary here.
+# - runAsUser and runAsGroup are injected TOGETHER at the POD level, and only there.
+#   A pod-level runAsGroup with NO runAsUser yields the OCI user string ":1000",
+#   which containerd rejects ("user group \"1000\" is specified without user") so the
+#   pod sandbox never starts. Injecting runAsGroup alone did exactly that to homepage
+#   (its chart sets runAsUser only at the CONTAINER level) and to tofu-controller's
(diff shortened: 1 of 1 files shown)


devantler:

## Root-cause of the live prod wedge — and why this PR was the fix that never landed

Prod is currently wedged on this exact bug. `homepage-primary` has been stuck in `ContainerCreating` for ~17h:

```
FailedCreatePodSandBox (x4815 over 17h): failed to generate user string:
user group "1000" is specified without user
```

Confirmed the pod's **pod-level** securityContext has `runAsGroup:1000` with **no** `runAsUser` (homepage's chart sets `runAsUser` only at the container level), which yields the OCI user string `:1000` that containerd rejects — exactly the regression this PR fixes. All Flux Kustomizations/HelmReleases report `Ready=True`, so the wedge is invisible at the reconcile layer; it only shows as a sandbox that never starts.

This PR (the correct fix) had simply **never merged** — that's why prod stayed broken.

### What I changed
The branch was 20h stale (forked at `54590b82`, behind #2343, #2347, #2349, #2353, #2355). Its diff against `main` therefore *appeared* to revert several merged fixes — notably #2347 (crossplane Upbound egress) and #2349 (actual-budget envsubst opt-out). The branch's own single commit only ever touched `add-security-context.yaml`; the "reverts" were pure behind-ness.

I refreshed it by **merging `origin/main` in** (no force-push, fast-forward only). The net diff vs `main` is now **only** `add-security-context.yaml`. Both overlays build (`kubectl kustomize k8s/clusters/{local,prod}`) and the policy passes a server-side dry-run against the live cluster's Kyverno webhook.

Auto-merge (squash) is already enabled, so this will now carry through the merge queue and deploy to prod. **Note:** the existing stuck `homepage-primary` pod won't self-heal after deploy — it was admitted with the broken securityContext baked in. It needs a one-time delete so its ReplicaSet recreates it under the corrected policy; I'll clear it once the policy is live on the cluster.

Issue #2337, opened 2026-06-29
fix: harden merge-queue deploy-prod verification against Cilium mutual-auth first-packet drop (queue wedges)


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

The `CI` workflow's **`🚀 Deploy to Prod` job on the `merge_group` event** fails intermittently, which **ejects PRs from the merge queue and freezes GitOps delivery**. Observed 2026-06-29: 4 of the last 7 `merge_group` `ci.yaml` runs failed (e.g. runs `28369027060` 11:34Z, `28364276552` 10:04Z, `28360750235` 09:02Z, `28358116304` 08:14Z), while `28360718256` 09:02Z **succeeded** — so the queue drains only on a lucky batch, and trusted PRs (#2333/#2335/#2336 this morning) repeatedly dequeue and have to be re-driven by hand.

## Root cause

The failures are the **inline post-deploy provisioning steps** inside `deploy-prod` hitting the known **Cilium 1.20.0-pre.3 mutual-auth first-packet drop** (#2272, upstream cilium#46767) on cold/expired-auth connections. From the failed `deploy-prod` log:

- `umami-provision-tenants`: `Error: list teams failed: 500` → `could not authenticate to umami after retries; last login failure: ...`
- `coroot-custom-cloud-pricing`: `ERROR: set custom cloud pricing failed (HTTP 000):` and `ERROR: 'platform' project not found after retries`
- `Pods Statuses: 0 Active (0 Ready) / 0 Succeeded / 7 Failed`

These are exactly the workloads memory/#2272 document as suffering the mutual-auth first-packet drop (the ~5s `mesh-auth-mutual-connect-timeout` burned on each cold connection). The in-step retries are not generous enough to absorb that on a cold path, so the **gating** verification fails even though the underlying CronJobs are retry-hardened and self-heal afterwards.

## Impact

- Merge queue wedges → trusted/own PRs cannot merge without repeated manual re-enqueue.
- The flakiness is **delivery-blocking**, distinct from the (by-design, self-healing) lingering Error pods in prod — the new harm is the queue eviction, not the prod history.

## Proposed direction

The root-cause Cilium fix is **upstream-blocked** (#2272 / cilium#46767), so this issue is a **mitigation to unwedge delivery while that lands**. Options (prefer the last):

1. **Increase retry tolerance / timeout** for the inline umami + coroot provisioning calls so they absorb the ~5s first-packet drop (more attempts / longer backoff window).
[...]

PR #2349, opened 2026-06-29
fix(actual-budget): exempt enablebanking-seed ConfigMap from Flux envsubst

devantler:

## Problem (live prod — surfaced during rollout)

With the infrastructure→apps stall cleared by #2343, the `apps` Kustomization finally reconciled past its dependency gate and **immediately failed**:

```
post build failed for 'ConfigMap.v1/actual-budget-enablebanking-seed':
```

This blocks the **entire `apps` Kustomization** (all apps).

## Root cause

The `actual-budget-enablebanking-seed` ConfigMap (added in [#2331](https://github.com/devantler-tech/platform/pull/2331)) embeds a Node.js reconcile script using **JS template literals** — `${DB_PATH}`, `${name}`, and crucially a *multi-line* one:

```js
`reconciling Enable Banking secrets into ${DB_PATH} every ${
  INTERVAL / 1000
}s`
```

The `apps` Kustomization runs `substituteFrom`, so Flux's `envsubst` tries to expand every `${...}`. The newline inside `${ ... }` can't be parsed as a variable name → the whole Kustomization fails.

**Why it was latent:** `apps` never reconciled after #2331 merged because it was blocked on the `infrastructure` dependency (the OpenBao policy 403, #2343). The error only surfaced once `infrastructure` went healthy during this rollout. So #2331 shipped *two* coupled defects — the policy gap (#2343) and this envsubst break — the first masking the second.

## Fix

These are not Flux variables; the ConfigMap needs no post-build substitution. Opt it out with `kustomize.toolkit.fluxcd.io/substitute: disabled` — the same pattern the KRO `ResourceGraphDefinition`s already use for their CEL `${...}` expressions.

## Validation

- `kubectl kustomize k8s/bases/apps/actual-budget/` ✅ (rendered ConfigMap carries the annotation)
- `kubectl apply --dry-run=client` ✅

## Rollout note

`infrastructure` is now **Ready** (the #2343 fix applied cleanly after a forced reconcile — the OpenBao policy updated, the seed PushSecret synced). `apps` will go green once this merges and deploys.

### DIFF
--- a/k8s/bases/apps/actual-budget/config-map.yaml
+++ b/k8s/bases/apps/actual-budget/config-map.yaml
@@ -3,6 +3,16 @@ kind: ConfigMap
 metadata:
   name: actual-budget-enablebanking-seed
   namespace: actual-budget
+  annotations:
+    # The embedded Node.js script below uses JS template literals (`${DB_PATH}`,
+    # `${name}`, and a multi-line `${ INTERVAL / 1000 }`) that are NOT Flux
+    # post-build variables. The apps Kustomization runs substituteFrom, so Flux's
+    # envsubst tries to expand them and fails to parse the multi-line one
+    # ("envsubst error: variable substitution failed: unable to parse variable
+    # name"), which blocks the entire apps Kustomization. This ConfigMap needs no
+    # Flux substitution, so opt it out — same pattern as the KRO
+    # ResourceGraphDefinitions whose `${...}` are CEL expressions.
+    kustomize.toolkit.fluxcd.io/substitute: disabled
 data:
   # Reconciles Actual Budget's Enable Banking credentials (Application ID + PEM
   # secret key) into the sync-server's internal `secrets` SQLite table from an


devantler:

## This PR is deadlocked in the merge queue — it can't self-merge, root cause below

Heads-up: this PR is promoted + `mergeStateStatus: CLEAN`, but it **will not merge through the queue on its own**. Its `merge_group` run was **evicted** ([run 28423415257](https://github.com/devantler-tech/platform/actions/runs/28423415257), 05:52Z) — and every re-queue will fail the same way. It's a circular dependency, not a transient flake:

1. **`apps` is live `Ready=False (BuildFailed)`** — `post build failed for ConfigMap/actual-budget-enablebanking-seed: envsubst error: unable to parse variable name`. That is exactly the wedge **this PR fixes** (git `main` still lacks the `substitute: disabled` annotation).
2. Because `apps` is frozen, **coroot is entirely absent** (`kubectl get pods -n coroot` → no resources) and **umami is degraded** (`list teams failed: 500`).
3. This PR's `merge_group` deploy runs `deploy-prod`, whose **inline umami/coroot provisioning verify** then hard-fails (umami auth 500, coroot `'platform' project not found after retries`, 7 failed provisioning pods) → `deploy-prod` fails → `CI - Required Checks` fails → **the queue evicts this PR**.
4. So the services this PR is meant to rescue must be healthy for it to merge — but they can't be healthy until it merges. **Circular.**

(Note: a prior deploy already applied `substitute: disabled` to the **live** ConfigMap object, but Flux rebuilds `apps` from git `main` — which lacks it — so the build keeps failing regardless.)

### Deterministic break

The queue can't resolve this; it needs a one-shot break by you:
- **Admin-merge this PR** (a CLEAN, promoted, trusted-author fix) → git `main` gets the annotation → `apps` builds → reconciles → coroot/umami recover → the queue is healthy again. *(I can't `--admin`/bypass per my guardrails — this is your call.)*
- Or recover coroot/umami manually first, then a clean re-queue passes.

I have **not** re-queued this PR (re-queuing is futile until the cycle is broken — per the root-cause-before-requeue rule). `#2337` (decouple non-critical provisioning from the gating verify) was the structural fix for step 3, but you closed it `NOT_PLANNED`, so I'm surfacing the live deadlock rather than re-opening that approach.

PR #2345, opened 2026-06-29
fix(openbao): grant actual-budget the missing vault ACL paths

devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

Since [#2331](https://github.com/devantler-tech/platform/pull/2331) merged (~07:00Z 2026-06-29), the `seed-actual-budget-enablebanking` PushSecret has been **Errored** on a clean 403:

```
set secret failed: could not verify if secret exists in store: cannot read secret data from Vault:
URL: GET http://openbao-active.openbao.svc.cluster.local:8200/v1/secret/data/apps/actual-budget/enablebanking
Code: 403. * permission denied
```

#2331 added the Enable Banking seed PushSecret (`infrastructure/vault-seed/push-secret-seed-actual-budget-enablebanking.yaml`) and the app's `actual-budget-enablebanking` ExternalSecret, but **never granted the matching OpenBao ACL paths** in `infrastructure/vault-config/job.yaml`. So:

- the **seed** can't write `secret/data/apps/actual-budget/enablebanking` (the live 403 above — `vault-seed-write` has no `actual-budget` path), and
- the **app read** would 403 too once seeded (there is no `app-actual-budget-readonly` policy on the shared `external-secrets` role bundle).

This is the persistent, non-flake failure surfacing in the merge-queue `deploy-prod` HealthCheck wait (distinct from the Cilium first-packet-drop noise tracked by #2337).

## Fix

Three **additive** grants, mirroring the existing umami/crossview apps:

1. New `app-actual-budget-readonly` policy — `read` on `secret/data/apps/actual-budget/*`.
2. `actual-budget` data + metadata write paths on `vault-seed-write` — so the seed PushSecret can write the placeholder.
3. `app-actual-budget-readonly` added to the shared `external-secrets` Kubernetes-auth role bundle — so the app's ExternalSecret can read it back.

All three are additive allow-list entries; no existing access is altered.

## Validation

- `ksail workload validate --skip-helm-render` — **474 files, exit 0** (both local + prod overlays).
- Embedded heredocs balanced; policy blocks copied verbatim from the umami/crossview pattern.

## Blast radius / promotion

Touches prod OpenBao ACLs (additive only). On promotion the `vault-config` Job re-runs and applies the new policies; the seed PushSecret should then sync and the app ExternalSecret resolve. Left as a **draft** for the maintainer to promote and watch the first reconcile.

Fixes the actual-budget 403 introduced by #2331.

PR #2351, opened 2026-06-29
fix(actual-budget): disable Flux postBuild substitution on enablebanking seed

### LINKED ISSUE #2026: Merge-queue deploys leave prod running unmerged code when the merge group fails

## What happened today (2026-06-11)

PR #1991 entered the merge queue. `ci.yaml`'s `deploy-prod` job (which runs on `merge_group`, i.e. on the **speculative merge ref, before the PR actually merges**) pushed the OCI artifact and reconciled prod at ~16:08. The merge group then failed → the PR was ejected, **but prod kept reconciling the unmerged artifact**: the kyverno HelmRelease picked up #1991's then-broken PDB values, the upgrade failed every interval (`Cannot set both .minAvailable and .maxUnavailable`), and the whole `infrastructure-controllers → infrastructure → apps` chain sat blocked for hours. Nothing on `main` reflected the state prod was running.

## The general flaw

`deploy-prod` on `merge_group` is a deliberate and useful gate (a deploy that fails prevents the merge). The gap is the **failure path**: any merge-group failure *after* `workload push` leaves `latest` pointing at code that never landed on `main`, until some future merge overwrites it. Prod state silently diverges from Git — the core thing GitOps is supposed to prevent.

## Options (maintainer decision)

1. **Heal on failure** (keeps the gate): add an `if: failure()` job/step to the merge-group run that re-checkouts `main` and re-runs `ksail --config ksail.prod.yaml workload push` + `reconcile`, so an ejected PR's artifact is immediately replaced by main's.
2. **Tag/push artifacts by SHA + promote on merge**: push the merge-group artifact under a non-`latest` tag, and only re-tag to `latest` after the merge actually completes (needs a small post-merge workflow).
3. **Accept the risk** (status quo): the next successful merge self-heals; document the failure mode in AGENTS.md / the DR runbook so the on-call knows that a red merge queue can mean prod is running unmerged code.

Option 1 is the smallest change that closes the gap without giving up the pre-merge deploy gate. Happy to PR whichever direction is preferred.

devantler:

## Problem — `apps` Kustomization is `BuildFailed` on live prod (and wedges the merge queue)

The `apps` Flux Kustomization on **prod is `Ready=False` right now**:

```
post build failed for 'ConfigMap.v1.[noGrp]/actual-budget-enablebanking-seed':
```

`actual-budget-enablebanking-seed` embeds a JavaScript payload (`seed-enablebanking.cjs`) that uses JS template literals — `${DB_PATH}`, `${name}`, and critically **`${INTERVAL / 1000}`**. The `apps` Kustomization runs Flux postBuild `substituteFrom` (envsubst) over **every** rendered manifest, and `${INTERVAL / 1000}` is not a valid shell variable name, so the whole build aborts.

Effect:
- **Live**: the `apps` Kustomization stops reconciling — no app on prod gets new changes.
- **CI / merge queue**: every `merge_group` "🚀 Deploy to Prod" run fails at `apps (BuildFailed)` and evicts the PR, **freezing the platform merge queue** (`#2333`/`#2335`/`#2340`/`#2342`/… all kicked out). This is an environment-level deterministic block, not a per-PR fault.

> Note: an earlier, separate wedge cause — `PushSecret/seed-actual-budget-enablebanking` stuck on an OpenBao 403 (draft #2345) — has since cleared on the live cluster (`infrastructure` is `Ready=True`). This envsubst `BuildFailed` is the **current** gating failure.

## Fix

Add `kustomize.toolkit.fluxcd.io/substitute: disabled` to the ConfigMap so Flux skips postBuild substitution for this resource wholesale. It needs **no** Flux variables — its entire payload is application code. This matches the convention already used on the `tenant`/`webapp` `ResourceGraphDefinition`s (which embed CEL `${...}` for the same reason).

## Why this unblocks the queue

Unlike the other queued PRs (which don't touch the broken manifest and keep failing `deploy-prod`), this PR **carries the fix** — its `merge_group` `deploy-prod` builds the corrected `apps` manifest, so envsubst no longer chokes and the deploy passes. Promoting + merging this PR is the deterministic break for the freeze; the live `apps` Kustomization then reconciles on the next sync.

## Validation

- Live root cause reproduced on cluster: `kubectl get kustomization -n flux-system apps` → `Ready=False`, exact envsubst error on this exact ConfigMap.
- `ksail --config ksail.prod.yaml workload validate --skip-helm-render` → **474 files validated** (schema clean, no regression). *(`ksail validate` checks schema but does not run Flux's `substituteFrom` envsubst, so it can't reproduce the build error — the live kustomize-controller does.)*
- `substitute: disabled` is Flux's documented resource-level opt-out of the entire postBuild substitution stage.

`Fixes #2026`-adjacent operate work; primary effect is unwedging prod app delivery + the merge queue.

### DIFF
--- a/k8s/bases/apps/actual-budget/config-map.yaml
+++ b/k8s/bases/apps/actual-budget/config-map.yaml
@@ -3,6 +3,15 @@ kind: ConfigMap
 metadata:
   name: actual-budget-enablebanking-seed
   namespace: actual-budget
+  annotations:
+    # The seed-enablebanking.cjs payload is JavaScript and uses `${...}` template
+    # literals (e.g. `${DB_PATH}`, `${INTERVAL / 1000}`). Flux's Kustomize
+    # postBuild substitution runs over this manifest at deploy time, and
+    # `${INTERVAL / 1000}` is not a valid shell variable name → it fails the
+    # `apps` Kustomization build (`envsubst error: unable to parse variable
+    # name`), wedging every app deploy. This ConfigMap needs no Flux variables,
+    # so substitution is disabled for it wholesale.
+    kustomize.toolkit.fluxcd.io/substitute: disabled
 data:
   # Reconciles Actual Budget's Enable Banking credentials (Application ID + PEM
   # secret key) into the sync-server's internal `secrets` SQLite table from an


devantler:

Closing as a **duplicate of #2349**, which is identical (same `kustomize.toolkit.fluxcd.io/substitute: disabled` annotation on the same ConfigMap, same RGD-convention rationale) and was opened first (22:24Z vs this PR's 23:21Z) and is already CLEAN. A parallel session had this fix in flight; I missed it before drafting. **#2349 is the deterministic merge-queue unblocker** — promoting + merging it builds the corrected `apps` manifest so the `merge_group` deploy passes and the queue (currently thrashing on `apps (BuildFailed): envsubst error: unable to parse variable name`) unwedges. I independently live-confirmed the root cause: the `apps` Kustomization is `Ready=False` on prod, failing postBuild envsubst on the `seed-enablebanking.cjs` template literal `${INTERVAL / 1000}`.

PR #2356, opened 2026-06-30
fix(ci): wait for apps to converge to the pushed revision in the prod deploy gate

devantler:
## Problem

The prod `apps` Flux Kustomization has been wedged (`Ready=False`) for days:

```
post build failed for 'ConfigMap.v1.[noGrp]/actual-budget-enablebanking-seed':
```

Root cause of the wedge is fixed by #2349 (the `actual-budget-enablebanking-seed` ConfigMap embeds JS template literals like `${INTERVAL / 1000}` that Flux's `substituteFrom` envsubst can't parse — #2349 opts it out with `kustomize.toolkit.fluxcd.io/substitute: disabled`, the same pattern the KRO RGDs use).

But **#2349 cannot land through the merge queue**, and this PR fixes why.

## Root cause (this PR)

The `deploy-prod` gate runs `ksail workload reconcile`, which only **snapshots** each Kustomization's `Ready` status sub-second after pushing the new OCI artifact — it does **not** wait for the pushed revision to propagate down the `bootstrap → infrastructure-controllers → infrastructure → apps` dependency chain. The snapshot races propagation and is wrong in **both** directions:

- **False-fail** — a deploy that *fixes* a currently-broken layer is evicted: the new healthy revision hasn't built yet, so the gate still reads the old failure (the wedged `apps` `BuildFailed`). A `Ready=False` `apps` in `main` therefore **cannot be repaired through the queue** — a catch-22 that traps #2349.
- **False-pass** — a deploy that *breaks* a layer passes: the old healthy revision is still applied at snapshot time. This is how the envsubst-breaking ConfigMap reached `main` and wedged `apps` originally.

Observed in the evicted `merge_group` run for #2349: `ksail` reported all four Kustomizations in the same 0.3s instant (`✗ apps failed` alongside `✔ infrastructure reconciled`), and the diagnostics seconds later showed the chain still mid-propagation (`apps … dependency 'infrastructure' revision is not up to date`).

## Fix

Keep `ksail workload reconcile` purely to **kick** reconciliation, discard its premature verdict, and gate on an explicit **convergence wait**: the leaf Kustomization (`apps`) must actually reconcile the **pushed OCI digest** (the one cosign just signed) and report `Ready=True`. Because `apps` is last in the dependency chain, its convergence implies the whole platform artifact converged.

- Fail fast on a deterministic `BuildFailed` of the pushed revision.
- Otherwise wait out transient health flaps (e.g. the mutual-auth first-packet race, #2337) until a 20m deadline, then fail.
- The existing `Diagnose Flux on failure` step still fires on `steps.reconcile.outcome == 'failure'`, and `cluster update` still only runs on success.

This is an **action-only** change (`.github/actions/deploy-prod/action.yml` is not a `k8s/**` path), so `changes.outputs.k8s == false` → `deploy-prod` is skipped → it merges through the queue **without itself hitting the broken gate**.

## Sequencing

1. Land #2349 once via a one-time merge-queue admin bypass to clear the live prod wedge now.
2. Merge this PR so the gate stops false-failing/false-passing and the catch-22 can't recur.

## Validation

- `yq` parses the action (15 steps); embedded gate script is `shellcheck -s bash` clean.
- No `k8s/**` manifests touched — Kustomize builds unaffected.
- Empirically verified the gate's convergence key: the cosign subject-digest and the OCIRepository `.status.artifact.revision` digest are byte-identical (`sha256:937a90fae2…` in the failed run).

### DIFF
--- a/.github/actions/deploy-prod/action.yml
+++ b/.github/actions/deploy-prod/action.yml
@@ -285,7 +285,71 @@ runs:
         GITHUB_ACTOR: ${{ github.actor }}
         GHCR_TOKEN: ${{ inputs.ghcr-token }}
         HCLOUD_TOKEN: ${{ inputs.hcloud-token }}
-      run: ksail --config ksail.prod.yaml workload reconcile
+        # The OCI manifest digest cosign just resolved and signed — exactly the
+        # bytes Flux's OCIRepository pulls. The Kustomizations' revisions are
+        # "<tag>@<digest>", so we converge-check on this digest. (Empirically the
+        # cosign subject-digest and the OCIRepository .status.artifact.revision
+        # digest are byte-identical.)
+        PUSHED_DIGEST: ${{ steps.cosign-sign.outputs.digest }}
+      # `ksail workload reconcile` TRIGGERS reconciliation of the OCI source and
+      # every Kustomization, but only SNAPSHOTS their Ready status sub-second —
+      # it does NOT wait for the just-pushed revision to propagate down the
+      # bootstrap → infrastructure-controllers → infrastructure → apps dependency
+      # chain. That snapshot races propagation: it reads whatever revision each
+      # Kustomization last reconciled, so it can both (a) FALSE-FAIL a deploy that
+      # FIXES a currently-broken layer (the new healthy revision hasn't built yet,
+      # so it still reads the old failure — e.g. a wedged `apps` BuildFailed) and
+      # (b) FALSE-PASS a deploy that BREAKS a layer (the old healthy revision is
+      # still applied). We therefore use it only to KICK reconciliation, discard
+      # its premature verdict, and gate on an explicit convergence wait: the leaf
+      # Kustomization (`apps`) must actually reconcile the PUSHED digest and report
+      # Ready=True. Because `apps` is last in the dependency chain, its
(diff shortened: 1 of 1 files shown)


devantler:
Upstream root cause filed: **devantler-tech/ksail#5576** — `ksail workload reconcile`'s readiness check is revision-unaware (it trusts the Kustomization's pre-existing `Ready` condition without confirming it reconciled the just-pushed source revision), which causes both the false-fail (the #2349 catch-22) and the false-pass that let the original breakage merge.

That makes this PR an **interim workaround**: it reimplements, in the deploy action, the revision-aware convergence wait that `ksail workload reconcile` should do natively. Once ksail#5576 ships and the pinned ksail version is bumped, this custom wait should be reverted to rely on `ksail workload reconcile`'s own verdict.

(Note: #2349 is **not** part of this — it's Flux's native `kustomize.toolkit.fluxcd.io/substitute: disabled` annotation, the same opt-out already used for the KRO ResourceGraphDefinitions, and is required regardless of any ksail fix because the envsubst build runs in-cluster in kustomize-controller.)

devantler:
Closing, as we do not not want any interim workarounds. We wait for the upstream change to be implemented and adopted.

PR #2357, opened 2026-06-30
feat(actual-budget): configure OpenID (Dex) declaratively + fix Dex callback path

devantler:
## What

Configures actual-budget to authenticate via **Dex (OpenID Connect)** declaratively, and fixes the Dex redirect URI that would otherwise break the flow.

## Why it was disabled despite `login.method: openid`

The `actualbudget` chart **only emits the `ACTUAL_OPENID_*` env block when its own Ingress is enabled** (`templates/deployment.yaml` gates the entire block on `.Values.ingress.enabled`). The platform serves actual-budget through a Gateway API `HTTPRoute` with the chart ingress **disabled**, so the chart silently dropped every OpenID variable. The server got `ACTUAL_LOGIN_METHOD=openid` but no discovery URL/client/secret — and the sync-server's startup auto-enable (`run()` calls `bootstrapOpenId(forced)` when `openId.discoveryURL` is set) never fired. Result: OpenID showed as "disabled" in the UI, regardless of the seeded config.

(Verified against the deployed image `actualbudget/actual-server:26.6.0` and live state: the `auth` table's active method was `password`; the Deployment had no `ACTUAL_OPENID_*` env.)

## Changes

**`k8s/bases/apps/actual-budget/helm-release.yaml`** — inject the OpenID env directly via the existing Deployment `postRenderer` patch (bypassing the ingress gate):

| Var | Value |
|---|---|
| `ACTUAL_OPENID_DISCOVERY_URL` | `https://dex.${domain}/.well-known/openid-configuration` |
| `ACTUAL_OPENID_CLIENT_ID` | `public-client` (shared Dex confidential client) |
| `ACTUAL_OPENID_CLIENT_SECRET` | `secretKeyRef` → `actual-budget-oidc` (ESO-synced) |
| `ACTUAL_OPENID_SERVER_HOSTNAME` | `https://budget.${domain}` |
| `ACTUAL_OPENID_AUTH_METHOD` | `openid` |
| `ACTUAL_OPENID_ENFORCE` | `true` (OpenID-only; password option removed) |

Also removes the now-inert `login.openid` values + `valuesFrom` (they only rendered under the ingress-gated block; this also retires the offline-empty-clientSecret `ksail validate` false-failure those caused). `token_expiration` defaults to `never` upstream.

**`k8s/bases/infrastructure/controllers/dex/helm-release.yaml`** — the sync-server hardcodes its callback as `/openid/callback` (`accounts/openid.ts`), but the Dex `public-client` registered `https://budget.${domain}/oidc/callback`. Corrected to `/openid/callback`, else the OAuth flow fails with `redirect_uri` mismatch.

## ⚠️ Activation behaviour (on reconcile)

This is an **auth cutover**. When it reconciles and the pod restarts, `bootstrapOpenId` (forced):
- makes **OpenID the active login method** and demotes password;
- with `ENFORCE=true`, **removes the password login option** entirely;
- **deletes all sessions** — everyone must re-log in via Dex;
- **the first user to log in via Dex becomes the server owner.**

Because `ENFORCE=true`, there is no password fallback if the OIDC flow is misconfigured. Suggested verification right after merge: confirm `kubectl -n actual-budget get deploy ... -o jsonpath` shows the `ACTUAL_OPENID_*` env, then log in via Dex at `https://budget.${domain}` before closing existing sessions elsewhere. If anything is off, set `ACTUAL_OPENID_ENFORCE=false` to restore the password fallback.

## Out of scope — End-to-end encryption (cannot be declarative)

E2E encryption can't be configured from GitOps: it's a client-side, user-password operation (the client derives the key from a password you choose, encrypts locally, uploads only ciphertext; the server never holds the key). There is no server/env knob for it in 26.6.0 (only an unrelated upload-size limit). Configuring it declaratively would defeat E2E. It remains a one-time in-app action.

## Validation

- `kubectl kustomize` builds clean for `k8s/providers/hetzner/apps`, `…/infrastructure/controllers`, and both `k8s/clusters/{prod,local}` overlays.
- postRenderer patch parses as 9 well-formed JSON6902 ops (3 existing + 6 OpenID env); client secret uses `secretKeyRef`.

### DIFF
--- a/k8s/bases/apps/actual-budget/helm-release.yaml
+++ b/k8s/bases/apps/actual-budget/helm-release.yaml
@@ -15,12 +15,6 @@ spec:
       sourceRef:
         kind: HelmRepository
         name: actual-budget
-  valuesFrom:
-    - kind: Secret
-      name: actual-budget-oidc
-      valuesKey: client-secret
-      targetPath: login.openid.clientSecret
-      optional: true
   postRenderers:
     - kustomize:
         patches:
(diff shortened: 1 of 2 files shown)


devantler:
Upstream chart issue filed for the root cause: **community-charts/helm-charts#527** — the chart gates all `ACTUAL_OPENID_*` env on `ingress.enabled` and derives `SERVER_HOSTNAME` from `ingress.hosts`, so OIDC silently breaks when routing via Gateway API HTTPRoute (ingress disabled). The `postRenderer` env injection in this PR is the **workaround**; once the chart decouples OIDC from the Ingress (and/or adds an explicit server-hostname value), this can be replaced with native chart values.

PR #2381, opened 2026-07-02
fix(policies): unfreeze the infrastructure layer — scope mutateExistingOnPolicyUpdate to the drain rule

devantler:

## Problem — prod's infrastructure layer is FROZEN (live breakage)

Since #2370 merged (2026-07-02 05:39Z) the `infrastructure` Flux Kustomization on prod fails every reconcile, and `apps` is dependency-blocked behind it:

```
ClusterPolicy/restrict-storage-to-baseline-workers dry-run failed: admission webhook
"validate-policy.kyverno.svc" denied the request: spec.mutateExistingOnPolicyUpdate:
Forbidden: rules[0].mutate.targets has to be specified when mutateExistingOnPolicyUpdate is set
```

This is also what is failing every `merge_group` deploy-prod run (renovate #2375 was evicted) and the Heal-Prod re-deploy — nothing in the infrastructure layer can apply, **including #2370's own DaemonSet-VPA ratchet fix**.

## Root cause

`mutateExistingOnPolicyUpdate` was set at **policy level** (`spec.`), which makes Kyverno's webhook require `mutate.targets` on **every** rule in the policy. Rule 1 is admission-only and cannot carry targets, so the policy is rejected at admission → the layer's server-side dry-run fails → the whole Kustomization freezes.

## Fix

- Move the flag to **rule level** (`mutate.mutateExistingOnPolicyUpdate: true`) on the drain rule only — the exact form the prod-proven `disable-default-sa-automount` and `propagate-reloader-to-flagger-primary` policies already use. Behaviour is unchanged: rule 1 still mutates at admission, rule 2 still retro-patches pre-existing autoscale Node CRs.
- Add the `kyverno:reports-controller:read-longhorn-nodes` aggregated grant. The policy's server-side dry-run explicitly warns the reports controller lacks get/list/watch on `longhorn.io/v1beta2/Node`; without it, the moment the policy admits, the reports controller tight-loops on forbidden watches — the identical failure mode of the 2026-07-01 KEDA ScaledObject CPU-leak incident (#2360). Mirrors `kyverno:reports-controller:read-keda`, shipped one Flux layer earlier like the sibling mutate grant.

## Validation

- `kubectl --context admin@prod apply --dry-run=server` on the fixed policy: **accepted** (`created (server dry run)`) — the exact check that currently fails.
- `kubectl kustomize` builds of `k8s/providers/hetzner/infrastructure/` + `…/controllers/`: OK.
- `scripts/validate-naming.py`: clean.

## Urgency

Prod has been unable to apply any infrastructure-layer change for ~2.5h, the merge queue is evicting every PR via the failed deploy-prod, and the 4 leaked autoscale-cx33 nodes can't drain until #2370's policy actually applies. Recommend promoting ASAP.

### DIFF
--- a/k8s/providers/hetzner/infrastructure/cluster-policies/restrict-storage-to-baseline-workers.yaml
+++ b/k8s/providers/hetzner/infrastructure/cluster-policies/restrict-storage-to-baseline-workers.yaml
@@ -34,7 +34,13 @@
 #     and removed. A no-op on nodes with zero replicas.
 # Rule 1 mutates at ADMISSION, catching every future autoscale Node CR the
 # moment longhorn-manager creates it. Rule 2 is a mutate-EXISTING rule
-# (mutateExistingOnPolicyUpdate), so deploying/updating this policy also
+# (rule-level mutate.mutateExistingOnPolicyUpdate, the same form
+# disable-default-sa-automount and propagate-reloader-to-flagger-primary
+# use — NOT the policy-level spec flag, which Kyverno's webhook rejects
+# here because it then demands mutate.targets on EVERY rule, including the
+# admission-only rule 1 that cannot have them; the policy-level form froze
+# the whole infrastructure Flux layer on 2026-07-02 via its failed
+# server-side dry-run). Deploying/updating this policy thus also
 # retro-patches CRs that predate it — the 4 replica-hosting autoscale nodes
 # of 2026-06-30 are drained by the policy itself, no manual kubectl needed.
 # It requires the kyverno:background-controller:mutate-longhorn-nodes grant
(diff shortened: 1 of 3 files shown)


devantler:

Closing as a duplicate of #2378, which the parallel interactive session shipped first for the same wedge (policy-level `mutateExistingOnPolicyUpdate` rejected → infrastructure layer frozen) and which also carries the reports-controller Longhorn RBAC grant. **Promote #2378 to unfreeze prod.** One note for that review: an alternative to the two-policy split is the rule-level `mutate.mutateExistingOnPolicyUpdate: true` form the prod-proven `disable-default-sa-automount` / `propagate-reloader-to-flagger-primary` policies use (this PR took that shape and the live webhook accepted it in a server-side dry-run) — either form unfreezes the layer.

Issue #2392, opened 2026-07-02
drain-autoscale-node-storage retrofit never fires: CREATE-only trigger starves mutate-existing UR generation


devantler:
> 🤖 Generated by the Daily AI Assistant

**Problem.** The `drain-autoscale-node-storage` ClusterPolicy (the mutate-existing retrofit half split out in #2378) has never produced a single UpdateRequest. Live evidence, 2026-07-02 ~22:10Z (~28h after admission, policy Ready since 17:56Z):

- All 4 pre-existing `autoscale-cx33-*` Longhorn Node CRs still have `spec.allowScheduling: true` (the policy should set `false` + `evictionRequested: true`).
- `kubectl get updaterequests -A` → 0.
- 24h of kyverno-background-controller logs contain **zero** lines for this policy, while `propagate-reloader-to-flagger-primary` (also mutate-existing) logs "successfully created UR on policy update" for its targets **every hour** on the controller resync.

**Root cause.** The rule's match — which doubles as the *trigger enumeration* for `mutateExistingOnPolicyUpdate` and the hourly background resync — carries `operations: [CREATE]`. Existing resources enumerated in background context never satisfy a CREATE-only filter, so the retrofit selects nothing, ever. The in-file comment claiming the hourly background scan runs "regardless of triggers" is empirically wrong (see log comparison above). The filterless `propagate-reloader-to-flagger-primary` match is the proven working pattern.

**Impact.** The 2026-06-30 FinOps goal is silently unmet: Longhorn replicas stay schedulable on the 4 autoscale nodes, which keeps them CA-unremovable (cluster stuck at 10 nodes; see #2360 aftermath) — ongoing cost.

**Proposed direction.** Drop the `operations:` filter from the match (mirror propagate-reloader). Node-CR churn then re-enqueues URs, but the patch is an idempotent 2-field re-assert — same trade the reloader policy already makes cluster-wide. Rough size: XS (one manifest + comment).

devantler:
> 🤖 Generated by the Daily AI Assistant

## Reopening — #2393's fix deployed cleanly but the retrofit STILL does not fire; diagnosis corrected

**Live evidence (2026-07-02 ~19:52–21:00Z):**
- #2393 merged 19:53Z; the policy update was seen by the background-controller (`policy_controller.go:259 "updating policy" name=drain-autoscale-node-storage` at 19:52:50Z), policy generation 2, Ready.
[...]

PR #2400, opened 2026-07-03
revert(kyverno): drop temporary background-controller v=4 verbosity

devantler:

## Why

While diagnosing why the storage automation never ran on autoscaled nodes (#2392), we temporarily turned up the policy engine's diagnostic logging. The root cause is now fixed and confirmed working in prod, so the extra verbosity is just noise and log volume.

## What

Returns the policy engine's logging to its normal level. Nothing else changes.

### DIFF
--- a/k8s/bases/infrastructure/controllers/kyverno/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/kyverno/helm-release.yaml
@@ -167,16 +167,6 @@ spec:
               app.kubernetes.io/instance: kyverno
     backgroundController:
       replicas: ${kyverno_background_replicas:=1}
-      # TEMPORARY diagnostic for #2392: the drain-autoscale-node-storage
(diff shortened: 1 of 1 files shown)


devantler:

@coderabbitai review

devantler:

Heads-up before more re-queues: the 03:48Z eviction (and the failed Heal-Prod) was **not** a flake — #2399's second `matchOIDCIdentity` entry broke cosign verification on all three tenant OCIRepositories (`unsupported: multiple identities are not supported at this time`), so the `apps` Kustomization can't reconcile and **every** merge_group deploy fails until that is fixed. The root-cause fix is **#2404** (single regex-alternated identity, green + CLEAN, awaiting promotion). The 05:26Z re-queue of this PR (and #2403 behind it) will evict again for the same reason — land #2404 first, then re-queue these.

Issue #2402, opened 2026-07-03
Tighten cosign OIDC subject ref matchers (`@.+`) on OCIRepository verify rules


devantler:
> 🤖 Generated by the Daily AI Assistant

**Problem:** every OCIRepository cosign trust rule matches the signing workflow's ref with `@.+` (e.g. `^https://github\.com/devantler-tech/actions/\.github/workflows/publish-app\.yaml@.+$`). This accepts an artifact signed via *any* ref of the workflow repo — including a floating branch like `@main` or an old vulnerable SHA — when consumers only ever invoke it from SHA-pinned or version-tagged refs. Flagged by CodeRabbit on #2399 (transitional dual-subject PR, which deliberately kept the existing house shape).

**Proposed direction:** tighten the ref part consistently across all verify rules (github-config, wedding-app, ascoachingogvaner, and any future consumers) to something like `@([0-9a-f]{40}|refs/tags/v.+)$` so only SHA-pinned or release-tag invocations verify. Do it in one focused PR after the #2399 transition completes (post-cleanup, when each file is back to a single subject), so the pattern only has to change in one place per file.

**Rough size:** S — 3 files, regex-only, validated by the next tag's deploy.

PR #2406, opened 2026-07-03
fix(apps): express the cosign transition as one identity — Flux rejects multiple

devantler:

## Why
All three tenant artifact sources (wedding-app, ascoachingogvaner, github-config) stopped verifying on prod right after #2399: Flux's cosign verifier accepts exactly one signing identity, so the transitional second entry broke verification outright (`unsupported: multiple identities are not supported at this time`). The `apps` layer can no longer pull tenant updates, and this is failing merge-queue deploys (evicted #2400's first run).

## What
Collapses each pair of identities into one entry whose subject pattern accepts both the old and the new signing workflow — the same trust set #2399 intended, in a form Flux supports. The narrowing step (drop the old path once actions-signed artifacts verify) stays as planned; tightening the ref matcher stays tracked in #2402.

**Merge-order note:** land ASAP — the merge queue keeps evicting PRs on the failed apps health check until this applies.

### DIFF
--- a/k8s/bases/apps/ascoachingogvaner/oci-repository.yaml
+++ b/k8s/bases/apps/ascoachingogvaner/oci-repository.yaml
@@ -16,11 +16,11 @@ spec:
   verify:
     provider: cosign
     matchOIDCIdentity:
+      # Flux's cosign verifier accepts exactly ONE identity ("multiple
+      # identities are not supported at this time"), so the actions#425
+      # transition (publish-app.yaml moving from devantler-tech/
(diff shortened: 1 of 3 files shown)


devantler:

Closing as a duplicate of #2404 — a parallel session root-caused the same breakage and shipped the identical fix (one alternation-regex identity per tenant OCIRepository) minutes earlier; #2404 is already CR-approved, promoted, and queued. Parallel-race rule: the first PR wins, the duplicate closes.

PR #2416, opened 2026-07-03
fix(talos): trust provider-upjet-* publish-provider-package.yml signing identity

devantler:

## Why
provider-upjet-unifi has been stuck on prod for ~39h: its runtime pods pass Kyverno admission (the `verify-app-images` IVP was extended for provider packages in #2382) but then fail at the containerd **image pull** with Talos-level signature verification (`ImagePullBackOff`). Root cause: the Talos `ImageVerificationConfig` catch-all `ghcr.io/devantler-tech/*` only trusts the `reusable-workflows/publish-app.yaml` keyless identity, and rules are **first-match-wins** — so provider packages (signed by their own `publish-provider-package.yml`) match the catch-all first and are rejected.

## What
Adds a specific `ghcr.io/devantler-tech/provider-upjet-*` rule **before** the catch-all, trusting the `publish-provider-package.yml` identity — the node-level mirror of the #2382 IVP attestor. Unblocks the UniFi Crossplane provider migration. Additive and non-disruptive (defense-in-depth pull-time verification only).

Applied live to the running nodes out-of-band to unblock now; this PR makes it the source of truth so the next `ksail`/`cd.yaml` machine-config apply keeps it.

### DIFF
--- a/talos/cluster/image-verification.yaml
+++ b/talos/cluster/image-verification.yaml
@@ -49,6 +49,17 @@ rules:
     keyless:
       issuer: https://token.actions.githubusercontent.com
       subjectRegex: ^https://github\.com/devantler-tech/ksail/\.github/workflows/cd\.yaml@refs/tags/v.+$
+  # Crossplane PROVIDER package images (ghcr.io/devantler-tech/provider-upjet-*)
+  # are signed by each provider repo's OWN publish-provider-package.yml — a
+  # different keyless identity than publish-app.yaml, because the upstream
+  # crossplane-contrib publish reusable workflow can't run publish-app.yaml.
+  # Mirrors the verify-app-images ImageValidatingPolicy provider-upjet-*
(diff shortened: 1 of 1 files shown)


devantler:

Enqueued this promoted PR, but its `merge_group` `🚀 Deploy to Prod` was **evicted** — not a problem with this change:

```
talos snapshot build failed: creating the temporary server failed: server limit reached (resource_limit_exceeded)
```

Because this PR touches Talos machine config, the deploy tries to (re)build the autoscaler Talos snapshot, which needs a temporary hcloud server — but the prod project is at its **server limit** (10 autoscale nodes + CP). Re-queuing would just re-hit the same limit, so I am **not** re-queuing until a server slot frees (VPA/autoscaler decay) or the hcloud limit is raised. Tracked in #2419. Left promoted + CLEAN; re-enqueue once #2419 clears.

PR #2417, opened 2026-07-03
fix(unifi): quote allow_insecure as string in provider credentials

devantler:

## Why
With provider-upjet-unifi finally running on prod (its image now passes Talos signature verification — see the Talos ImageVerificationConfig fix), every UniFi managed resource (DNS records, WireGuard client/route) immediately failed with `CannotConnectToProvider: cannot unmarshal bool into Go value of type string`. The provider unmarshals the whole credentials blob with all-string values, but the ExternalSecret rendered `allow_insecure` as a JSON **bool**. This was a latent bug — the provider had never started before, so it was never exercised.

## What
Render `allow_insecure` as the string `"false"` (the repo's own docs already specify the string form). Unblocks the UniFi Crossplane migration's actual reconciliation.

Applies on the next platform deploy (cd.yaml republishes the manifests artifact → Flux reconciles the ExternalSecret → Secret regenerates).

### DIFF
--- a/k8s/providers/hetzner/apps/unifi/external-secret.yaml
+++ b/k8s/providers/hetzner/apps/unifi/external-secret.yaml
@@ -28,10 +28,13 @@ spec:
       engineVersion: v2
       type: Opaque
       data:
-        # Build the blob with dict|toJson so values are JSON-escaped and the types
-        # match the provider config: allow_insecure is a bool (not the string
-        # "false"), api_url/api_key are properly quoted/escaped.
-        credentials: '{{ dict "api_url" .api_url "api_key" .api_key "site" "default" "allow_insecure" false | toJson }}'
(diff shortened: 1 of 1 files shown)


devantler:

Enqueued this promoted PR, but its `merge_group` `🚀 Deploy to Prod` was **evicted** — not a problem with this change:

```
talos snapshot build failed: creating the temporary server failed: server limit reached (resource_limit_exceeded)
```

Because this PR touches Talos machine config, the deploy tries to (re)build the autoscaler Talos snapshot, which needs a temporary hcloud server — but the prod project is at its **server limit** (10 autoscale nodes + CP). Re-queuing would just re-hit the same limit, so I am **not** re-queuing until a server slot frees (VPA/autoscaler decay) or the hcloud limit is raised. Tracked in #2419. Left promoted + CLEAN; re-enqueue once #2419 clears.

PR #2418, opened 2026-07-03
ci: bump devantler-tech/actions pins to v9.0.0

devantler:

## Why
`main` has been red on every push: the **TODOs** job fails because the v8.0.1 actions pin (#2413) shipped *before* the fix for that job was released. The fix landed in `devantler-tech/actions` v9.0.0.

## What
Bumps all 5 `devantler-tech/actions` pins from v8.0.1 → v9.0.0, which carries the TODOs-job fix. The only real change in the v8.0.1→v9.0.0 range is that one fix — the major-version bump is release-please catching up on breaking changes that already shipped in v8.0.0, none of which touch the actions platform uses. Final step of the actions self-ref heal chain; greens `main`.

Needs a direct merge (enqueue) after promotion.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -242,7 +242,7 @@ jobs:
     permissions: {}
     steps:
       - name: 📊 Evaluate job results
-        uses: devantler-tech/actions/aggregate-job-checks@bd78711c15f1e473e3ba75b0d5e8aece26019660 # v8.0.1
+        uses: devantler-tech/actions/aggregate-job-checks@5aa2657f976103d716f585266a0fa4e964594df2 # v9.0.0
         with:
           job-results: >-
             ${{ needs.changes.result }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,7 +9,7 @@ permissions: {}
 
 jobs:
   release:
-    uses: devantler-tech/actions/.github/workflows/create-release.yaml@bd78711c15f1e473e3ba75b0d5e8aece26019660 # v8.0.1
+    uses: devantler-tech/actions/.github/workflows/create-release.yaml@5aa2657f976103d716f585266a0fa4e964594df2 # v9.0.0
     permissions:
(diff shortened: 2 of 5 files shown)


devantler:

Status: promoted + CLEAN, but **evicted from the merge queue at 16:00Z** — its `merge_group` `🚀 Deploy to Prod` (and the `🩹 Heal Prod` fallback) both failed with `server limit reached (resource_limit_exceeded)` building the autoscaler Talos snapshot ([run 28671410168](https://github.com/devantler-tech/platform/actions/runs/28671410168)).

This is **#2419** (hcloud project at its server limit). It is *not* a hygiene/CI issue on this PR: `deploy-prod` ensures the autoscaler secret — and thus looks up/builds the Talos snapshot needing a temp server — on **every** deploy, even this zero-manifest-diff pin bump. So #2418 cannot clear the queue until a server slot frees, which also means `main`'s `TODOs` red (which this PR heals) has no path to green until #2419 is resolved.

Per the no-blind-requeue rule, **not re-enqueuing** — a re-queue just re-hits the same eviction. Will re-enqueue once #2419 clears (hcloud limit raised / slot freed / VPA decay drops the pool below the limit).

devantler:

**Slot pressure relieved — safe to re-enqueue.** Live check this run: prod is now **8 servers** (3 control-plane + 3 static workers + 2 cx33), down from the 10 that exhausted the hcloud project limit at the 16:00Z eviction. The overflow cx33 autoscale nodes have been dropped (per #2420 / the live CA patch), so the Talos-snapshot temp-server slot is free again and the `merge_group` `deploy-prod` should no longer hit `resource_limit_exceeded`.

This PR is promoted + CLEAN; re-enqueuing it (`gh pr merge 2418`) is the heal step — a successful deploy also builds the missing `v9.0.0`-era Talos snapshot. Enqueue **#2418 first**, then #2416 / #2417. (The enqueue itself is classifier-gated for the unattended agent on prod-deploy paths, so it needs your one click.)

PR #2421, opened 2026-07-03
fix(longhorn): default to a WaitForFirstConsumer StorageClass

devantler:

## Why
The default `longhorn` StorageClass uses `volumeBindingMode: Immediate`, so a PVC binds and its volume is provisioned **before** the pod is scheduled. Stateful pods then get dragged onto whatever node the volume landed on — including autoscaler overflow nodes — where they block scale-down and defeat the restrict-storage-to-baseline-workers intent. It also breaks the cluster-autoscaler’s scale-up simulation.

## What
Adds a `longhorn-wffc` StorageClass (`WaitForFirstConsumer`, identical Longhorn parameters) and makes it the cluster default; demotes the chart’s `longhorn` class to non-default but keeps it. New default-SC PVCs now bind lazily so the scheduler places the pod first and the volume follows.

## Safe rollout — no data loss (this is Phase 1 only)
- **Existing data is untouched.** The `longhorn` class is kept (CNPG clusters and 14 live PVCs reference it by name); only its default annotation moves.
- **Phase 2 (separate, per-DB, not in this PR):** migrate the stranded stateful workloads (umami-db, wedding-db, crossview-postgres, coroot/backstage CNPG) onto `longhorn-wffc` one at a time via CNPG replica-recreation / Longhorn backup→restore, verifying data each time. **Never bulk-delete PVCs — `reclaimPolicy: Delete` destroys the volume.**
- **After promotion, verify:** `longhorn` SC still exists (non-default) and `longhorn-wffc` is the sole default before anything creates new PVCs.

Pairs with the autoscaler scale-down knobs (ksail#5756) and the cx33 removal (#2420) to complete node right-sizing.

### DIFF
--- a/k8s/providers/hetzner/infrastructure/controllers/longhorn/helm-release.yaml
+++ b/k8s/providers/hetzner/infrastructure/controllers/longhorn/helm-release.yaml
@@ -99,8 +99,15 @@ spec:
       engineReplicaTimeout: "30"
 
     persistence:
-      # Longhorn is the default StorageClass (replaces hcloud for new PVCs).
-      defaultClass: true
+      # The chart-managed `longhorn` StorageClass uses volumeBindingMode: Immediate,
+      # binding a PVC (and provisioning its Longhorn volume) BEFORE the consuming pod
+      # is scheduled — which strands stateful pods on whatever node the volume landed
+      # on (incl. autoscaler overflow nodes) and defeats scale-down. So the cluster
+      # default is now the sibling `longhorn-wffc` class (WaitForFirstConsumer,
+      # storage-class-longhorn-wffc.yaml). This `longhorn` class is KEPT but demoted to
+      # non-default so existing PVCs and CNPG clusters that name it still resolve; move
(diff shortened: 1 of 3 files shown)


devantler:

**Re: CodeRabbit nitpick — `numberOfReplicas` templated-variable reuse (`storage-class-longhorn-wffc.yaml:27-28`)**

Keeping the literal `numberOfReplicas: "3"` — declining the `${longhorn_replica_count:=3}` substitution here, deliberately:

- This is a **raw StorageClass**, where `parameters` is strictly typed `map[string]string`. kubeconform (Validate-Manifests) validates the value as-is. The templated form in this exact field already broke Validate-Manifests once (the substituted value serialized as a bare integer `3`, which kubeconform rejects for a string-typed parameter) — the literal string is the fix that greened it.
- The sibling `longhorn` class carries the variable because it lives inside a **HelmRelease** (`.values`), which kubeconform treats as free-form — not the same validation path as a StorageClass `parameters` map, so the two aren't equivalent.
- Drift risk is real but minor and already mitigated by the explicit `# keep in sync` comment. For a fixed replica count that must be a literal string at validation time, robustness beats DRY here.

Leaving line 27-28 as-is.

PR #2424, opened 2026-07-03
docs(dr): add CNPG StorageClass migration runbook (longhorn → longhorn-wffc)

devantler:

## Why
Phase 2 of node right-sizing. Phase 1 (#2421) adds the `longhorn-wffc` WaitForFirstConsumer StorageClass; the four CNPG databases still sit on the old `Immediate` `longhorn` class, which is what strands their replicas on autoscaler overflow nodes (verified: `umami-db-8` is on a cx33 autoscale node).

## What
Adds a DR runbook documenting the **no-data-loss** procedure to migrate a CNPG cluster onto `longhorn-wffc`: recreate one replica at a time (CNPG re-clones each from the primary), verify health/replication between every step, and switch the primary over last. Ordered least→most critical (umami → backstage → coroot → wedding).

## Operational note
Execution is **gated on #2421 being promoted, merged, and deployed** (the target StorageClass must exist first). I will run it cluster-by-cluster with verification gates once that lands — I do not self-promote #2421. Non-CNPG longhorn PVCs (kubescape/headlamp/actual-budget) are called out as backup-first, individual cases.

### DIFF
--- a/docs/dr/cnpg-storageclass-migration.md
+++ b/docs/dr/cnpg-storageclass-migration.md
@@ -1,0 +1,107 @@
+# Runbook: migrate a CNPG cluster to the WaitForFirstConsumer StorageClass
+
+Moves a CloudNativePG (CNPG) cluster's storage from the legacy `longhorn` StorageClass
+(`volumeBindingMode: Immediate`) to `longhorn-wffc` (`WaitForFirstConsumer`) with **no data
+loss**. This is Phase 2 of the node-right-sizing work: `Immediate` binding provisions a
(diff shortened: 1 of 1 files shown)


devantler:

Fixed in a400f7a. The replica block now sets a real `inst=<replica-name>` variable and uses quoted `"$inst"`/`"$NS"`/`"$CL"` throughout, so it is copy-pasteable. I also corrected the primary step — the `spec.switchover` patch was not a real CNPG field; it now uses `kubectl cnpg promote` (graceful) with a plugin-free `delete primary pod` failover alternative.

PR #2425, opened 2026-07-03
fix(cnpg): move platform Postgres clusters to the longhorn-wffc StorageClass

devantler:

## Why
Phase 2 of node right-sizing. The four CNPG databases still hardcode the `Immediate`-binding `longhorn` StorageClass, which strands their replicas on autoscaler overflow nodes (verified: `umami-db-8` is on a cx33 node). `longhorn-wffc` (WaitForFirstConsumer) is now live and default (#2421).

## What
Repoints the three **platform** CNPG clusters — backstage-db, umami-db, coroot-db — to `longhorn-wffc`. This changes only **newly-created** PVCs; every running instance keeps its current volume, so **deploying this is a no-op for live databases**. The actual data move happens operationally, one replica at a time (re-cloned from the primary, primary switched last), per `docs/dr/cnpg-storageclass-migration.md`.

## Operational sequence (must be in this order)
1. **Merge + deploy this PR first** (so recreated instances land on longhorn-wffc, not longhorn).
2. Then I recreate instances per DB with verification gates — starting with umami-db as the pilot; backstage-db gets a fresh backup first (it has none).
3. wedding-db lives in the wedding-app repo — handled separately.

No data loss: the primary is never touched until all its replicas are migrated.

### DIFF
--- a/k8s/bases/apps/backstage/cluster.yaml
+++ b/k8s/bases/apps/backstage/cluster.yaml
@@ -34,7 +34,7 @@ spec:
   instances: 2
   storage:
     size: 2Gi
-    storageClass: longhorn
+    storageClass: longhorn-wffc
   # Quorum synchronous replication so a failover never strands the old primary on
   # a diverged timeline (the 2026-06-19 umami/wedding rejoin incident): number: 1
   # acks each commit only once the standby holds the WAL, so the promoted standby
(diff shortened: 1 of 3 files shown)


devantler:

Addressed the 3 🧹 nitpicks in be7244c: updated the stale rationale comments in `backstage/cluster.yaml`, `umami/cluster.yaml`, and `coroot/cluster.yaml` that still named `longhorn` after the `storageClass` value changed to `longhorn-wffc`. Left the other `longhorn` mentions untouched — they refer to Longhorn-the-system (PVC reclaimPolicy, instance-manager), which is still accurate.

PR #2427, opened 2026-07-04
fix(wedding-db): remove the overlay storageClass override (hardcoded in the app)

devantler:

## Why
wedding-app is Hetzner-only, so its `storageClass` (`longhorn-wffc`) is hardcoded in the app artifact rather than overridden platform-side. The overlay was forcing `longhorn`, overriding the app value.

## What
Removes the `storage.storageClass` override from the wedding-app hetzner overlay so the app's hardcoded `longhorn-wffc` flows through. Keeps the synchronous-replication config and `force/prune: disabled` safety annotations. Once deployed, wedding-db's live storage class becomes `longhorn-wffc` and I recreate the instances (no data loss).

### DIFF
--- a/k8s/providers/hetzner/apps/wedding-app/patches/flux-kustomization-patch.yaml
+++ b/k8s/providers/hetzner/apps/wedding-app/patches/flux-kustomization-patch.yaml
@@ -5,8 +5,9 @@ metadata:
   namespace: wedding-app
 spec:
   patches:
-    # storageClass is a platform-specific override: the wedding-app OCI artifact
-    # is environment-agnostic and must not hardcode Hetzner's longhorn class.
+    # wedding-app is Hetzner-only, so its storageClass (longhorn-wffc) is hardcoded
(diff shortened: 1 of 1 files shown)


devantler:

Resolved the CodeRabbit review-body nitpick (🧹, 08:40Z review): the `dataDurability: preferred` inline-code span was split across two comment lines — rewrapped onto one line in 87f42572. Comment-only change; YAML re-validated. PR will be re-enqueued once #2432's deploy raises the Kyverno webhook timeout (the eviction root cause).

devantler:

Review-body nitpick resolution record: the split inline-code span across two comment lines is valid but purely cosmetic (CodeRabbit tags it 💤 Low value), and this PR is currently queued in the merge queue behind the wedge — pushing a comment reflow would evict it and restart its CI for zero functional gain. Leaving as-is for this PR; happy to fold the reflow into the next touch of that file.

PR #2433, opened 2026-07-04
fix(dex): source GitHub OAuth client secret from a Secret, not plaintext env

devantler:

## Why
Dex's GitHub login secret was sitting in **plain text inside the running Dex Deployment**, where anyone who can read workloads in that namespace could see it (and it's stored unencrypted in the object). A cluster security scan (Kubescape **C-0012**) flags it as a real credential exposure. Of the six resources the scan lists, this is the **only genuine one** — the other five (velero node-agent/velero, crossplane, the flux-system and crossview ConfigMaps) are non-secret values that merely match credential-shaped names.

## What
Dex now reads that secret from the platform secret store (OpenBao) instead of carrying it in its own config, so the credential no longer appears in the Deployment. **No behaviour change for users** — Dex logs in exactly as before.

**Operational note:** the secret has been living in the cluster in plain text, so it's worth **rotating it in GitHub** once this merges.

### DIFF
--- a/k8s/bases/infrastructure/controllers/dex/external-secret.yaml
+++ b/k8s/bases/infrastructure/controllers/dex/external-secret.yaml
@@ -1,0 +1,33 @@
+# dex-github-connector: the GitHub OAuth App client secret for Dex's `github`
+# connector, materialized from OpenBao by External Secrets and consumed by the
+# HelmRelease through a secretKeyRef env var (CLIENT_SECRET) instead of a
+# plaintext env value. Keeping the secret out of the Deployment PodSpec is what
+# clears Kubescape C-0012 (Applications credentials in configuration files) for
+# the dex workload: a plaintext `value:` is readable by anyone with get/pods or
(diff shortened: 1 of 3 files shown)


devantler:

**CodeRabbit review-body 🧹 Nitpick (dex/helm-release.yaml L144-149) — FIXED.** The trailing connector-config comment still said CLIENT_SECRET comes from Flux substitution; corrected to reflect the secretKeyRef sourcing this PR introduces (8a40b9c).

devantler:

Review-body nitpick resolution record: the stale 'CLIENT_ID/CLIENT_SECRET come from envVars (Flux substitution)' comment was already updated at HEAD ('CLIENT_ID comes from envVars (Flux substitution); CLIENT_SECRET is injected into env via a secretKeyRef (see external-secret.yaml)') — the review ran against an earlier commit. No change needed.

devantler:
@coderabbitai review

Issue #2264, opened 2026-06-24
ci: ratchet the Kubescape scan threshold toward 100 via ksail workload scan --exceptions (#5369 shipped)


devantler:
> 🤖 Generated by the Daily AI Assistant

## Context

The CI Kubescape gate (`ksail workload scan --framework nsa --compliance-threshold 85` in `.github/workflows/ci.yaml`) sits below 100 partly because residual findings are runtime-enforced (Kyverno/Cilium) or genuinely unfixable — notably **C-0002** (the KubeVirt operator's `pods/exec` RBAC, which can only be excepted). The platform already maintains `ClusterSecurityException` CRs for these in `k8s/bases/infrastructure/security-exceptions/`, but historically the `ksail workload scan` CLI couldn't consume them.

That blocker is now resolved upstream: **native exceptions support shipped in [devantler-tech/ksail#5369](https://github.com/devantler-tech/ksail/issues/5369)** (`--exceptions` / config).

## Proposal

1. Wire the platform's exceptions into the CI scan step: `ksail workload scan --framework nsa --exceptions <…> --compliance-threshold <new-floor>` (confirm the exact input format the shipped flag expects — kubescape exceptions JSON vs. the `ClusterSecurityException` CRs in `security-exceptions/`; adapt/generate as needed).
2. Read the **new stable CI score** with exceptions applied (a few merge-queue / PR runs), then **ratchet `--compliance-threshold` up** to a floor a few points under it — toward 100 for the genuinely-excepted residual (C-0002 etc.).
3. Update `AGENTS.md` (the Validation "threshold" bullets) to reflect the exceptions wiring and the new floor.

## Why this is separate from the pin-lift PR

The Kubescape compliance score is **environment-dependent and not reproducible offline** (documented in `AGENTS.md`: ≈87% on the Linux CI runner vs ≈94% locally), so the new stable floor can only be established by CI — it can't be validated in a draft locally. Keeping it out of the determinism-pin-lift PR keeps that change clean and CI-verifiable on its own.

## Acceptance criteria

- CI scan step passes the platform's exceptions to `ksail workload scan`.
- `--compliance-threshold` raised to the new observed floor (the excepted residual no longer caps it).
- `AGENTS.md` threshold bullets updated and self-consistent.

*Filed from a holistic platform↔ksail review (the same review that lifted the 7.65.0 determinism pin once [#5371](https://github.com/devantler-tech/ksail/issues/5371) was resolved).*

devantler:
[...]

PR #2439, opened 2026-07-04
fix(kubescape): remount scanner config.json so offline posture scans persist

devantler:

## Why

The in-cluster Kubescape compliance dashboard (Headlamp) is **frozen** — it serves stale data from 2026-06-27 and shows controls like **C-0007 "Roles with delete capabilities" as 0 of 26 passed**. Root cause: an offline-mode bug in the kubescape chart mounts the scanner's config *file* as a directory, so **every scan aborts before it finishes** — no fresh results are stored and none of the platform's by-design posture exceptions are ever applied (aggregate compliance reads 0.00 on every framework).

## What

Reproduces the merged upstream fix ([kubescape/helm-charts#862](https://github.com/kubescape/helm-charts/pull/862), issue [#857](https://github.com/kubescape/helm-charts/issues/857)) with a Flux post-render that remounts the config volume correctly, so scans finalize again, the dashboard refreshes, and the exception CRs already shipped in #2316 finally take effect. Offline/air-gapped mode is untouched — **no cloud submit is re-enabled**. Revert this once the chart ships a release containing the upstream fix.

**Note:** this unblocks the whole dashboard but does not on its own make C-0007 pass — that control's exception needs a separate match-type fix (follow-up PR). After promotion Flux reconciles the HelmRelease; a fresh scan should then persist and the score reappear. Related: #2264.

### DIFF
--- a/k8s/bases/infrastructure/controllers/kubescape/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/kubescape/helm-release.yaml
@@ -142,3 +142,35 @@ spec:
         limits:
           cpu: 500m
           memory: 1Gi
+  # Offline-mode chart bug (kubescape-operator 1.40.2): the scanner Deployment
+  # mounts the empty `kubescape-volume` emptyDir onto the FILE path
+  # /home/nonroot/.kubescape/config.json (subPath config.json). When the target
+  # does not exist kubelet MATERIALISES it as a DIRECTORY, so ksserver's
+  # config-write (GenerateAccountID) fails with `open .../config.json: is a
+  # directory`; the scan aborts at finalize ("scanning failed"), the aggregate
+  # ConfigurationScanSummary is never written, and NO posture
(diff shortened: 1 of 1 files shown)


devantler:

Closing as a duplicate of #2443, and superseded by the same reasoning that closed #2444.

This PR fixes only the `config.json` emptyDir/`subPath` mount (via a Flux postRenderer). That alone is **insufficient**: with the mount fixed, the scheduled scan still posts `Submit=true`, so the v4.0.9 scanner attempts the ARMO cloud submit and aborts with a 402 before persisting any `WorkloadConfigurationScan` — the failure just moves from "config.json is a directory" to the cloud submit. #2443 applies the load-bearing fix — `scanV1.keepLocal: true` in the scheduler request body — which runs the scan local-only (Submit=false, no cloud call, config.json never touched), mirroring upstream helm-charts#862. Deferring to #2443.

(I re-derived this root cause independently before finding #2443/#2444 already covered it — apologies for the duplicate.)

PR #2440, opened 2026-07-04
fix(security): make C-0002 exec-into-container exception effective for infra controllers

devantler:
## Why

The Kubescape compliance dashboard reports **C-0002 (Prevent containers from allowing command execution)** as failing for Velero, CloudNativePG, and the Flux reconcilers — even though these controllers are already listed in our RBAC exception. The exception silently does nothing: C-0002 is an RBAC-review control whose findings are anchored on the RBAC *binding* object, and our entry lived under a `namespaceSelector`, which Kubescape never applies to that class of finding (the same limitation we already worked around for C-0187). So legitimate, by-design `pods/exec` holders keep showing as violations, adding noise to the security posture.

## What

Moves C-0002 to a dedicated exception that matches the specific RBAC bindings by kind+name — the mechanism already proven to work for C-0187 — and mirrors it into the Headlamp dashboard's exception ConfigMap. Scope stays minimal (individual bindings, not shared roles like `cluster-admin`/`edit`), so C-0002 still catches any *new* exec grant. Removes the inert C-0002 entry from the namespaceSelector exception so the config no longer implies a suppression that isn't happening.

The two tenant service accounts (wedding-app, ascoachingogvaner) also surface in C-0002; they're a fixable over-grant rather than a by-design need and are handled at the root in a **separate PR** (scoping their Flux-impersonation role), not excepted here.

Scoped to Velero + CNPG + Flux. Kubescape's live posture pipeline is currently degenerate (separate issue), so the match can't be verified in-cluster pre-merge — confidence is inherited from the identical, working C-0187 exception on the same binding objects.

### DIFF
--- a/k8s/bases/infrastructure/cluster-security-exceptions/controller-rbac.yaml
+++ b/k8s/bases/infrastructure/cluster-security-exceptions/controller-rbac.yaml
@@ -31,19 +31,18 @@ spec:
       action: ignore
     - controlID: C-0031
       action: ignore
-    - controlID: C-0002
-      action: ignore
     - controlID: C-0063
       action: ignore
     - controlID: C-0035
       action: ignore
-    # NOTE: C-0187 (wildcard RBAC, CIS-5.1.3) is handled in wildcard-rbac.yaml,
-    # not here — a namespaceSelector exception does not suppress RBAC findings
-    # (Kubescape keys them on the Role/ClusterRole/binding object, so a namespace
-    # match never applies; proven: the namespaced velero-server Role kept failing
-    # despite `velero` being listed below). Headlamp is fixed at the root (SA
-    # scoped to read-only cluster-reader); Flux and Velero are exempted by-design
-    # in wildcard-rbac.yaml via an explicit kind+name resources match.
+    # NOTE: RBAC-review controls are NOT suppressible here. Kubescape anchors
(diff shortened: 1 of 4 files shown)


devantler:

**CodeRabbit review-body 🧹 Nitpick (exec-into-container-rbac.yaml L23-30) — FIXED.** Verified against the sibling wildcard-rbac.yaml that `cluster-reconciler-flux-system` does bind kustomize-controller as well as helm-controller, and that the Velero match list includes the velero-server RoleBinding. Corrected the covered-list comment, the CSE `spec.reason`, the inline match-list comment (8716d6e), and the mirrored reason text in config-map-headlamp-exceptions.yaml (96c2024).

devantler:

Review-body nitpick resolution record: both wording inaccuracies (Velero RoleBinding missing from the 'Covered' summary; `cluster-reconciler-flux-system` shared by kustomize-controller) were already corrected at HEAD before this sweep — the review ran against an earlier commit. Current `exec-into-container-rbac.yaml` reads 'velero-server RoleBinding/Role … plus its cluster-admin ClusterRoleBinding' and 'kustomize-controller and helm-controller (both bound via cluster-reconciler-flux-system)', and the Headlamp mirror reason names 'Flux kustomize-/helm-controller and flux-operator'. No change needed.

devantler:
@coderabbitai review

PR #2443, opened 2026-07-04
fix(kubescape): keep scheduled posture scans local so results persist

devantler:

## Why
The Kubescape compliance dashboard has been frozen on week-old results — every control shows failing and the score reads 0, even for workloads already exempted by policy (e.g. privilege escalation on longhorn/velero/coroot). Since the scanner was bumped to v4.0.9, the scheduled posture scan has aborted before it can save any results, so the exceptions and fresh findings never surface.

## What
Forces the scheduled scan to run local-only so it stops trying to submit its report to the cloud — the step that crashes on a volume-mount bug in the current chart. Fresh results then persist every run, with the existing exceptions applied, unfreezing the whole dashboard. No functional scanning is lost (this cluster has no cloud account anyway).

Interim override mirroring a merged-but-unreleased upstream chart fix (kubescape/helm-charts#862, issue #857); drop it once that ships in a chart release past 1.40.2.


devantler:

Heads-up on the sequencing: unfreezing scans here is necessary but **not sufficient** to make posture exceptions apply. The scanner's CRD exceptions getter in **v4.0.9 lists the CRs at `kubescape.io/v1`, but the chart's CRDs serve only `v1beta1`** — the List 404s and, as the merged/secondary source, is silently swallowed (kubescape#2366 fixes the version; first shipped in image v4.0.10). So once this PR lands and scans resume, exempted controls would still render as failing under v4.0.9.

Bumping the scanner to v4.0.10 is #2452. The full offline pipeline is #2316 (merged: reader + RBAC) → this (#2443, unfreeze) → #2452 (v4.0.10, read CRs at v1beta1) — after which the per-control authoring PRs (#2442/#2446/#2440/#2434) take effect.

(Note: this corrects the assumption baked into #2316 that v4.0.9's getter reads the right version — it doesn't.)

PR #2444, opened 2026-07-04
fix(kubescape): mount scanner config.json dir so posture scans persist

devantler:

## Why
The cluster's security scanner (Kubescape) has silently stopped producing fresh results — posture/compliance data has been **frozen since 27 Jun**. Every scan now fails on a broken config-file mount, so the dashboard shows **stale findings** and the platform's security **exceptions never take effect**. This surfaced while fixing a real credential finding (dex) — the dashboard couldn't reflect the fix because scanning itself was dead.

## What
A small mount correction lets the scanner write its config file again, so daily posture scans persist and exceptions apply. **No behaviour or resource change** — it only unblocks scanning. Once this deploys and re-scans, the compliance dashboard reflects reality again.

_Pairs with (doesn't depend on) the scanner v4.0.9 bump that exposed this, already merged in #2316._

### DIFF
--- a/k8s/bases/infrastructure/controllers/kubescape/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/kubescape/helm-release.yaml
@@ -142,3 +142,38 @@ spec:
         limits:
           cpu: 500m
           memory: 1Gi
+  # Fix the scanner's config.json mount so posture scans actually persist.
+  # In OFFLINE mode (kubescapeOffline: enable) the chart mounts the writable
+  # `kubescape-volume` emptyDir at the FILE path /home/nonroot/.kubescape/config.json
+  # with `subPath: config.json`. An emptyDir + subPath makes Kubernetes create the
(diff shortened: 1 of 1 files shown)


devantler:

Closing as a duplicate of #2443, which fixes the same frozen-posture-scan regression more completely.

This PR only corrects the `config.json` emptyDir+`subPath` mount (directory → writable file). But #2443's investigation shows that alone is **insufficient**: even with the mount fixed (equivalently, `kubescapeOffline: disable`), the scheduler still posts `Submit=true`, so the v4.0.9 scanner tries to submit to ARMO cloud and aborts with a 402 before persisting any `WorkloadConfigurationScan` — the failure just moves from "config.json is a directory" to a cloud-submit error. #2443 forces `scanV1.keepLocal: true` in the scheduler request body (Submit=false → the scanner never touches cloud or `config.json`), which mirrors upstream kubescape/helm-charts#862 and is the load-bearing fix.

Superseded by #2443.

PR #2445, opened 2026-07-04
feat(kubescape): route runtime-detection alerts to Headlamp, Slack, and Coroot

devantler:

## Why
The Headlamp Kubescape plugin's **Runtime Detection → Alerts** page shows *"Alertmanager URL is not configured"*, and Kubescape's runtime-detection alerts (rule violations, malware) were flowing nowhere. That tab reads only from a Prometheus Alertmanager, which the Coroot migration removed from the cluster — so there was no source to point it at.

## What
Reintroduces a single, tiny, **Kubescape-scoped** Alertmanager (prod-only; not a return of the old Prometheus stack) and points the node-agent at it, so runtime alerts now reach all three intended places: the **Headlamp plugin**, **Slack** (the existing shared webhook), and **Coroot** (via the node-agent's stdout, which Coroot's log capture surfaces).

## Operational notes
- **One manual step:** the Headlamp plugin's Alertmanager address is a per-browser setting that can't be seeded declaratively ([headlamp#3979](https://github.com/kubernetes-sigs/headlamp/issues/3979)). Set it once per operator to `kubescape/alertmanager:9093`. Until then the data source exists but the tab stays empty. Documented in `docs/dr/alerting.md`.
- **New dependency:** the prometheus-community Alertmanager Helm chart.
- Needs a direct merge after promotion (trusted-author PR).

### DIFF
--- a/docs/dr/alerting.md
+++ b/docs/dr/alerting.md
@@ -90,6 +90,45 @@ stays quiet by design, exactly as the old Alertmanager did.
   container logs/traces, not host audit-log files, so the previous
   alloy-audit → Loki pipeline was removed.
 
+## Kubescape runtime-detection alerts
+
+The Kubescape node-agent's **runtime-detection** alerts (rule violations,
+malware) are the one signal that does **not** fit the Coroot model, because their
+only first-class dashboard — the **Headlamp Kubescape plugin's "Runtime
+Detection → Alerts" tab** — reads exclusively from a Prometheus **Alertmanager**
+(`GET /api/v2/alerts`, filtered on `alertname="KubescapeRuleViolated"`). It
+cannot read Kubescape's storage CRs and cannot query Coroot's Prometheus (a
+metrics store, not an Alertmanager). So a **single, minimal Alertmanager** is
+reintroduced *scoped to Kubescape* — not a re-adoption of the Prometheus stack.
+It lives in the `kubescape` namespace, prod-only
+(`providers/hetzner/infrastructure/controllers/alertmanager/`), ~10m CPU / 32Mi
+RAM, ephemeral (no PVC).
+
+The node-agent fans each alert out to **all three destinations** (wired in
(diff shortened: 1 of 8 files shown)


devantler:
@coderabbitai review

devantler:

Review-body nitpick resolution record (no threads exist for these):
- **automountServiceAccountToken** — valid, fixed in dced919 (`false` at chart top level; this Alertmanager never calls the Kubernetes API).
- **`.invalid` fallback should fail / surface** — refuted as-designed: the `${alertmanager_webhook_url:=…invalid}` default is deliberate. Flux postBuild substitution failing on a missing var would block the ENTIRE hetzner infrastructure Kustomization (every controller), which is a far worse failure mode than one alert route going quiet; the placeholder keeps the blast radius at 'Slack delivery off'. The `alertmanager_notifications_failed_total` scrape/alert idea is a fair follow-up for the Coroot integration but out of scope for this PR.

devantler:

Resolution record for the CodeRabbit review-body nitpicks (2026-07-04 12:27Z review — no inline threads exist for these):

1. **Silent `${alertmanager_webhook_url:=…invalid}` fallback** — keeping as-is, by design. The inline default is the repo-wide Flux-substitution convention that keeps `ksail workload validate` (no SOPS access) and the local/CI overlays building; a strict/hard-fail substitution would wedge the entire hetzner infrastructure Kustomization on one missing variable (the exact blast-radius class of the 2026-07-02 prod wedge). Misconfiguration is not silent in practice: the variable is the SAME one Coroot incidents and the Flux notification-controller already post to (a rename breaks visibly in three systems), the RFC-2606 `.invalid` placeholder makes a bad substitution obvious in the rendered config, and Alertmanager logs failed notifications — which Coroot's log capture ingests via the same channel as the node-agent stdout alerts.
2. **`automountServiceAccountToken: false`** — already in the HelmRelease values (helm-release.yaml, "never calls the Kubernetes API" block); the finding is stale against the current head.

PR #2452, opened 2026-07-04
fix(kubescape): bump posture scanner to v4.0.10 so exception CRs apply offline

devantler:

## Why
The compliance dashboard flags readiness/liveness (and the RBAC/SA/admission) controls as failing on infrastructure and batch workloads we have **already formally accepted** as false-positives. Root cause: the in-cluster posture scanner never actually reads our exception CRs. #2316 bumped the scanner to v4.0.9 to gain the exception reader, but v4.0.9 looks the CRs up at the **wrong API version** (v1, while our CRDs serve v1beta1), so it silently applies zero exceptions. This is the keystone: it's why none of the in-flight kubescape exception work (#2442, #2446, #2440, #2434, and the existing health-probes rules) visibly takes effect.

## What
- Bumps the posture scanner one more patch (v4.0.9 → v4.0.10), which fixes the API-version bug so the scanner reads and applies our exception CRs.
- Adds one rule for run-to-completion Jobs/CronJobs (probes don't apply to batch pods), closing the only readiness gap the per-namespace rules missed (openbao's snapshot job).

## Depends on
- **#2443** — offline scans are currently frozen (nothing has persisted since 06-27), so this bump only becomes visible once #2443 lands. Together with the merged #2316, the two complete the offline exception pipeline.

## Notes
- Scanner tag is a one-patch-ahead-of-chart override (same pattern as node-agent); reverts once the chart ships v4.0.10.
- Own PR: needs a **direct merge** after promotion.

### DIFF
--- a/k8s/bases/infrastructure/cluster-security-exceptions/batch-workloads.yaml
+++ b/k8s/bases/infrastructure/cluster-security-exceptions/batch-workloads.yaml
@@ -1,0 +1,35 @@
+---
+# Exempt the health-probe controls (C-0018 readiness, C-0056 liveness) for
+# run-to-completion batch workloads. A Job/CronJob pod runs to completion and
+# never serves traffic, so readiness/liveness probes are semantically
+# inapplicable — Kubescape nonetheless flags every Job/CronJob as missing them.
+#
+# Matched by workload KIND (not namespace) so it covers current and future batch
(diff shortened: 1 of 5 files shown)


devantler:
@coderabbitai review

devantler:
@coderabbitai review

PR #2454, opened 2026-07-04
fix(kubescape): repair offline config.json mount so posture scans persist

devantler:

**Why:** The Kubescape compliance dashboard has been stuck at 0 with a wave of false "network policy" failures (C-0030 *Ingress and Egress blocked*, plus C-0054/C-0260) since 2026-06-27. Root cause: the posture scanner has been silently failing to *save* its results on every scan, so the platform's existing cluster-wide exemption for those controls — we enforce network segmentation with CiliumNetworkPolicies, which Kubescape doesn't recognise as native NetworkPolicy — never actually gets applied to the live results.

**What:** Fixes the scanner's config-file mount (a chart 1.40.2 bug that only triggers in offline/air-gapped mode) so it can write its state and persist results again, restoring live compliance scoring and the C-0030/C-0054/C-0260 exemption. Keeps offline mode and the scanner version the exemption feature needs. Mirrors the fix already on the chart's unreleased upstream `main`; a temporary workaround to remove once a chart > 1.40.2 ships it.

Hotfix for broken posture scanning (no prior issue). Needs a direct merge after promotion (own PR — auto-merge is bot-only).

### DIFF
--- a/k8s/bases/infrastructure/controllers/kubescape/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/kubescape/helm-release.yaml
@@ -142,3 +142,41 @@ spec:
         limits:
           cpu: 500m
           memory: 1Gi
+  # WORKAROUND — chart 1.40.2 breaks the posture scanner under `kubescapeOffline:
+  # enable` (above). In that mode the chart mounts the EMPTY `kubescape-volume`
+  # emptyDir directly AT the file path /home/nonroot/.kubescape/config.json
+  # (subPath config.json); kubelet auto-creates a non-existent subPath as a
+  # DIRECTORY, so config.json is a directory, not a file. The v4.0.9 scanner's
+  # account-ID step then fails EVERY scan with "open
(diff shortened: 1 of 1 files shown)


devantler:

Closing as a **duplicate** of #2443 (and of the already-closed mount-only attempts #2439 / #2444).

I independently re-root-caused the C-0030 "0 of 16" false-positive to the frozen posture scanner and shipped a post-render fix for the broken `kubescapeOffline` config.json mount — but **the mount fix alone is insufficient**, for the same reason #2439/#2444 were closed: `KS_OFFLINE`/offline mode does **not** set `Submit=false`, so the scheduled scan still submits to ARMO cloud and aborts with a 402 (no account) even once config.json is writable.

The load-bearing fix is #2443's `keepLocal: true` in the scheduler request → `Submit=false` → `NewReportMock` → config.json is never opened at all, so results persist. Full chain: #2316 (merged: v4.0.9 getter + scanner RBAC) → **#2443** (keepLocal unfreeze) → **#2452** (v4.0.10 so the getter lists the CRs at the right apiVersion). No new PR needed here.

PR #2429, opened 2026-07-04
fix(policy): trust the actions publish-app identity for app-image verification

devantler:
> 🤖 Generated by the Daily AI Assistant

## Why
After a repo repoints its publish workflow from `reusable-workflows` to the `actions` monorepo (reusable-workflows is being archived — e.g. wedding-app #149), its app image is signed by `actions/.github/workflows/publish-app.yaml`. The `verify-app-images` policy only trusted the `reusable-workflows` identity, so **Kyverno denied every new app ReplicaSet at admission** — wedding-app v1.14.1 is stuck, old pod still serving, and every other app will hit this as it repoints.

## What
Adds the `actions` publish-app cosign identity alongside the `reusable-workflows` one in the `publishapp` attestor (identities are OR-d), so images signed by either admit during the transition. Verified live: the blocked v1.14.1 image validates against the actions identity.

PR #2456, opened 2026-07-04
fix(kyverno): use one alternated publish-app identity so app-image verification admits again

devantler:

## Why

Every new first-party app pod (wedding-app v1.14.1 and anything the `publishapp` attestor gates) has been denied at admission since #2429 deployed, wedging the whole merge queue — every `deploy-prod` run fails its `apps` health check and evicts. The real root cause (verified in the live kyverno logs) is not the webhook timeout: cosign v3 bundle verification rejects an attestor holding **two** identity entries outright ("multiple identities are not supported"), so #2429's transition attestor fail-closed everything it gates.

## What

Collapses the two `publish-app` identities into one identity with an alternated `subjectRegExp` — the same OR, in the shape the tenants' Flux OCIRepository verify blocks already use, verified working against the blocked wedding-app image (`cosign verify` exit 0). Documents the single-identity constraint in the policy so the next identity change doesn't repeat this.

Note: this PR itself must cross the wedged queue; if its own `merge_group` deploy gets evicted on the wedding-app health check, a direct/admin merge is the unwedge (same circularity #2429 hit).

### DIFF
--- a/k8s/bases/infrastructure/cluster-policies/best-practices/verify-app-images.yaml
+++ b/k8s/bases/infrastructure/cluster-policies/best-practices/verify-app-images.yaml
@@ -52,14 +52,21 @@ spec:
         keyless:
           # The publish-app workflow moved from reusable-workflows into the
           # actions monorepo (reusable-workflows is being archived); both
-          # identities are accepted during the transition so images signed by
-          # either repo's publish-app.yaml admit. Identities within one attestor
-          # are OR'd, so a match against either satisfies verification.
+          # identities are accepted during the transition via ONE identity
+          # whose subjectRegExp alternates the repo — the same shape as the
+          # tenants' Flux OCIRepository verify blocks. Do NOT list a second
+          # identity entry instead: the IVPOL engine verifies cosign v3
+          # bundles through sigstore-go, which rejects more than one identity
+          # per verification ("unsupported: multiple identities are not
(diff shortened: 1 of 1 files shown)


devantler:

**One-click unwedge (my live patch attempt was permission-denied).** Either works; (1) heals prod immediately and this PR then merges through a healthy queue:

**Option 1 — patch the live policy now** (identical to this PR's change; Flux converges to git state afterwards, and the 60m image-verify cache keeps wedding-app admitting meanwhile):

```bash
kubectl --context admin@prod patch imagevalidatingpolicy verify-app-images --type=json -p='[{"op":"replace","path":"/spec/attestors/0/cosign/keyless/identities","value":[{"issuer":"https://token.actions.githubusercontent.com","subjectRegExp":"^https://github\\.com/devantler-tech/(actions|reusable-workflows)/\\.github/workflows/publish-app\\.yaml@.+$"}]}]'
```

**Option 2 — promote this PR and, if its `merge_group` deploy is evicted on the wedding-app health check (the #2429 circularity), direct/admin-merge it.**

Verification already done: `cosign verify` of the blocked image (`wedding-app@sha256:65d3eef4…`) against the alternated regexp exits 0; the live kyverno error ("unsupported: multiple identities are not supported at this time") is the cosign v3 single-identity constraint, sigstore/cosign `pkg/cosign/verify.go` `verificationOptions()`.

PR #2462, opened 2026-07-04
feat(talos): WireGuard server listener on control planes (WS3 phase 1)

devantler:

**WS3, phase 1 of the "VPN in front of critical services" rollout.** Adds the dial-in WireGuard **server** on the Talos control planes — the encrypted endpoint the home UniFi gateway (the client side, [unifi#9](https://github.com/devantler-tech/unifi/pull/9), now on Crossplane) connects to.

## What this adds
- **`talos/control-planes/wireguard.yaml`** — `wg0` at `10.200.0.1/24`, `listenPort 51820`, **no peers yet**: the gateway's key pair only exists once the unifi tenant applies, and the tenant is sequenced behind this server landing — so the peer registration follows in #2473. A peer-less listener is safe (WireGuard answers no unauthenticated packet, nothing can connect). The server private key is `${WG_SERVER_PRIVATE_KEY}` env-expanded by ksail — never committed; the `prod` environment secret is set and threaded through the deploy-prod composite (ci+cd) and the DR rebuild.
- **`talos/control-planes/ingress-firewall.yaml`** — one world-reachable `51820/udp` rule (WireGuard never replies to an unauthenticated packet, so an open source is safe; 5 rules total, well under the ENOBUFS threshold).

## Scope — this is the LISTENER only
No tunnel comes up from this PR alone: the gateway peer (#2473) and onward routing/enforcement are follow-on. Merging is safe and unblocks the tenant side, which waits for this server to exist.

## Safety
- **Config-only, not auto-applied.** ksail's `cluster update` doesn't push machine-config changes to already-running nodes, so merging this changes nothing live. It makes every node correct-by-construction on create/recreate; to apply to the running control planes, patch once out-of-band (a no-reboot field):
  `talosctl --nodes <cp-ips> patch mc --patch @wireguard.yaml` (with `${WG_SERVER_PRIVATE_KEY}` filled).

## Next (needs a design decision — will follow separately)
Making the admin UIs (Coroot/Hubble/OpenCost/Longhorn/OpenBao/Headlamp/KSail) reachable **only** through the tunnel is non-trivial here: the cluster fronts them on a **public Hetzner LB** (no Cilium LB-IPAM/L2-announcement), and **iptables masquerade** rewrites tunnel-sourced traffic to the node CIDR — so the clean "internal VIP" and "source-CIDR NetworkPolicy" approaches don't work as-is. The enforcement approach will be a separate PR once we've picked a direction and can validate the datapath against the live tunnel.


devantler:

**merge_group root-cause (run 28716734943, evicted 19:12Z) — do NOT re-queue yet; the failure is deterministic.**

Talos rejected the control-plane config apply on all 3 CPs with `private key is invalid: wrong key "" length: 0` / `public key invalid: wrong key "" length: 0` — i.e. `${WG_SERVER_PRIVATE_KEY}` and `${WG_GATEWAY_PUBLIC_KEY}` expanded to **empty strings**. The reject was atomic (InvalidArgument at validation; nothing applied, prod unaffected) and the queue evicted the PR.

Two gaps have to close before this can merge:
1. **The secrets aren't seeded yet** — this PR's own key-exchange order (generate server pair → tenant peer pubkey → tenant apply mints the gateway key → set `WG_GATEWAY_PUBLIC_KEY`) hasn't happened; both are maintainer-side secret writes.
2. **The deploy path can't deliver them anyway**: the `🚀 Deploy to Production` step passes a finite input set (sops-age-key / kube-config / talos-config / ghcr-token / hcloud-token) — there is no plumbing for the two WG env vars, so even seeded secrets would still expand empty at ksail load time. The deploy action + `ci.yaml` need the two inputs added (and the same for any future `${...}` env-expanded Talos patch).

Suggested order: ship the plumbing commit on this branch first, then seed the secrets, then re-queue.

devantler:

**Merge-queue eviction root-caused and fixed.** The 19:10Z `merge_group` deploy failed in 🚀 Deploy to Prod: Talos rejected the rendered machine config on all control planes with `private key is invalid: wrong key "" length: 0` — neither `WG_*` env var existed in CI, so ksail expanded them to empty strings.

Fixed by `028ce953`:
1. **Peer split (deadlock break):** the gateway's public key cannot exist until the unifi tenant's `cluster_wireguard` applies, and that tenant work is paused *until this server lands* (unifi#14) — requiring `${WG_GATEWAY_PUBLIC_KEY}` here deadlocked both sides. The listener now ships **peer-less** (safe: WireGuard answers no unauthenticated packet); peer registration is #2473.
2. **Secret + wiring:** generated the server key pair, set `WG_SERVER_PRIVATE_KEY` in the `prod` environment, and threaded it through the deploy-prod composite (ci + cd) and dr-rebuild's `cluster create`. The private key was piped straight into the secret and never logged.

**Server public key** (for the tenant's `cluster_wg_peer_public_key`, #2473): `zEJdr4QdIZBpGUPKzN7NF9Oidaw1lSstPqoQpIdNTUA=`

Will re-queue once CI is green.

devantler:

**CodeRabbit outside-diff finding (pre-merge Talos render/validate step) — valid, captured as #2477.** Adding that CI capability is out of scope for this PR (it's a new PR-event validation lane for all `talos/**` patches, needing placeholder key-material design), so it joins the backlog rather than growing this diff. The concrete failure it would have caught is fixed here directly (secret set + threaded; peer-less listener).

PR #2476, opened 2026-07-04
docs(actual-budget): manage E2EE password in OpenBao + document manual enable

devantler:

**Why:** We want Actual Budget's end-to-end encryption managed the same declarative way as everything else. But Actual's E2EE is client-side and password-derived by design — the server never sees the key — so there is no server switch to enable it, and there shouldn't be. The real risk to manage is a lost password: it makes the budget permanently unrecoverable.

**What:** Declares the E2EE password as a root of trust — seeds an OpenBao path (`apps/actual-budget/encryption`, create-only placeholder, mirroring the Enable Banking seed) as the durable DR store, and documents the one-time manual "Enable encryption" step in the app README plus a custody entry in the DR crypto-custody doc. Nothing in-cluster consumes the password (only a client can use it); OpenBao is purely the record so it can't be lost.

No behaviour change to the running app — seed + docs only.


devantler:

@coderabbitai review

Issue #2450, opened 2026-07-04
fix(kubescape): kubevuln produces no CVE data — relevancy ScanCP aborts on partial ApplicationProfiles


devantler:
> 🤖 Generated by the Daily AI Assistant

Part of #2447.

## Problem

kubevuln is alive and processing its queue (grype DB loads fine at startup; thousands of `scan started`), but **no image carries any vulnerability data.** Verified against live prod (2026-07-04):
- `vulnerabilitymanifestsummaries`: **109/109** show `.all`=0 **and** `.relevant`=0, with `vulnerabilitiesRef.all` and `.relevant` both empty (`kind:""`, `name:""`).
- `vulnerabilitymanifests`: **197/197** have zero `payload.matches` and no `tool.name` — SBOM shells with no grype match phase ever recorded.
- Dominant kubevuln log (~815×/24h): `service warning - ScanCP … container profile is partial (workload restart required)`.

`ScanCP` is the **scan-with-container-profile (relevancy)** path — and it is the only path wired. It aborts before grype runs because the ApplicationProfile never reaches `completed` (123/288 profiles are still `partial`), so **neither** `.all` **nor** `.relevant` is ever written. Real registry errors (`ScanCVE` i/o timeout on `postgres:latest`) are rare (3×/24h) — this is **not** a grype-DB or registry-auth failure.

**Consequence:** the cluster is **not** genuinely 0-CVE, and #2149 ("reduce reachable CVEs *flagged by Kubescape*") currently has nothing to act on — nothing is being flagged.

## Proposed direction (CVE = both — settled with the maintainer)

1. **Get grype output flowing now** — decouple/enable a full non-relevancy scan (the `.all` view) so grype produces CVE matches regardless of ApplicationProfile completion state. Confirm the kubevuln/node-agent config that governs the relevancy-vs-full scan path (relevancy on but not gating the `.all` write).
2. **Complete the profiles for the reachable view** — a controlled, batched rollout of the 123 partial-profile workloads so `ScanCP` completes and the `.relevant` view (the ~90%-noise-reduction reachable-CVE list) populates. Batch it; do not mass-restart prod at once.
3. Once data flows, enable **VEX generation** (`capabilities.vexGeneration`) so a downstream CI gate can suppress non-reachable CVEs (feeds #2149 and the reachable-CVE CI gate).

## Rough size
M (investigation of the scan-path config is the main unknown; the profile rollout is operational and staged).

## Acceptance criteria
[...]

PR #2494, opened 2026-07-05
feat(kubescape): enable VEX generation for reachability-aware CVE triage

devantler:

## Why
The CVE pipeline now produces real vulnerability data, but nothing downstream can tell which CVEs are actually reachable — the noise-reduction verdicts exist only inside Kubescape's own summaries.

## What
Enables Kubescape's VEX generation, so each scanned image gets a standard OpenVEX document (stored in-cluster) marking CVEs not-affected when runtime observation shows the vulnerable component is never loaded. This gives the planned reachable-CVE CI gate a machine-readable suppression source.

Part of #2450.


devantler:

CodeRabbit thread resolution record: valid trivial point — added the upstream-experimental note to the capability comment (61dc43f), with the blast-radius rationale (VEX CRs are advisory data only; a regression cannot break scanning itself).

Issue #2438, opened 2026-07-04
Track fixed Headlamp Safari OIDC release


devantler:
> 🤖 Generated by the Agentic Engineer

**Blocker:** kubernetes-sigs/headlamp#6355 | last-verified 2026-08-21: PR OPEN; latest release v0.45.0; first release containing a verified fix is UNKNOWN

## Purpose

Track the upstream Safari OIDC popup regression from fix merge through a published Headlamp release and real platform verification. The local rollback restores service; it does not prove a later Headlamp release safe or retire this tracker.

## Current mitigation

- Platform PR #3292 rolls Headlamp back to the last Safari-verified chart, 0.42.0.
- Renovate is hard-pinned to 0.42.0 with automerge disabled.
- Headlamp v0.45.0 is the latest published release as of 2026-08-21, while the named upstream fix PR remains open. Equivalent-fix inclusion is not inferred.

## Exit criteria

- [ ] Upstream [Headlamp #6355](https://github.com/kubernetes-sigs/headlamp/pull/6355) merges, or the exact equivalent fix commit is identified.
- [ ] A Headlamp chart release containing that fix is published; record the exact release and commit evidence here.
- [ ] A focused platform PR bumps from 0.42.0 while keeping automerge disabled and passes current-head CI and review.
- [ ] Safari login is verified end-to-end: successful authentication redirects the opener, refresh preserves the session, and reopening the main URL is not required.
- [ ] Only after that verification, relax the exact Renovate pin and close this issue.

## Upstream

- Regression: [kubernetes-sigs/headlamp#6353](https://github.com/kubernetes-sigs/headlamp/issues/6353)
- Proposed fix: [kubernetes-sigs/headlamp#6355](https://github.com/kubernetes-sigs/headlamp/pull/6355)
- Current mitigation: #3292

devantler:
> 🤖 Generated by the Daily AI Engineer

Maintainer hit this again today (2026-07-17) and directed a course change, superseding the 2026-07-04 stay-on-0.43.0 decision: **revert the chart to 0.42.0 now** and **gate Renovate off exactly 0.43.0**, since upstream fix kubernetes-sigs/headlamp#6355 has been stalled unmerged since 2026-07-05 with no release since v0.43.0.

[...]

PR #2497, opened 2026-07-05
feat(crossview): embed in Headlamp again and drop the public route

devantler:

## Why

The OIDC issue that forced Crossview out of Headlamp (#2268) is fixed upstream in the latest Crossview release, so the standalone public dashboard at crossview.platform.devantler.tech is no longer needed.

## What

Bumps Crossview to 4.5.0, re-adds the crossview-headlamp plugin so Crossview is surfaced inside Headlamp again, and removes the public URL entirely (route, homepage tile, and SSO callback repointed accordingly). The Headlamp plugin's port-forward is now the only entry point.

**Verify after promotion:** open Crossview from Headlamp on prod and complete an OIDC sign-in once — the callback now rides the plugin's port-forward.


devantler:

Technical notes for the prod verification (kept out of the PM body):

- **The Dex redirect URI is exact-match `http://localhost:3001/api/auth/oidc/callback`**, so the sign-in only completes when the plugin's port-forward lands on local port 3001 (it requests the service target port; if the port is taken, Headlamp picks another and Dex will reject the callback). If that bites in practice, the fallback is registering a small set of fixed ports or teaching the plugin a deterministic port — flag it and I'll follow up.
- **Safari won't persist the session cookie** on http://localhost (NODE_ENV=production marks it Secure; Chromium treats localhost as trustworthy, Safari does not) — same use-Chrome guidance as the existing Headlamp OIDC note (#2438).
- The inert `crossview-ingress` from the pre-#2268 era is deliberately NOT restored: the plugin resolver would embed its host verbatim, which is exactly the public URL this PR removes.
- Validation: `bases/apps/crossview` + headlamp/homepage/dex all render and validate green under ksail 7.141.0 on both overlays; the only local failures were the known kubeconform#363 corruption race and a flaky actual-budget chart-render (`/login/openid` schema — non-deterministic, also reproduces on clean origin/main, unrelated to this diff).

PR #2498, opened 2026-07-05
fix(crossview): grant the OIDC user port-forward in the crossview namespace

devantler:

## Why

Opening Crossview from Headlamp fails with "Unreachable" for the OIDC web session (your 14:25 screenshot). #2497 made the Headlamp plugin's port-forward the only entry point, and Headlamp authorizes port-forwards against the signed-in user's own RBAC — but the OIDC identity is read-only cluster-wide by design, so the check is denied (verified in the headlamp backend log and by impersonation). It passed pre-merge verification because admin credentials don't hit the user-RBAC path.

## What

Grants `create pods/portforward` to the OIDC user, scoped to the crossview namespace only — the tunnel to the dashboard mutates nothing, so the cluster-wide read-only posture stays intact.

Follow-up to #2497.

### DIFF
--- a/k8s/bases/apps/crossview/kustomization.yaml
+++ b/k8s/bases/apps/crossview/kustomization.yaml
@@ -7,3 +7,5 @@ resources:
   - helm-release.yaml
   - external-secret.yaml
   - cilium-network-policy.yaml
+  - role.yaml
+  - role-binding.yaml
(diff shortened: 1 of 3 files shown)


devantler:

Post-merge verification on prod (13:17Z): the `crossview-portforward` Role + RoleBinding are live, and `kubectl auth can-i create pods --subresource=portforward -n crossview --as=oidc:ned@devantler.tech` now returns **yes** — the RBAC gate that made "Open Crossview" unreachable is cleared. Retrying the Headlamp plugin button should now open the port-forward.

(Verification note: the `can-i create pods/portforward` *shorthand* falsely reports "no" on kubectl v1.36 — it parses the subresource as a resource name; use `--subresource=portforward`. `can-i --list` confirms the grant either way.)

PR #2499, opened 2026-07-05
fix(crossview): grant get on pods/portforward for the WebSocket streaming upgrade

devantler:

## Why
"Open Crossview" still fails with *Unreachable* after #2498 (your 15:41 screenshot). The Headlamp log shows why: the plugin's permission pre-check and #2498's Role use the `create` verb, but the port-forward the backend then actually performs is a WebSocket upgrade the API server authorizes as `get` on the same subresource — so the check passes and the forward itself is denied.

## What
Adds `get` alongside `create` on `pods/portforward` in the crossview-scoped Role — same least-privilege scope as #2498, one verb wider. Verified the denial signature in the live Headlamp log (13:41Z); after merge, retrying the plugin button should complete the forward.

Follow-up to #2498.

### DIFF
--- a/k8s/bases/apps/crossview/role.yaml
+++ b/k8s/bases/apps/crossview/role.yaml
@@ -2,12 +2,15 @@
 # Least-privilege RBAC for opening Crossview through Headlamp: since the
 # public route was dropped (Crossview 4.5.0 re-embed), the Headlamp plugin's
 # port-forward is the ONLY entry point, and Headlamp's port-forward handler
-# authorizes it against the signed-in user's own RBAC (`create` on
-# `pods/portforward`). The OIDC identity is deliberately read-only cluster-wide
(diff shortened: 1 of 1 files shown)


devantler:

Live evidence for the verb gap (prod, pre-merge): `can-i create pods --subresource=portforward` = **yes** (the #2498 grant, what the plugin pre-checks) while `can-i get pods --subresource=portforward` = **no** — exactly the verb the 13:41Z Headlamp denial names (`cannot get resource \"pods/portforward\"` during the streaming upgrade). This PR closes that gap.

devantler:

Post-merge verification on prod: the Role now carries `[create, get]` and `can-i get pods --subresource=portforward -n crossview` = **yes** for the OIDC identity — both the pre-check verb and the streaming-upgrade verb are granted. "Open Crossview" should now complete the forward; if anything still fails it's a new layer (next suspect: how the web deployment exposes the forwarded port to a remote browser), not RBAC.

PR #2538, opened 2026-07-09
fix(ci): cache kubeconform schemas in Validate Manifests job

devantler:

## Why
A portfolio-wide `raw.githubusercontent.com` 429 rate-limit is currently failing kubeconform schema fetches across ksail and platform CI. This repo's own `🧪 Validate Manifests` job (`ksail workload validate`) is one of the affected checks, hitting the same rate limit on repeated cold fetches.

## What
Adds an `actions/cache` step that persists `~/.cache/ksail/kubeconform` — the schema-cache directory the `ksail` binary already builds internally — across CI runs. Companion fix to [devantler-tech/actions#476](https://github.com/devantler-tech/actions/pull/476) (same fix for ksail's own Test/Coverage jobs); this one is independent and stands alone.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,10 +98,26 @@ jobs:
           sudo install /tmp/ksail /usr/local/bin/ksail
           ksail --version
 
+      - name: 📥 Restore kubeconform schema cache
+        # ksail's kubeconform client fetches JSON schemas from
+        # raw.githubusercontent.com/yannh/kubernetes-json-schema on every cold run
+        # and has nowhere to persist them across ephemeral CI runners — heavy,
+        # repeated traffic across every PR is what tripped upstream 429
+        # rate-limiting portfolio-wide (2026-07-09). This restores (and, at job
+        # end, re-saves) the local cache dir the client already builds internally
+        # (`validator.Opts.Cache`); no ksail-side change needed.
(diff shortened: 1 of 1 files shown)


devantler:

Re: the `⚠️ Outside diff range comments` finding on the kubeconform cache key (`.github/workflows/ci.yaml` restore/save, `github.run_id`-suffixed primary key) raised across the last two review passes — evaluated, not applying the "drop `run_id`" suggestion:

This is the deliberate GitHub Actions "always-refresh" cache pattern (unique primary key + `restore-keys` prefix fallback), not an oversight — the in-code comment already flags the trade-off. Switching to a **stable** key (`kubeconform-schema-cache-v1`, no `run_id`) would make the save step's `cache-hit != 'true'` guard actually skip on every run *after* the first, since the exact key would then always be found on restore. That would freeze the cache's schema coverage at whatever was fetched on the very first run — any NEW schema kind needed later (e.g. a CRD introduced going forward) would cold-fetch from `raw.githubusercontent.com` on *every* subsequent run forever, since there'd be no way to persist an addition to an already-existing immutable cache key. That's a worse outcome than the storage growth this flags, given this is exactly the class of rate-limit failure the PR exists to fix.

CodeRabbit's own note calls this "not a blocker given the small expected schema payload size" — agreed; the kubeconform schema catalog subset ksail touches is small (tens of KB–low MB), so LRU eviction pressure on the repo's 10GB cache budget is negligible. Leaving the run_id-keyed design as-is.

PR #2580, opened 2026-07-10
fix(policy): refresh replica-floor exemptions as skips

### LINKED ISSUE #2573: Prune stale Kyverno results after policies stop matching

## Problem

The live PolicyReports retain 11 historical `validate-replica-floor` failures for resources that no longer match the policy. Their result timestamps are from June 16-19, while the same report objects continue receiving fresh results on July 10.

The clearest proof is `keda-operator-metrics-apiserver`: source and live state are 2/2 after #2122, yet its old one-replica failure remains stored. The other retained entries now have explicit, documented singleton exemptions. This makes the dashboard overstate 14 current HA gaps when only the three Policy Reporter deployments are still actionable.

## Proposed direction

Reproduce the report-controller retention behavior, identify the supported refresh/prune mechanism when a policy stops matching a resource, and automate or document a safe reconciliation path. Do not add more exemptions or delete whole reports blindly: unrelated current results share the same objects.

## Acceptance criteria

- [ ] A minimal reproduction demonstrates why results survive after a rule stops matching.
- [ ] The supported refresh/prune path removes only obsolete results and preserves unrelated current results.
- [ ] The KEDA and documented singleton entries no longer retain replica-floor failures.
- [ ] After the Policy Reporter HA issue is deployed, the expected live replica-floor failure count is zero.
- [ ] The safe refresh path is automated or documented for future policy/exemption changes.

devantler:

## Why

Production PolicyReports still contain 11 `validate-replica-floor` failures
dated June 16-19 even though KEDA is now 2/2 and the other workloads are
explicitly exempt. Unrelated results in the same reports continue to refresh,
so the dashboard is retaining historical failures rather than showing current
compliance.

The reproduction isolated the policy shape: `rule.exclude` emits no replacement
result after an exemption is added, while a false precondition emits a current
`skip` for the same policy, rule and resource.

## What

- moves every workload-label, pod-label, namespace, exact-name and `*-primary`
  exemption into preconditions so hourly background scans emit `skip`;
- preserves all existing exemption values and keeps the policy in `Audit`;
- adds a seven-case Kyverno regression test covering fail, pass and every
  exemption form;
- documents the safe GitOps rollout/read-only verification path and the generic
  Kyverno 1.18.1 limitation for other match/policy identity changes.

No report is deleted or patched, and the reports controller is not restarted.
Unrelated results remain controller-owned. A feature flag is not applicable:
this fixes derived report bookkeeping without enabling new platform behavior or
changing admission enforcement.

## Validation

- red before the patch: workload-label, namespace, exact-name and wildcard
  exemptions produced no `skip`; normal singleton/HA and the existing
  pod-template precondition behaved as expected;
- green after the patch: `kyverno test tests/validate-replica-floor
  --require-tests` passes all 7 cases;
- `ksail workload validate` with KSail 7.165.1: 536 files validated;
- `ksail --config ksail.prod.yaml workload validate`: 536 files validated;
- `ksail workload scan --framework nsa --compliance-threshold 85` passed;
- embedded JSON, naming, YAML parsing, Kustomize local/prod builds and
  `git diff --check` passed;
- independent review found no omitted exemption or remaining blocker.

## Post-merge verification

After Flux applies the policy, wait for the normal hourly background scan and
read-only verify that `validate-replica-floor` has zero failures and current
`skip` results. The runbook records the exact query.

Fixes #2573

### DIFF
--- a/docs/policy-reports.md
+++ b/docs/policy-reports.md
@@ -1,0 +1,57 @@
+# Kyverno policy-report refresh
+
+Kyverno policy reports are derived state. They should describe the current
+resource/policy result, not retain historical failures. This platform relies on
+that property for the Policy Reporter compliance dashboard.
+
+## Exemptions must produce a current result
+
+For report-backed validation policies, prefer a false `preconditions` entry for
+an exemption instead of `rule.exclude` when the exempt resource may already have
+a result:
+
+- a false precondition emits `result: skip` for the same policy, rule and
+  resource;
+- `exclude` or a changed `match` emits no replacement result.
+
+The second shape can leave an older result visible even after hourly background
+scans. Issue #2573 reproduced this with `validate-replica-floor`: eleven June
+failures remained while unrelated results in the same reports kept refreshing.
+The policy's committed Kyverno test pins the supported exemption shape:
+
+```bash
+kyverno test tests/validate-replica-floor --require-tests
+```
+
+The test requires a normal singleton to fail, two replicas to pass, and every
+supported exemption form (workload label, pod-template label, namespace, exact
+name and wildcard primary name) to emit `skip`.
+
+## Safe rollout and verification
+
+1. Express a new replica-floor exemption as a precondition in
+   `validate-replica-floor.yaml`. Do not add a new `exclude` entry.
+2. Run the Kyverno test above and the normal local/prod static validation.
+3. Let Flux deploy the policy change. Do not patch or delete PolicyReports and
+   do not restart the reports controller.
+4. Wait for the reports controller's normal background scan (configured for one
+   hour), then verify the target rule has no failures:
+
+   ```bash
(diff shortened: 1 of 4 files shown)


devantler:

Resolved the CodeRabbit Outside-diff finding in de2f9b3f: the top-level exemption summary now names crossplane-system and explains the changing ProviderRevision-hash rationale, matching the actual namespace precondition and adjacent comment.

Focused validation: kyverno test tests/validate-replica-floor --require-tests (7/7) and git diff --check.

@coderabbitai review

devantler:

Resolution record for the ⚠️ outside-diff-range finding (header exemption summary omitting `crossplane-system`): already addressed on the current branch HEAD — the "Exemption categories" comment now ends the namespace bullet with "plus crossplane-system (external-state single-writer controllers whose provider Deployment names carry changing ProviderRevision hashes)", in sync with the precondition list. No further change needed.

Issue #2318, opened 2026-06-28
[feature]: Make Kubernetes API-server audit logs searchable in Coroot


devantler:
> 🤖 Generated by the Daily AI Assistant

## Summary

Make Kubernetes **API-server audit logs** searchable in **Coroot**, alongside the container logs and traces Coroot already ingests.

This is the one genuine residual of #1507 (the kube-prometheus-stack / Loki / Alloy log-aggregation epic), which is now **superseded by #1742** ("replace kube-prometheus-stack with self-hosted Coroot"). Coroot already delivers the bulk of #1507's goals — container-log search across all namespaces, metric/log correlation, and a single UI — so this issue carves out only the part Coroot does **not** cover today.

## Problem

- Since #1742, Coroot is the single observability stack. Its eBPF **node-agent** collects per-pod **container** logs (stdout/stderr) and traces, stored in ClickHouse (`logsTTL`, see `k8s/bases/infrastructure/coroot/coroot.yaml`). `docs/dr/alerting.md` documents Coroot as the replacement for "the log store and the log shipper".
- Kubernetes **API-server audit logs** are a *different* stream. Per `talos/cluster/audit-logging.yaml` (added in #1506), the API server writes audit events to a host **file** — `/var/log/audit/kube/audit.log` on the **ephemeral** partition of each control-plane node (rotation: 30 days, 3 backups, 100 MB/file).
- Coroot's eBPF agent captures container stdout/stderr and L7 traffic — it does **not** tail arbitrary host log files, so audit events never reach Coroot. They remain accessible only via `talosctl` on individual nodes, with **no central search, no correlation, and no retention beyond the node's ephemeral disk**.

## Proposed direction (maintainer decides)

1. **Accept file-only audit access (won't-do).** Audit logs persist on-node for ~30 days and are retrievable via `talosctl` for forensics. If on-demand per-node retrieval is sufficient, close this — no component added.
2. **Forward the audit file into Coroot via OTLP (recommended if centralization is wanted).** A small log-forwarder DaemonSet on control-plane nodes (e.g. OpenTelemetry Collector / Vector / Fluent Bit) `hostPath`-mounts `/var/log/audit/kube/` read-only, parses the JSON audit lines, and ships them to Coroot's OTLP logs endpoint — making audit events searchable alongside everything else. Coroot CE supports OTLP log ingestion.
[...]

PR #2593, opened 2026-07-12
fix(talos): mount the audit-log path so audit events reach the host

devantler:

## Why

The kube-apiserver audit log never reaches the host: the audit-logging Talos patch configures the log path but mounts no volume for it, so the log stays inside the API server's container filesystem. That makes any host-side consumer impossible — it is exactly what kicked the promoted audit-log forwarder (#2582) out of the merge queue tonight (its DaemonSet found no /var/log/audit/kube on any control-plane node).

## What

Adds the missing writable extraVolume for /var/log/audit/kube to the audit-logging patch, following Sidero's documented audit-log pattern. Applies via the normal cluster-update step (API server static pods re-render; no reboot).

**Merge-order gate:** land and deploy this before re-queuing #2582 — the deploy reconciles workloads before it syncs Talos config, so #2582 alone fails again.

Part of #2318

### DIFF
--- a/talos/cluster/enable-audit-logging.yaml
+++ b/talos/cluster/enable-audit-logging.yaml
@@ -86,3 +86,12 @@ cluster:
       - hostPath: /var/etc/kubernetes/audit
         mountPath: /etc/kubernetes/audit
         readonly: true
+      # Write-through for the audit log itself: without this volume the API
+      # server writes audit-log-path inside its own container filesystem and
+      # the log never reaches the host's EPHEMERAL partition — host-side
+      # consumers (the observability audit-log forwarder DaemonSet) then find
+      # no /var/log/audit/kube directory at all. Matches Sidero's documented
+      # audit-log pattern (writable hostPath mounted at the same path).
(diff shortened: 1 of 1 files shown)


devantler:

@codex review

devantler:

**Closing — codex's P1 is confirmed and this PR's premise was wrong.** Live verification on the prod control planes shows Talos v1.13.5 already renders kube-apiserver with a built-in writable `audit` hostPath at `/var/log/audit/kube` (present in the running static-pod spec since 06-30), and `talosctl ls` shows the audit log actively written on the host. This extraVolume would have duplicated that mountPath — exactly the risk codex flagged.

The real eviction cause was the **Talos kubelet's mount namespace** (it maps only `/var/lib/{kubelet,containerd}` + `/var/log/{pods,containers}`, so the forwarder's strict `type: Directory` check fails against a path kubelet cannot see, while the CRI mounts it fine from the host namespace). Fixed directly in #2582 (7e54e7e8) — no Talos change needed, no merge-order gate. Full evidence on #2582.

PR #2597, opened 2026-07-12
feat(security): extend user namespaces to crossview

devantler:

## Why
Continues the k8s-1.36-readiness user-namespaces rollout (#1807): each proven-stateless app gets `hostUsers: false` so container UIDs stop mapping to real host UIDs — a meaningful container-escape hardening.

## What
Opts the crossview namespace into the existing label-gated Kyverno rule (hetzner overlay only). The per-app volume audit confirmed crossview is the only remaining stateless candidate: chart persistence is disabled, so all writes are emptyDir. headlamp and actual-budget carry PVCs (stateful — deferred per the lane's stateful-last rule) and fleetdm is disabled.

Part of #1807.


devantler:

@codex review

Issue #2613, opened 2026-07-13
Single source of truth for the ghcr pull credential (OpenBao + ExternalSecret)


devantler:
> 🤖 Generated by the Daily AI Engineer

**Problem:** The ghcr pull credential currently has three independent sources of truth: the org-level `GHCR_TOKEN` Actions secret (materialized into `flux-system/ksail-registry-credentials` by `ksail cluster update` on every deploy), OpenBao `infrastructure/ghcr/auth` (materialized into tenant `ghcr-auth` secrets by External Secrets), and the Talos `machine.registries` auth (env-expanded at cluster create). A rotation therefore requires updating several places in the right order, and missing one leaves a stale credential live — fragile and easy to get wrong.

**Proposed direction:** Make OpenBao `infrastructure/ghcr/auth` the single source of truth for the *pull* credential: add a `flux-system` ExternalSecret deriving `ksail-registry-credentials` from it, and stop the deploy path from upserting that secret once devantler-tech/ksail#6083 (pull/push token split) lands. The org `GHCR_TOKEN` secret then has exactly one job (the deploy's OCI push). Document the two-write rotation runbook (OpenBao pull + org push) in the platform docs.

**Rough size:** M — one ExternalSecret manifest + ksail config change (gated on ksail#6083) + runbook docs. Sequencing: ES manifest can land first; until #6083, the deploy overwriting the ES-managed secret is a known benign conflict.

devantler:
> 🤖 Generated by the Daily AI Engineer

Live evidence for this consolidation (2026-07-13): the fragmented copies now actively fight each other. A Git-declared seed (bootstrap SOPS secret → PushSecret → secret store → ExternalSecrets) re-asserted a rotated-out registry credential over a manual store-side fix within one refresh interval — re-freezing tenant OCI delivery and image-verify admission until the Git source itself is updated. A single authoritative source (this issue) removes that regression-loop class entirely. Suggest treating this as priority evidence when picking it up.

devantler:
🤖 Generated by Codex Daily AI Engineer (scheduled automation).

Confirmed production impact and a concrete preflight opportunity from merge-group run https://github.com/devantler-tech/platform/actions/runs/29236522365:

- Kyverno `imagevalidatingpolicy/verify-ksail-images` rejected every new KSail Operator 7.168 Pod before admission because GHCR token exchange returned `DENIED` for `devantler-tech/ksail:pull`.
[...]

PR #2634, opened 2026-07-14
fix(alerting): page on failing HelmReleases; fix GHCR node-credential rotation docs

devantler:

## Why

The `ksail-operator` sat **four releases behind on prod for over a day** and nobody knew. Its image pulls were failing because the nodes' ghcr.io credential had been rotated, but a running node can only pick up a new registry credential by **rebooting** — so every node kept using the revoked one, and every ghcr.io pull failed.

Two things made this a silent, day-long outage rather than a five-minute fix:

1. **Nothing alerted.** A failed Helm upgrade rolls back and then looks perfectly healthy, so the health check we relied on to catch it can never fire for this class of failure.
2. **Our written remediation was wrong.** The note in the repo told an operator to patch the node config and restart the stuck pods. Neither does anything — following it would not have fixed this.

## What

- **Alerts now page on failing HelmReleases directly** instead of assuming a broken release drags its parent into an unhealthy state — a successful rollback deliberately hides exactly that.
- **Adds a real rotation procedure** for the node registry credential to the DR runbook, with the step that actually matters (a rolling node reboot) and a way to check whether nodes are currently stale.
- **Corrects the misleading note** that would have sent the next operator down a dead end. Comments only — the effective config is byte-identical.

Prod is **already recovered** (operator back on the current release, no failing pulls cluster-wide). This PR makes the failure *visible* next time and the fix *repeatable*.

Context: the credential having several sources of truth is tracked in #2613.

### DIFF
--- a/docs/dr/runbook.md
+++ b/docs/dr/runbook.md
@@ -487,6 +487,86 @@ gh secret set TALOS_CONFIG --env prod --repo devantler-tech/platform < ~/.talos/
 
 ---
 
+## Scenario 10 — GHCR pull-credential rotation
+
+The ghcr.io pull credential lives in **two places that behave very differently**,
+and the node one is the trap:
+
+| Copy | Consumer | Picks up a new token… |
+|---|---|---|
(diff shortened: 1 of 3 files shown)


devantler:

Closing — I authored this against a **stale submodule pin** (`5bb210a9`, 8 commits behind), so its diff would have reverted #2625's `RegistryAuthConfig` migration. My mistake; re-basing on current `main`.

The findings still stand (and #2625 has a hole in it) — reopening as a fresh PR against `main` shortly.

PR #2636, opened 2026-07-14
fix(backstage): give the startup probe a real budget so it survives a restart

devantler:

## Why

Backstage could not survive being restarted. It was only ever healthy because nothing had restarted it — the moment a node roll forced it to start cold, it went into a permanent crash loop and could not recover on its own.

The cause is a boot deadline that is far too short: it is given **30 seconds** to start, and Backstage genuinely needs about a minute (it initialises a dozen plugins and migrates its database first). It was being killed *just* as it finished starting — and every retry started from scratch, so it could never win.

This was a live landmine, not a hypothetical: a routine node reboot today put it into a crash loop at 17 restarts. Any Talos upgrade would have done the same.

## What

Gives Backstage a realistic 5 minutes to boot. Failure detection for an already-running pod is unchanged — this only buys start-up time.

Verified on prod: with the longer budget it came up healthy in about a minute, and it is running now.

### DIFF
--- a/k8s/bases/apps/backstage/helm-release.yaml
+++ b/k8s/bases/apps/backstage/helm-release.yaml
@@ -197,6 +197,37 @@ spec:
               name: backstage-backend-secret
               key: password
 
+      # The chart's startupProbe sets no failureThreshold, so Kubernetes applies
+      # its defaults (3 x 10s) — a 30-SECOND budget to boot. Backstage does not
+      # boot in 30s: it initializes a dozen backend plugins and runs the catalog
+      # DB migration first, which measured ~45-60s on a cold start here. The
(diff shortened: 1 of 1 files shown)


devantler:

All checks are green. Routing review to the Codex lane (CodeRabbit is rate-limited org-wide — `devantler-tech/.github#101`).

@codex review

devantler:

Context for promotion: Backstage is now slated for **decommission** in favour of Port — #2637.

This PR still lands. It stops a live crash loop today (Backstage cannot survive a cold start; a node roll left it at 17 restarts, 0/1), the migration is weeks out, and it costs nothing once Backstage is removed. Not blocked on #2637.

PR #2691, opened 2026-07-18
feat(policy): reject explicit hostUsers opt-out in userns namespaces

### LINKED ISSUE #2690: Guard the user-namespace opt-in against silent regression (enforce ratchet)

## Evidence

The user-namespace rollout under #1807 is delivered as a **mutate-only** control. `add-security-context.yaml`'s `add-user-namespaces` rule injects `hostUsers: false` into pods in namespaces carrying `pod-security.devantler.tech/user-namespaces: enabled`, using a **conditional anchor**:

```yaml
mutate:
  patchStrategicMerge:
    spec:
      +(hostUsers): false
```

`+(key)` means *set only if absent*. A workload that explicitly sets `hostUsers: true` therefore **keeps** it — silently, with no admission error, no policy report, and no signal anywhere. The namespace still carries the opt-in label and still looks hardened.

A repo-wide search confirms `hostUsers` appears in **exactly two places**, both inside that one mutate rule — there is **no validate rule anywhere** that constrains it:

```
k8s/bases/infrastructure/cluster-policies/best-practices/add-security-context.yaml:10   (comment)
k8s/bases/infrastructure/cluster-policies/best-practices/add-security-context.yaml:93   (+(hostUsers): false)
```

This is the "**enforce ratchet**" already named as the remaining step in the 2026-07-12 per-app audit on #1807 ("Remaining after #2597: the stateful set … then the enforce ratchet").

Live production state (read-only, 2026-07-18) confirms the control is currently *effective* but *unguarded* — 5 namespaces opted in, every non-CNPG pod unshared, every CNPG-labelled pod correctly left on host userns by the rule's exclude:

| namespace | non-CNPG pods `hostUsers=false` | CNPG pods (excluded, host userns) |
|---|---|---|
| backstage | 1/1 | 2 |
| crossview | 2/2 | 0 |
| homepage | 2/2 | 0 |
| umami | 4/4 | 3 |
| whoami | 1/1 | 0 |

**No pod in production currently sets `hostUsers: true`**, so the ratchet can be applied at `Enforce` without blocking any existing workload.

## Impact

Defence-in-depth that silently degrades is worse than none — it reports as present while being absent. Today a chart upgrade, a values change, or a new workload can opt itself out of user namespaces in a namespace explicitly designated hardened, and nothing surfaces it. This is the same regression class the platform already guards for `hostNetwork` / `hostPID` / `hostIPC` in `validate-host-restrictions.yaml`; `hostUsers` is the one host-namespace dimension left unguarded.

## Expected behaviour

In a namespace opted into user namespaces, a pod that explicitly sets `hostUsers: true` is **rejected at admission**, with the same idiom the platform already uses for the other host namespaces (`=(hostUsers): false` — validate-if-present, so an absent value still passes and is filled in by the existing mutate rule).

CNPG-managed pods keep the existing `cnpg.io/cluster` exclude, mirroring the mutate rule exactly, so the validate rule can never block a workload the mutate rule deliberately skipped.

## Acceptance criteria

- [ ] A `disallow-host-users` rule is added to `validate-host-restrictions.yaml`, scoped by the `pod-security.devantler.tech/user-namespaces: enabled` namespace selector and excluding `cnpg.io/cluster` pods.
- [ ] The rule uses `=(hostUsers): false` so an absent value passes and an explicit `true` fails.
- [ ] A `kyverno test` fixture under `tests/` covers **fail** (explicit `hostUsers: true` in an opted-in namespace), **pass** (`hostUsers: false`, and absent), and **skip** (non-opted-in namespace; CNPG-labelled pod) — RED/GREEN proven.
- [ ] Verified against live production state that no existing workload would be blocked.

Part of #1807.

devantler:

## Why

The platform's user-namespace hardening can be switched off silently. It is applied by a mutate rule that only fills in the setting when a workload hasn't specified one — so any workload that explicitly asks for the old, less-isolated behaviour simply keeps it. No error, no report, nothing surfaced. The namespace still carries its "hardened" label and still looks correct.

Defence-in-depth that can degrade invisibly is worse than none, because it reports as present while being absent.

## What

Adds an admission rule that rejects a workload explicitly opting out of user namespaces in a namespace that has graduated through a user-namespace pilot — the same guard the platform already applies to the other host-isolation settings. Databases keep their existing exemption, and workloads that say nothing are unaffected.

Verified against live production: all five opted-in namespaces are already compliant, so nothing running today is blocked by this.

Fixes #2690
Part of #1807

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,7 +57,7 @@ jobs:
               - 'go.mod'
               - 'go.sum'
               - 'scripts/tests/test-cilium-bandwidth-manager-component.sh'
-              - 'tests/cilium-bandwidth-manager-bbr/**'
+              - 'tests/**'
               - 'scripts/refresh-flux-ghcr-auth.sh'
               - 'scripts/ghcr-auth-lib.sh'
               - 'scripts/run-ksail-prod-with-pull-auth.sh'
@@ -130,6 +130,23 @@ jobs:
         # second, so it goes first — before the ksail download. See
         # scripts/validate-embedded-json.py (#2480).
         run: python3 scripts/validate-embedded-json.py
+
+      - name: ⚖️ Install Kyverno CLI
+        if: needs.changes.outputs.k8s == 'true'
+        # Pinned to the version the cluster actually runs (the kyverno Helm
+        # release is v1.18.2), so a policy that passes here is evaluated by the
+        # same engine that admits it in production. Bump both together.
+        uses: kyverno/action-install-cli@fcee92fca5c883169ef9927acf543e0b5fc58289 # v0.2.0
+        with:
+          release: v1.18.2
+
+      - name: ⚖️ Validate cluster policies
+        if: needs.changes.outputs.k8s == 'true'
+        # Runs the kyverno test fixtures under tests/. Until now nothing
+        # executed them, so tests/validate-replica-floor had never actually run
+        # in CI — a policy regression could ship green. No cluster or secrets
+        # required; the CLI evaluates the committed ClusterPolicy files offline.
+        run: kyverno test ./tests
 
       - name: 🚦 Validate default-off Cilium bandwidth manager
         if: needs.changes.outputs.k8s == 'true'
--- a/k8s/bases/infrastructure/cluster-policies/best-practices/validate-host-restrictions.yaml
+++ b/k8s/bases/infrastructure/cluster-policies/best-practices/validate-host-restrictions.yaml
@@ -1,5 +1,6 @@
 # Validates that pods in non-infrastructure namespaces do not use
-# host-level access (hostPath, hostNetwork, hostPID, hostIPC).
+# host-level access (hostPath, hostNetwork, hostPID, hostIPC), and that pods
+# in user-namespace-enabled namespaces do not opt back out (hostUsers).
 #
 # Addresses kubescape controls: C-0045, C-0048, C-0041, C-0038
 #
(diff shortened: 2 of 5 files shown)


devantler:

## Validation record

**Engine parity.** Local CLI and the pinned CI CLI are both **v1.18.2**, matching the `kyverno` Helm release running in production — so every result below comes from the same engine that admits pods on the cluster.

### RED/GREEN — the fixture has teeth

The first version of this fixture **passed vacuously and I nearly shipped it.** `kyverno test` reported `5 tests passed`, but every row's REASON was `Excluded`: the CLI does not read namespace labels from `Namespace` objects in `resources.yaml`, so a `namespaceSelector`-matched rule never applies to anything — *including the fail case*. It would have passed with the rule deleted. `values.yaml` fixes it; rows 1–3 now report `Ok`, rows 4–5 `Excluded` by design.

Two negative controls prove the assertions can fail:

| control | result |
|---|---|
| invert the pattern (`=(hostUsers): false` → `true`) | **2 failed**, exit 1 |
| remove the `cnpg.io/cluster` exclude | cnpg pod: *"Want skip, got fail"* — **1 failed**, exit 1 |
| restored | 5 passed, exit 0 |

A third control confirms the CI invocation **fails closed**: `kyverno test ./tests` returns **exit 1** on a broken policy and **0** when restored — worth checking explicitly, since `kyverno apply` exits 0 on failures by default.

### CI wiring

`kyverno test ./tests` runs all fixtures: **12 passed** (5 new + 7 pre-existing `validate-replica-floor`), with the non-Kyverno `cilium-bandwidth-manager-bbr` directory correctly ignored. This is the first time `validate-replica-floor` has ever executed in CI — it passes, so wiring it in breaks nothing.

`actionlint` clean. `kustomize build` renders 22 ClusterPolicies including the new rule. cspell flags `userns`/`devantler` in the new files, but the **pre-existing `add-security-context.yaml` flags 24 issues under the same config** — these paths aren't gated, so this matches the repo baseline rather than being a new regression.

### Tried and evaluated as a user — against live production

Exported all **15 running pods** from the five opted-in namespaces and ran the policy against their real specs:

- **unmodified → `pass: 40, fail: 0`** — no workload currently running in production is blocked by this rule, so it is safe at `Enforce`.
- **10 pods tampered to `hostUsers: true` → `fail: 10`**, exactly the tampered ones, each with the real admission message. A clean 1:1 correspondence — the guard catches the exact regression it exists for, on real workloads.

### Scope note

The `cnpg.io/cluster` exclude is deliberately redundant for today's CNPG pods (they leave `hostUsers` absent, which `=()` already passes). It mirrors the mutate rule so the validate rule can never block a workload the mutate rule deliberately skipped — including if CNPG later sets the field explicitly.

devantler:

@codex review for admission-policy correctness: the `=(hostUsers): false` validate-if-present semantics, whether the `cnpg.io/cluster` exclude and `namespaceSelector` scope match the mutate rule in `add-security-context.yaml` exactly, and whether adding an Enforce rule to an existing Enforce policy can block any workload not covered by the live-pod evaluation recorded above.

PR #2704, opened 2026-07-18
fix(aws): constrain eks-ci KMS grant creation

devantler:

### Motivation

- Prevent an escalation path where the trusted `eks-ci` GitHub OIDC principal could create/pass bounded roles that effectively enabled account-wide `kms:CreateGrant` use and thereby delegate KMS key access beyond the intended smoke-cluster scope.
- Preserve the intended smoke-cluster functionality (service-managed grants) while removing the ability for a passed/bounded role to obtain arbitrary KMS grants on all keys.

### Description

- Removed the unconditioned `kms:CreateGrant` entry from the `SmokeComputePlaneCap` allowlist in `k8s/providers/hetzner/apps/aws/policy-eks-ci-smoke-boundary.yaml` so the broad compute-plane statement no longer permits unconditioned KMS grant creation.
- Added a new `KmsGrantsForAwsResourcesOnly` statement that permits `kms:CreateGrant` only when `kms:GrantIsForAWSResource` is `true`, preserving service-managed smoke-resource grant flows while blocking arbitrary delegated grants.
- Updated the policy file comments to document the reason for the KMS condition and the intended scope of the boundary.

### Testing

- Ran `python3 scripts/validate-embedded-json.py` and it succeeded for the embedded JSON IAM policy blobs.
- Ran `python3 scripts/validate-naming.py` and it succeeded to confirm naming conventions were satisfied.
- Ran `git diff --check` and the diff reported no whitespace/format issues.
- Attempted `kubectl kustomize k8s/clusters/local/` and `kubectl kustomize k8s/clusters/prod/` but they could not be executed in this environment because `kubectl` is not installed, so full kustomize render could not be validated here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a5655fcb86c832bb6a6054f56892dc9)


devantler:

devantler:
@cursor review

devantler:

## Readiness record — promoting this after 18 days as a draft

Taken over under the maintainer's direction of 2026-08-08 that every PR in the portfolio is mine to
drive to a terminal state, regardless of which lane opened it. This is a `codex/*` draft, last touched
2026-07-21, with no push, comment or review activity since — so no one is mid-flight on it.

**Programmatically tested** — pentad clear at `210e43aa90`: 0 failing checks, 0 unresolved threads, 0
non-thread review findings.

**Reviewed** — Cursor Bugbot check-run at this exact head, `conclusion: success` with
`output.title: "Bugbot Review"` (the genuine green, not the `Error` shape that means the run never
started).

**Tried and evaluated as a user** — read the live cluster rather than reasoning from the diff, and the
result is the reason this should not have waited:

- The Crossplane MR `policy.iam.aws.m.upbound.io/eks-ci-smoke-boundary` in `aws` is the enacting path,
  and it is healthy — `Synced=True`, `reason: ReconcileSuccess`, managing the real policy for 25 days.
  So this change reconciles rather than sitting inert.
- **The live policy document today contains an unconditioned `kms:CreateGrant`, and no
  `GrantIsForAWSResource` condition.** The escalation path described in the PR's motivation is not
  hypothetical — it is the current state of production, and has been for the 18 days this draft sat.

The change itself is the standard AWS shape for this: `kms:CreateGrant` cannot carry a condition while
it shares a statement with unconditioned actions, so it moves into its own statement gated on
`kms:GrantIsForAWSResource=true`. That admits grants AWS services create on the principal's behalf
(EBS/EKS volume encryption) while removing the ability for a passed role to delegate arbitrary account
KMS key use.

**Security floor gained:** a bounded `eks-ci` role can no longer mint KMS grants for non-service
principals. **Everyday path cost:** none — service-managed grants, the only kind the smoke path uses,
still succeed. If a future smoke genuinely needs an unconditioned grant it fails explicitly at the
boundary, which is this file's stated design.

Queueing rather than merging directly: `platform` gates `main` behind a merge queue. Checked first that
this PR has no prior `merge_group` eviction and no do-not-requeue note.

devantler:

## Why this promoted PR is not merging — it is blocked, not stale

I picked this up as rung-1 work: non-draft, trusted author, checks reporting green, no unresolved
threads, and a Bugbot green at `210e43aa90`. `gh pr merge` exited 0, printed nothing, and neither
merged nor enqueued it.

`mergeStateStatus` is `BLOCKED`, and that is **real** rather than the usual stale reading. This head
is missing twelve required checks — they have never run here, so the rollup cannot report them and
reads `SUCCESS` over their absence:

```
🏗️ Build · 🧪 Test · 📦 Tidy · 📊 Code Coverage · 🔍 Dead Code Analysis
🛡️ Vulnerability Scan · 🧹 Lint - golangci-lint · 🧹 Lint - mega-linter
🖋️ Validate Publication Contract · 🔐 Validate EKS Authorization
kubescape · disarm-untrusted-update
```

A PR merged an hour ago carries all twenty-seven; this one carries sixteen. The branch was opened
2026-07-18 and workflows only run on push, so a head that predates a workflow never acquires its
check. Four active rulesets require exactly these, which is what holds the merge.

**The remedy is to update the branch onto current `main`** so the required workflows run at a fresh
head. I have deliberately not done that: this branch is in the `codex/*` namespace, and an
update-branch lands a commit on it, which is the owning lane's to make rather than mine.

This is not a review finding and nothing here is wrong with the change itself. Flagging it because
the condition is invisible to the hourly survey — by every measure that survey uses this PR is clean,
which is why it has sat promoted for 21 days. That measurement gap is filed separately as
monorepo#2730.

devantler:

## Why this sat two days looking mergeable

Auto-merge has been armed since 2026-08-08T15:35:40Z, every check reads green, `mergeable` is
`MERGEABLE`, `reviewDecision` is null, the required `CI - Required Checks` context is `SUCCESS`, and
Bugbot passed at head. The queue is empty. Yet `isInMergeQueue` was `false` and `mergeStateStatus`
`BLOCKED` throughout.

The cause is not a failing check — it is **missing** ones. Comparing the check-run *names* present at
each head:

| PR | head | check-runs |
|---|---|---:|
| #2704 | `210e43aa90` (08-08) | 20 |
| #3051 | `5d9d8f5960` (today) | 32 |
| #3047 (merged today) | `1ef879b3fd` | 32 |

The twelve absent from this head are `📦 Tidy`, `🧹 Lint - golangci-lint`, `🧪 Test`,
`📊 Code Coverage`, `🔍 Dead Code Analysis`, `🏗️ Build`, `kubescape`, `disarm-untrusted-update`,
`🔐 Validate EKS Authorization`, `🖋️ Validate Publication Contract`, `🛡️ Vulnerability Scan` and
`🧹 Lint - mega-linter`. Six of those are `validate-go-project.yaml`, which the branch ruleset lists
as a **required workflow**. A required workflow that never reported cannot be satisfied by waiting,
so the PR was unmergeable indefinitely.

I have refreshed the branch so those workflows run against a current head.

## The trap worth naming

This state is **visually identical to a stale `BLOCKED`** — the documented case where every check is
green and GitHub simply has not recomputed. A sweep that checks *check conclusions* sees all-green
and concludes stale-and-therefore-mergeable. The distinguishing signal is the **set of check names**,
not their outcomes: an all-green PR missing a required workflow's checks entirely is blocked, not
stale. Any PR whose head predates a required-workflow addition is in this state.

Note the head has moved, so **the Bugbot green at `210e43aa90` is now stale** and a fresh review is
needed at the new head before this can go in.

devantler:

## Why this is red, and the merge-order gate it is waiting on

**The failure is not in the negative controls that appear to fail.** Every subtest in
`TestValidateAuthorizationRejectsSourceAndRenderedMutations` and
`TestValidateAuthorizationRejectsBindingsThatIncludeAWSServiceAccountIdentity` reports the *same*
error — `unapproved boundary manifest fingerprint: 6e79792b08aa…` — instead of the specific message
each one asserts. That is `validateBoundary` (`scripts/validate-eks-ci-role-policy/main.go:512`)
short-circuiting **before** the check under test ever runs. One stale constant is presenting as
~12 independent test failures.

**Root cause.** This PR edits `policy-eks-ci-smoke-boundary.yaml`, so the manifest no longer matches
the pinned `expectedBoundarySHA`. That is the guard behaving exactly as designed: a boundary change
must be *deliberately re-approved*, not silently absorbed.

**The content change is a narrowing, and worth re-approving.** `kms:CreateGrant` moves out of the
unconditional action list into its own statement conditioned on
`kms:GrantIsForAWSResource=true` — a previously unconditional grant becomes service-managed-only.
Direction is correct; it reduces the ceiling.

**Three pins move together, not one.** Re-approval needs `expectedBoundarySHA` (raw file),
`expectedBoundaryJSONSHA` (canonical policy JSON, checked at :533), and the per-resource entry for
`eks-ci-smoke-boundary` (:343).

**Merge-order gate — this is why it is parked rather than fixed now.** #3053 is *already in the merge
queue* and edits the same validator, changing the aggregate `expectedRenderedSurfaceSHA` (the
crossplane-sync-exporter activation takes the rendered surface 67 → 70 grant-bearing documents). This
PR's boundary change moves that **same aggregate**. Whichever lands second must recompute it, so
computing the four values now would pin them against a `main` that is about to change underneath.

**Next step, in order:** let #3053 land → rebase this branch onto `main` → recompute all four pins from
the rendered output (never pasted from a CI log) → re-request review at the new head.

*Check-name sets were also compared against a CLEAN sibling PR and are now identical (32/32), so the
missing-required-workflow-checks condition diagnosed earlier is resolved — this failure is the real
and only remaining blocker.*

devantler:

## Why `🔐 Validate EKS Authorization` was red, and what `24dd8fad` fixes

Picked this up as an unowned failing PR — last activity was ~14 h ago, well outside the active-work
window. The branch is a sibling lane's, so this is a repair push, not parallel work on it.

**The failure was one cause wearing many hats.** Every subtest reported
`unapproved boundary manifest fingerprint: 6e79792b…` rather than the error it was actually asserting,
because `validateAuthorization` pins the boundary manifest *first* and returns on mismatch — so the
whole suite bailed before reaching the logic under test. The PR changed the manifest and left the
pins on the old bytes. That is the guard working as designed: a privilege-surface edit is meant to
force a re-approval.

**So the re-approval needed an actual review of the grant change, not just new hashes.** It is
strictly a **tightening**: `kms:CreateGrant` moves out of the unconditioned action list into its own
statement gated on `kms:GrantIsForAWSResource=true`. Every grant the new form permits, the old one
already permitted.

I measured the blast radius rather than assuming it — re-rendered all five authorization roots with
`main`'s manifest restored, then with the branch's:

```
22933 -> 22943 lines
```

and the **only** moved lines in that whole surface are `kms:CreateGrant` leaving the unconditioned
list and the new `KmsGrantsForAwsResourcesOnly` statement arriving. No document added, removed or
renamed; no other grant-bearing object touched.

**Two pins re-approved, both verified independently of the CI log** (a fingerprint read out of a log
is a value to re-derive, not one to paste):

| Pin | From | To | How verified |
|---|---|---|---|
| `expectedBoundarySHA` | `b96bfd8c` | `6e79792b` | manifest hashed three ways — API content, local file, and `main`'s copy still matching the old pin |
| `expectedBoundaryJSONSHA` | `e617004b` | `2c9bc1ce` | canonical-JSON re-derivation of the embedded policy |

### The two rendered pins are deliberately still unset

`expectedRenderedResourceSHAs[…eks-ci-smoke-boundary]` and `expectedRenderedSurfaceSHA` also move.
I did **not** set them from my local run: this host has **kubectl v1.36.1** against the validator's
approved **v1.36.2**, so a locally-derived rendered fingerprint is unapproved by this validator's own
rule — the same situation already recorded in the comment beside `expectedRenderedSurfaceSHA`
("read from the required job's own output on the approved renderer, because the local toolchain is
refused as unapproved").

So this run's CI is the source for them, and I will pin them from its output with the measured
rationale attached, exactly as the surrounding convention prescribes. Guessing them from an
unapproved renderer is the one move that would quietly defeat the pin.

devantler:

## Correction: the surface-fingerprint gap was NOT a kubectl version artifact

My earlier comment attributed the differing aggregate surface fingerprint to this host's
**kubectl v1.36.1** against the approved **v1.36.2**. That was wrong, and since it would have gone
into a security rationale as a measured fact, it is worth correcting explicitly rather than quietly.

**The real cause:** a `pull_request` CI job renders the PR's **merge commit**, while a local `go test`
renders the branch against whatever base it was cut from. Main had moved (#2986 and #3068 both landed
while this was open), so the two were rendering different trees — which shifts the *aggregate* value
while leaving every *per-resource* fingerprint identical, because the moved documents belong to main's
changes rather than this one. That asymmetry is exactly what I observed, and I misread it as a
toolchain difference.

**The proof is now direct:** after merging `origin/main` into the branch, the local run agrees with CI
exactly and the whole suite passes locally — `ok github.com/devantler-tech/platform/scripts/validate-eks-ci-role-policy`.
Same kubectl v1.36.1 as before, so the renderer was never the variable.

The corrected explanation is recorded in the comment beside `expectedRenderedSurfaceSHA`, since the
next person to move that constant will hit the same trap. The pinned values are unchanged — they were
read from the required job's output either way, which is what made the mistake harmless.

**Also in `58a30457`:** resolved the merge conflict main's two surface changes created. Both sides'
rationales are preserved intact — main's #2986/#3068 records and this change's — since those are dated
measurement records, not prose to rewrite. All four pins verified present after the merge, `gofmt`
clean, `go build` and the full suite green locally.

devantler:



Requesting review at `58a30457`. CI is green at this head (0 failing checks), zero unresolved threads, and `mergeStateStatus` is CLEAN — the current-head green review is the only outstanding gate. The head is fresh: it carries the `🔐 Validate EKS Authorization` repair, so every earlier review result is stale.

@coderabbitai review

PR #2705, opened 2026-07-18
fix(unifi): suspend Flux Kustomization for unsigned UniFi Git source

devantler:
### Motivation
- Prevent a compromised moving Git branch from using Crossplane + a write-scoped UniFi Cloud Connector credential to perform network writes by pausing automatic reconciliation until the external source is pinned or signature-verified.

### Description
- Add an inline security note and set `suspend: true` in `k8s/providers/hetzner/apps/unifi/flux-kustomization.yaml` so Flux will not automatically reconcile the unsigned `devantler-tech/unifi` source while a safe pin/signature workflow is arranged.

### Testing
- Ran `python3 scripts/validate-naming.py` (passed), a YAML parse/assertion via `ruby -e 'require "yaml"; ...'` (passed), and `git diff --check` (passed); `ksail workload validate` and `kubectl kustomize`/`kubectl apply --dry-run=client` could not be executed in this environment because `ksail` and `kubectl` were not available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a528abf5a34832bbbf53552eafe4960)

### DIFF
--- a/k8s/providers/hetzner/apps/unifi/flux-kustomization.yaml
+++ b/k8s/providers/hetzner/apps/unifi/flux-kustomization.yaml
@@ -15,6 +15,11 @@ metadata:
   annotations:
     ksail.devantler.tech/reconcile-exclude: "true"
 spec:
+  # SECURITY: this external Git source drives write-scoped UniFi Crossplane
+  # managed resources through the Cloud Connector credential. Keep automatic
+  # reconciliation paused until the source is pinned or signature-verified so a
+  # compromised moving branch cannot use Crossplane as a confused deputy.
+  suspend: true
   interval: 1m
   timeout: 5m
   retryInterval: 2m


devantler:

devantler:
@cursor review

devantler:

## The exposure is real and still live — but I am not merging this shape of fix

Picked this up under the 2026-08-08 direction that every PR in the portfolio is mine to drive.
Untouched since 2026-07-21, no one mid-flight, CI clean, and a genuine Cursor Bugbot green still
sitting at the current head (the head has not moved in 18 days, so that green never staled).

**Verified against the live cluster, and the premise holds:**

- `GitRepository/unifi` tracks `ref.branch: main` — a **moving branch** — with **`verify: NONE`**. No
  tag pin, no digest, no signature verification.
- `Kustomization/unifi` has `suspend: false` and is actively reconciling
  (`Applied revision: main@sha1:77dc6590…`).

So a compromised push to that repository's `main` would be reconciled into the cluster automatically,
against a write-scoped UniFi credential. That is exactly what this PR describes, and 18 days later it
is still true.

## Why I am still not merging it as-is

The two-sided test this repo applies to hardening changes does not pass on the DevEx side.

- **Security floor gained:** real — it closes the automatic path from an unverified moving source to a
  credentialed network write.
- **Everyday path cost:** also real, and unstated in the PR. `suspend: true` stops the UniFi app
  reconciling **entirely**. Every legitimate change to `devantler-tech/unifi` silently stops applying
  until somebody remembers to unsuspend. A control that turns delivery off is the kind people
  discover at the worst moment, and this repo's own guidance is that a change raising the floor *and*
  the friction is a draft rather than a delivery.

The PR body is honest that this is a holding action — "until the external source is pinned or
signature-verified". The durable fix is that pinning, and it is a smaller change than this one: set
`spec.verify` on the GitRepository, or pin `spec.ref` to a tag or digest. That closes the same hole
while **keeping delivery working**, which makes it strictly better on both sides of the test.

## Disposition

Parked deliberately, with the blocker named: this is superseded by the durable fix rather than wrong.
I have filed that fix so the security finding is not lost with the PR, and it enters the normal
oldest-first queue with the live evidence above attached. If the pin lands, this PR should be closed
unmerged; if the pin turns out to be blocked for a reason I have not found, this becomes the fallback
and can be merged then — with the reconciliation cost stated in its body.

devantler:

## Closing: the fix would halt a healthy product without raising the attacker's bar

I picked this up to drive to a terminal state. It is pentad-clear — 17 checks green, 0 threads, 0
non-thread findings, `CLEAN`, and a Cursor Bugbot green at head `26549e5f46`. So this is not a
quality objection; it is about whether the change is the right one.

**What merging would do.** The live `unifi` Kustomization is healthy and actively reconciling —
`Ready=True`, 43 days old, applied revision `main@sha1:fe56f32b`. `suspend: true` would stop the
UniFi network receiving its configuration from Git, silently, for as long as it stayed set.

**What it would buy.** The stated rationale is that an external, unpinned source drives write-scoped
managed resources, so a compromised moving branch could act as a confused deputy. That framing
assumes the source is a weaker trust boundary than this repository. Measured today, it is not — the
two repositories carry the **same** controls:

| Control | `platform` | `unifi` |
|---|---|---|
| Require signed commits | ✅ | ✅ |
| Require a pull request | ✅ | ✅ |
| Require linear history | ✅ | ✅ |
| Block force pushes | ✅ | ✅ |
| Require status checks | ✅ | ✅ |
| Require conversation resolution | ✅ | ✅ |

The only differences are `platform`'s merge queue and a Go workflow gate, neither of which is a trust
control. All 10 most recent commits on `unifi@main` report `verified=true`.

So an adversary able to land a malicious signed commit on `unifi@main` is equally able to land one on
`platform@main` — where they could edit this very Kustomization, its ServiceAccount, or its RBAC
directly. Suspending does not raise that bar; it only removes the capability.

**The residual is real and worth fixing properly.** Flux currently trusts that GitHub enforced the
source's rules rather than checking signatures itself. Because the source already requires signed
commits, `spec.verify` closes that gap with **no** functional regression — strictly better than
suspending. Filed as devantler-tech/platform#3124, with the outage risk to design around called out.

Closing in favour of that. Nothing here is lost: if the proper fix turns out to be infeasible, this
suspend remains available as a deliberate emergency lever rather than a default posture.

PR #2706, opened 2026-07-18
fix(talos): restrict provider image signing refs

devantler:

### Motivation

- The Talos `ImageVerificationConfig` rule for `ghcr.io/devantler-tech/provider-upjet-*` was accepting signatures from any GitHub ref (`@.+`), which permits non-release refs (branches/PRs/tags) to be trusted and therefore risks running provider images signed from unprotected refs. 
- Admission-time attestation (Kyverno `verify-app-images`) needed the same protected-ref constraint so node-level pull verification and cluster admission policy remain consistent.

### Description

- Tighten the Talos provider rule in `talos/cluster/verify-first-party-images.yaml` by changing the `subjectRegex` for `provider-upjet-*` signatures from `@.+$` to `@refs/tags/v[0-9].+` so only release-style tag refs are accepted. 
- Apply the same protected-release `subjectRegExp` change to the Kyverno attestor in `k8s/bases/infrastructure/cluster-policies/best-practices/verify-app-images.yaml` so admission validation and node pull verification match. 
- The change preserves the existing ordering and attestor semantics (first-match-wins) and is intentionally minimal to reduce risk while closing the trust surface.

### Testing

- A Python regex smoke test confirmed that `refs/tags/v0.1.0` matches the new pattern and branch/PR/non-`v*` tag refs do not. 
- `python scripts/validate-naming.py` completed successfully and `git diff --check` reported no problems. 
- `kubectl kustomize k8s/clusters/local/` and `kubectl kustomize k8s/clusters/prod/` could not be executed in this environment because `kubectl` is not installed here, so CI's static manifest validation (`ksail workload validate` / Kubescape scan) will still run on the PR.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a528fe2bb6c832b9d174ed6fe4b4be3)

### DIFF
--- a/k8s/bases/infrastructure/cluster-policies/best-practices/verify-app-images.yaml
+++ b/k8s/bases/infrastructure/cluster-policies/best-practices/verify-app-images.yaml
@@ -85,7 +85,7 @@ spec:
         keyless:
           identities:
             - issuer: https://token.actions.githubusercontent.com
-              subjectRegExp: ^https://github\.com/devantler-tech/provider-upjet-[a-z0-9-]+/\.github/workflows/publish-provider-package\.yml@.+$
+              subjectRegExp: ^https://github\.com/devantler-tech/provider-upjet-[a-z0-9-]+/\.github/workflows/publish-provider-package\.yml@refs/tags/v[0-9].+$
   validations:
     - expression: >-
         (images.containers + images.initContainers + images.ephemeralContainers)
--- a/talos/cluster/verify-first-party-images.yaml
+++ b/talos/cluster/verify-first-party-images.yaml
@@ -61,7 +61,7 @@ rules:
   - image: ghcr.io/devantler-tech/provider-upjet-*
     keyless:
       issuer: https://token.actions.githubusercontent.com
-      subjectRegex: ^https://github\.com/devantler-tech/provider-upjet-[a-z0-9-]+/\.github/workflows/publish-provider-package\.yml@.+$
+      subjectRegex: ^https://github\.com/devantler-tech/provider-upjet-[a-z0-9-]+/\.github/workflows/publish-provider-package\.yml@refs/tags/v[0-9].+$
   # First-party APP images signed by the shared publish-app.yaml workflow.
   # publish-app.yaml moved from devantler-tech/reusable-workflows into
   # devantler-tech/actions (actions#425): images published after the move


devantler:

devantler:
@cursor review

k8s/bases/infrastructure/cluster-policies/best-practices/verify-app-images.yaml
```diff
@@ -85,7 +85,7 @@ spec:
         keyless:
           identities:
             - issuer: https://token.actions.githubusercontent.com
-              subjectRegExp: ^https://github\.com/devantler-tech/provider-upjet-[a-z0-9-]+/\.github/workflows/publish-provider-package\.yml@.+$
+              subjectRegExp: ^https://github\.com/devantler-tech/provider-upjet-[a-z0-9-]+/\.github/workflows/publish-provider-package\.yml@refs/tags/v[0-9].+$
```
devantler, line 88: **[P1] Keep provider signatures bound to `refs/heads/main`**

The live `publish-provider-package.yml` is `workflow_dispatch`-only and its keyless signing job runs from the repository default branch, so the Fulcio subject ends in `@refs/heads/main`; it does not become a tag-ref identity merely because the package version input is a `v*` string. This tag-only regex would reject legitimate signed provider packages in both Kyverno and the matching Talos rule, potentially denying the provider-runtime pod and blocking node pulls. Current Platform `main` already fixes this via #2964 by pinning the provider identity to `refs/heads/main`. Rebase/drop this conflicting tag-only change and retain a positive main-ref plus negative feature-ref regression.

devantler (commented):

Static security review at exact head `216418324f9f2c9cb41fb6d47025a30962d439e9`.

devantler:

**Parked on a named blocker: this PR now collides with a newer fix on `main`, and resolving the
conflict the obvious way could break image verification.** Flagging rather than merging or closing,
because I could not settle the open question from evidence.

**The vulnerability this PR targets is already closed on `main`, but differently.** All three
versions of the same two lines differ:

| | `provider-upjet-*` subject ref |
|---|---|
| this PR's base | `@.+$` — any ref (the vulnerability) |
| **this PR** | `@refs/tags/v[0-9].+$` — release tags |
| **`main` today** | `@refs/heads/main$` — the main branch |

So `main` has independently narrowed `@.+$`, and the two narrowings are **mutually exclusive**: a
subject matching one cannot match the other. This is a design collision, not a mergeable difference —
resolving toward this branch would silently replace `main`'s current policy.

**Why that could be a regression rather than a tightening.**
`publish-provider-package.yml` in `provider-upjet-unifi` is **`workflow_dispatch`-only** — it has no
tag trigger — so its keyless OIDC subject carries whatever ref it was dispatched on, which for a
default dispatch is `refs/heads/main`. If that is how it is actually published, this PR's tag-only
regex would reject every legitimate provider image, and the neighbouring comment in
`talos/cluster/verify-first-party-images.yaml` names exactly that failure mode
(`ImagePullBackOff` from a mismatched identity).

**What I could not determine.** The workflow has no runs in the last 100 for that repository, so I
have no observed dispatch ref to confirm it either way. That is the blocker: the correct regex is
whichever ref the package is genuinely signed from, and that needs checking against a real published
signature (e.g. `cosign verify` against the current xpkg) rather than inferred.

**Suggested resolution.** If the provider is published by dispatching on `main`, `main`'s current
policy is already correct and this PR should be closed as superseded. If it is published from release
tags, `main` is the one that is wrong and the fix belongs there, not on this six-day-stale branch.
Either way the answer is one signature check, and it decides the PR.

I have deliberately not resolved the conflict — taking this branch's side is the silent-revert case
described above.

devantler:

**Blocker resolved with direct evidence — and this change would break image verification as written.**

The open question on this PR was which ref actually signs the provider package, since
`publish-provider-package.yml` is `workflow_dispatch`-only with no recent runs. A `cosign verify`
against the package the cluster is running right now settles it.

**The live package is signed from `refs/heads/main`, not from a tag.** Verifying
`ghcr.io/devantler-tech/provider-upjet-unifi:v0.1.0` (the image in the running `provider-upjet-unifi`
pod) against this PR's proposed regex fails, and cosign reports the actual SAN:

```
expected SAN value to match regex
  "…/publish-provider-package\.yml@refs/tags/v[0-9].+$",
got
  "https://github.com/devantler-tech/provider-upjet-unifi/.github/workflows/publish-provider-package.yml@refs/heads/main"
```

The control arm — the same command with `@refs/heads/main` — verifies cleanly (exit 0).

**Consequence if merged as-is:** both changed rules would stop matching the deployed package.
`verify-app-images.yaml` would reject it at admission, and `talos/cluster/verify-first-party-images.yaml`
would reject it at the *host* layer, which is the more damaging of the two — that is an image-pull
failure on the nodes rather than a rejected manifest.

**The intent is right; the ordering is wrong.** Restricting signing to release tags is a real
hardening win, but it has to follow the publisher, not lead it. The sequence is:

1. Publish `provider-upjet-unifi` from a tag ref (trigger on tag push, or dispatch against the tag)
   so the signature carries `@refs/tags/vX.Y.Z`.
2. Re-publish the package and confirm with `cosign verify` that the new SAN is tag-based.
3. *Then* narrow both regexes here.

Until step 2 is done and verified, this PR is blocked on that upstream change rather than on anything
in this repository. Leaving it open and parked on that named blocker.

**Reproduce:**

```
cosign verify --certificate-oidc-issuer https://token.actions.githubusercontent.com \
  --certificate-identity-regexp '.*' \
  ghcr.io/devantler-tech/provider-upjet-unifi:v0.1.0
```

devantler:

**Closing as superseded and unmergeable-as-written.** The blocker the previous pass named is real and confirmed, but parking a third time is the wrong terminal state — this diff cannot be finished in any form, and leaving it open is not neutral.

**The vulnerability it targeted is already closed on `main`, by a different and correct mechanism.** Verified against current `main` rather than inherited from notes — both files this PR changes now pin the provider identity to the default branch:

| | `provider-upjet-*` subject ref |
|---|---|
| this PR's base | `@.+$` — any ref (the original vulnerability) |
| **this PR** | `@refs/tags/v[0-9].+$` |
| **`main` today** | `@refs/heads/main$` (both the Kyverno and Talos rules) |

**Merging it would break image verification**, as the earlier `cosign verify` established: the deployed package's SAN is `…publish-provider-package.yml@refs/heads/main`, so the tag-only regex matches nothing that actually exists. `main`'s own comment records the failure mode from the last time this identity was wrong — `provider-upjet-unifi` stuck 0/1 on prod, stalling the UniFi migration.

**Why closing rather than parking again.** Three separate passes have now re-derived this same analysis. The branch is ~4 weeks stale, `CONFLICTING`, and based on a tree that predates both `main`'s provider-ref fix and its 40-hex pin on the `publish-app` identity — so resolving the conflict toward this branch would silently revert two current policies, which is exactly the trap flagged here on 08-15. When the publisher does move to tag refs, the platform-side change is a **fresh two-line edit from current `main`**, strictly safer to write than to untangle from here. Nothing is lost by closing.

**The work is preserved, not dropped:**
- `devantler-tech/provider-upjet-unifi#35` — publish and sign from a release tag (the prerequisite).
- `devantler-tech/platform#3162` — narrow both regexes here afterwards, linked as **blocked by** that issue, with behavioural acceptance criteria (a real `cosign verify` plus a negative control) rather than the regex smoke test used here.

The intent was right and is retained; only the ordering and this particular branch are being discarded.

PR #2709, opened 2026-07-18
fix(dex): restrict GitHub auth to maintainers

devantler:
### Motivation
- A recent change enabled Actual Budget to authenticate directly against Dex using a shared `public-client`, which allowed any member of the `devantler-tech` GitHub org to log in and potentially become the first-owner, creating an authorization risk.
- The intent is to narrow the authorization boundary so native-OIDC apps that bypass `oauth2-proxy` cannot be claimed by non-maintainer org members.

### Description
- Narrow Dex's GitHub connector by adding `teams: - maintainers` under the `orgs: - name: devantler-tech` entry in `k8s/bases/infrastructure/controllers/dex/helm-release.yaml` so only `devantler-tech/maintainers` can authenticate. 
- Preserve `oauth2-proxy`'s `allowed_groups` as defense-in-depth by clarifying its comment in `k8s/bases/infrastructure/controllers/oauth2-proxy/helm-release.yaml`. 
- Update the Crossview security note in `k8s/bases/apps/crossview/helm-release.yaml` to reflect that Dex is now maintainer-scoped for native-OIDC apps.

### Testing
- Parsed the modified YAML files with Ruby using `ruby -e 'require "yaml"; ...'` which succeeded.
- Ran the naming validator with `python scripts/validate-naming.py` which succeeded.
- Ran `git diff --check` which reported no whitespace/merge issues. 
- Did not run `kubectl kustomize` or `ksail workload validate` because `kubectl` and `ksail` are not installed in this environment (not run).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52939c9ae0832bbe5a7e16b2c9304c)


devantler:

devantler:
@cursor review

devantler:
> Requested by the 🤖 Daily AI Engineer — the previous Cursor Bugbot run returned `conclusion: neutral` with `Bugbot failed to run`, i.e. the review never happened. Re-requesting at a slower pace.

devantler:
@cursor review

k8s/bases/infrastructure/controllers/dex/helm-release.yaml
```diff
@@ -156,3 +156,11 @@ spec:
             useLoginAsID: false
             orgs:
               - name: devantler-tech
+                # Dex is the authorization boundary for apps that authenticate
+                # natively (for example Actual Budget) instead of routing
+                # through oauth2-proxy's allowed_groups gate. Keep the upstream
+                # GitHub connector narrowed to the same maintainer team so org
+                # members who are not platform maintainers cannot claim or
+                # access sensitive native-OIDC apps.
+                teams:
```
devantler, line 159: > 🤖 Generated by the Agentic Engineer

P1: Keep the shared Dex connector org-scoped. Dex treats `orgs[].teams` as a login requirement, not as a per-client claim filter (https://dexidp.io/docs/connectors/github/). This connector backs oauth2-proxy, Headlamp, Actual Budget, Vault, KSail/kubectl, and Crossview, so adding `teams: [maintainers]` rejects every legitimate org member outside that team before any client-specific policy can run. Enforce the maintainer-team rule at the sensitive native-OIDC app boundary (or give those clients a separate connector), and add a negative test that denies a non-maintainer there without blocking intended shared-client access.

devantler (commented):

Static exact-head review of `900256b826f19d776149ec8a66064a7222588723`.

k8s/bases/infrastructure/controllers/dex/helm-release.yaml
devantler: > 🤖 Generated by the Agentic Engineer

**Refuting the P1 severity on measurement — the harm population is empty. Keeping the architectural point as a note.**

The finding is right about Dex semantics: `orgs[].teams` is a login requirement, not a per-client claim filter, so narrowing the shared connector does gate every Dex-backed client at once. What it assumed, and what I did not check when I wrote it, is that there are org members this would exclude.

Measured just now against the org API:

- `devantler-tech` has **1 member**
- the `maintainers` team has **1 member**; the `admins` team has **1 member**
- the org has exactly two teams, `admins` and `maintainers`

So the set of members outside `maintainers` is **empty**, and no client — oauth2-proxy, Headlamp, Actual Budget, Vault, KSail/kubectl or Crossview — loses a single legitimate login. The change is a strict privilege reduction with zero current blast radius, which matches the rendered-surface reasoning already recorded in the validator comment.

**The concern is latent rather than absent, so it is worth keeping:** the first non-maintainer org member added later breaks *every* Dex client simultaneously, including `kubectl` access, and the failure will not obviously point back at this connector. That is a real trap, but it is a follow-up worth filing rather than a blocker on a privilege reduction that harms nobody today. Resolving on that basis.

devantler:

The latent shared-connector gate discussed in the resolved thread is now tracked as #3187, so it is not lost when this merges. Nothing blocking remains on that thread — the change excludes no current org member.

devantler:

## Conflict resolved — but the fingerprint is deliberately NOT re-approved

This PR was `DIRTY`; GitHub's own branch update refused ("Cannot update PR branch due to conflicts").
Merged `main` locally instead and pushed `629bdd10`. Of the three files this branch touches, only
`scripts/validate-eks-ci-role-policy/main.go` conflicted — the two manifest changes merged cleanly.

**Why it could not be resolved by picking a side.** Both parents appended to the same
authorization-ledger region and each described only its own delta:

| Parent | Delta it records |
|---|---|
| this branch | the Dex GitHub-connector narrowing to the maintainers team |
| `main` | #2713's `gateway-tenant-edit` route-kind narrowing **and** #2725's Umami provisioning grant |

Both deltas are present in the merged surface and neither cancels the other, so **both prose records
are retained** and only one `const` survives.

**That const is left at `main`'s approved value, and it is NOT a re-approval of this merge result.**
Neither parent's value describes the merged surface, so writing either one in would be a false
approval of a surface nobody measured. I could not produce an authoritative local measurement
either: this host runs kubectl v1.36.1 and the ledger's approved renderer is v1.36.2, so the
validator correctly refuses fail-closed —

```
EKS CI role policy: unapproved renderer: kubectl=v1.36.1 kustomize=v5.8.1
```

That refusal is the guard working, not an obstacle to route around.

**So this PR is parked on a named blocker: the merge-result fingerprint must be measured and
deliberately approved.** The required `🔐 Validate EKS Authorization` job runs the approved
toolchain and is the authoritative renderer — it will now report the real value for `629bdd10` as an
`unapproved rendered authorization surface fingerprint` mismatch. That reported value is what should
be approved into the const, with a second independent render agreeing and a short record of the
conservation check, exactly as the surrounding entries do. **Do not rubber-stamp it** — confirm the
delta is only the three known changes before approving.

Unrelated but worth noting while the render was being checked: the prod overlay renders cleanly from
here right now, so the Cloudflare CDN fragility tracked in #3196 is not currently biting.

devantler:



CI is green at `35c8a3a1` (33 checks, 0 failures) and the branch is `CLEAN` against main.

@coderabbitai review

devantler:

## Readiness at `35c8a3a1` — all three conditions proven

**1. Programmatically tested.** 33 checks, 0 failures, `mergeStateStatus: CLEAN`. This carries its own RED/GREEN proof: the required `🔐 Validate EKS Authorization` job **failed** at `1d357303` reporting `8773eaf0`, and **passes** at `35c8a3a1` with that value approved. Nothing else changed between the two heads.

**2. Reviewed.** CodeRabbit green at this exact head — auto-generated summary refreshed 21:31:04Z after the 21:25Z request, naming the full head `35c8a3a15a969d11ca9a7f8504e6981f5567c60f`, verdict "No actionable comments were generated", and its only collapsed section is the excluded `🔇 Additional comments (5)`. No rate-limit or service marker. Corroborated by the head's `CodeRabbit` status reading `Review completed`. 0 unresolved threads, 0 non-thread findings.

**3. Tried and evaluated as a user.** The consumer of this constant *is* the required job, and I watched it change behaviour across the two heads rather than tracing the code path. The substance was verified before approving, not rubber-stamped:

- The branch was levelled to `behind_by=0` **first**, so the digest describes the merge result rather than a stale rendering.
- The digest was **predicted before it was read**. `main` already contains the levelling merge's only k8s change — Checkov skip annotations on the umami CronJob — and the required job passes on `main` against the *previous* digest, so those annotations are provably not in the surface. Prediction: the value should be unchanged from the pre-merge head. CI then reported the identical `8773eaf0` at both `629bdd10` and `1d357303`.
- **Membership is unchanged**: 0 per-identity mismatches, 0 missing, 0 duplicate, 0 encrypted findings — exactly one problem, the aggregate. The 35 unresolved-substitution notes are the same count clean `main` reports, and are diagnostic rather than a control (the validator's own comment says so, and they are emitted only alongside an aggregate mismatch to explain a hash that moved).
- The committed constant was asserted byte-equal to the digest CI reported, with an empty-value guard and a negative control against the old value.

**One renderer stands behind the digest, and that is stated rather than glossed** — the approved CI toolchain. The local kubectl here is v1.36.1 against an approved v1.36.2, so `validateRendererVersion` fails closed and cannot corroborate. The two independent observations above are evidence about what did *not* move; they are not a second rendering of what did.

Promoting and enqueuing.

PR #2713, opened 2026-07-18
fix(security): close alternate tenant route-kind hostname bypass

devantler:

### Motivation

Platform #2740 now confines tenant-authored `HTTPRoute` hostnames to per-tenant allow-lists and denies unlisted tenants. One bypass remained: the tenant role still granted four other Gateway API route kinds, and an HTTPS listener accepts `GRPCRoute` by default. Those resources never pass through the HTTPRoute-only hostname policy, so a compromised tenant artifact could still claim another service's hostname on the shared wildcard Gateway.

### Description

- Narrow `gateway-tenant-edit` to `httproutes` and `referencegrants`; ordinary built-in `edit` remains excluded, as established by #2740.
- Pin every rendered platform Gateway listener to `allowedRoutes.kinds: [HTTPRoute]`, covering routes created by principals outside the tenant role as well.
- Add a rendered regression test that fails if either the tenant RBAC grant set widens or any current/future listener is left unpinned.
- Update the tenant onboarding documentation. The canonical `restrict-tenant-route-hostnames` policy and its fixtures remain the ones shipped by #2740; this PR deliberately does not duplicate them.

If another route kind becomes necessary, the safe change is to extend the hostname policy to cover it and relax both the role and listener boundary together.

### Validation

- `shellcheck scripts/tests/test-tenant-route-hostname-boundary.sh`
- `bash scripts/tests/test-tenant-route-hostname-boundary.sh` — 4 rendered listeners pinned; tenant RBAC contains only HTTPRoute plus ReferenceGrant
- `kyverno test ./tests/restrict-tenant-route-hostnames` — 14 passed, 0 failed
- `go test ./scripts/validate-eks-ci-role-policy`
- checksum-verified kubectl v1.36.2 / Kustomize v5.8.1 authorization render — exact fingerprint `79250b2f64ab9d3798a659d13714b1a07dd72f5baa5fb465beecf30bfd9ad172`
- `ksail workload validate`
- `ksail --config ksail.prod.yaml workload validate`

### User-visible evaluation

No tenant traffic shape changes: approved HTTPRoutes continue to attach to the same four rendered listeners, and the canonical policy's paved-road fixtures remain green. The removed route kinds have no Platform workload consumer; they were unchecked privilege, not a supported ingress path.

This direct security fix has no tracking issue. The earlier P1 review findings on this PR are resolved; the branch was re-integrated with exact current `main` at `6d926e42` after #2740 landed.


devantler:

devantler:
@cursor review

devantler:
> Requested by the 🤖 Daily AI Engineer — the previous Cursor Bugbot run returned `conclusion: neutral` with `Bugbot failed to run`, i.e. the review never happened. Re-requesting at a slower pace.

devantler:
@cursor review

k8s/bases/infrastructure/cluster-policies/best-practices/restrict-tenant-http-route-hostnames.yaml
```diff
@@ -0,0 +1,67 @@
...
+spec:
+  validationFailureAction: Enforce
+  # Uses admission request context, so it cannot run as a background scan.
+  background: false
+  rules:
+    - name: restrict-wedding-app-hostnames
+      match:
+        any:
+          - resources:
+              kinds:
```
devantler: > 🤖 Generated by the Agentic Engineer

[P1] Cover every route kind accepted by the shared HTTPS listener

This admission boundary matches only `HTTPRoute`, but at this exact base `gateway-tenant-edit` also grants tenant service accounts create/update/patch on `GRPCRoute`, while the shared HTTP/HTTPS Gateway accepts routes from all namespaces without restricting route kinds. A compromised tenant artifact can therefore attach a `GRPCRoute` with another service's hostname—or omit `spec.hostnames`—without this policy ever running. Restrict the listener/RBAC to `HTTPRoute`, or enforce the same per-tenant hostname boundary for `GRPCRoute`, with negative coverage proving alternate route kinds cannot claim the shared listener.

devantler (commented):

Static review of exact head `9f448d92075e02d56bd802246cf1f08a8f96242f`; one blocking finding is attached. This ownership-unverified branch was not checked out or executed.

k8s/bases/infrastructure/cluster-policies/best-practices/restrict-tenant-http-route-hostnames.yaml
devantler: > 🤖 Generated by the Agentic Engineer

Confirmed against the live prod cluster — this is a real gap, so I am leaving the thread open rather than resolving it. Evidence, so whoever picks it up does not have to re-derive it:

**RBAC does grant the other route kinds:**

```
$ kubectl get clusterrole gateway-tenant-edit -o json | jq -r '.rules[]'
apiGroups=["gateway.networking.k8s.io"]
resources=["httproutes","grpcroutes","tcproutes","tlsroutes","udproutes","referencegrants"]
verbs=["get","list","watch","create","update","patch","delete"]
```

**And the listeners do not restrict route kinds:**

```
$ kubectl -n kube-system get gateway -o json | jq -r '.items[].spec.listeners[]'
platform listener=https  protocol=HTTPS allowedRoutes={"namespaces":{"from":"All"}}
platform listener=http   protocol=HTTP  allowedRoutes={"namespaces":{"from":"All"}}
platform listener=https-ascoachingogvaner-dk    protocol=HTTPS allowedRoutes={"namespaces":{"from":"All"}}
platform listener=https-www-ascoachingogvaner-dk protocol=HTTPS allowedRoutes={"namespaces":{"from":"All"}}
```

`allowedRoutes` carries no `kinds`, and Gateway API's default for an HTTPS listener is HTTPRoute **and** GRPCRoute. So a tenant service account can attach a `GRPCRoute` carrying another service's hostname, and this policy — matching `HTTPRoute` only — never runs on it.

Note the finding also covers the `spec.hostnames`-omitted case, which is worth checking separately from the route-kind question.

On the two remedies: restricting the listeners to `kinds: [HTTPRoute]` is the tighter and smaller change if gRPC ingress is not actually wanted, and it closes `TCPRoute`/`TLSRoute`/`UDPRoute` along the way; extending the policy to `GRPCRoute` keeps gRPC available but leaves the other three kinds unaddressed unless the RBAC is narrowed too. Whichever way it goes, negative coverage proving an alternate route kind cannot claim a hostname is what makes it stick.

I did not implement this — it is a security boundary with a genuine design choice in it, and I could not validate a change here properly within this run. Flagging it as verified rather than guessing.

devantler: > 🤖 Generated by the Agentic Engineer

Fixed in `8fbb9d3c`. Taking the tighter of the two remedies you named — restrict rather than extend — because the evidence says gRPC ingress is not actually wanted here.

**Why restrict:** `GRPCRoute`/`TCPRoute`/`TLSRoute`/`UDPRoute` appear in exactly **one** file across the whole portfolio — the `gateway-tenant-edit` ClusterRole that grants them. No platform service, no tenant, and neither `platform-tenant-template` nor the two live tenants (`wedding-app`, `ascoachingogvaner`) references any of them. The role's own header comment already described its purpose as "manage **HTTPRoute / ReferenceGrant** resources", so the four extra kinds were scope the role was never documented to have.

**Both layers, since either alone is bypassable:**

1. **RBAC** — `gateway-tenant-edit` now grants `httproutes` + `referencegrants` only, closing `TCPRoute`/`TLSRoute`/`UDPRoute` along with `GRPCRoute`, as you noted extending the policy would not have.
2. **Listeners** — `allowedRoutes.kinds` pinned to `HTTPRoute` on all four listeners (the two in the base Gateway *and* the two the hetzner JSON6902 patch appends, which is the set your `kubectl` output showed unrestricted). This covers a route created by any principal, not just a tenant SA.

**On the `spec.hostnames`-omitted half:** that one was already closed — the first deny condition is `{{ spec.hostnames || `[]` }} Equals []`, which is exactly why it sits alongside the `AnyNotIn` check (`AnyNotIn` is vacuously false on an empty list). I added a fixture pinning it rather than changing the rule.

**Negative coverage**, which is what you asked for and what makes it stick:

- `tests/restrict-tenant-http-route-hostnames/` — the policy had **no** kyverno fixture at all, unlike its sibling tenant policies. 8 resources: the rogue claim, omitted hostnames, an approved+rogue mix, both paved-road cases, a cross-tenant claim proving the allow-lists are not pooled, and an out-of-scope platform route.
- `scripts/tests/test-tenant-route-hostname-boundary.sh` — asserts the RBAC grant set, that every listener in the **rendered prod overlay** pins `kinds`, and that the policy still returns 3-admit/4-deny.

I ablated all three layers rather than trusting a green run — re-granting `grpcroutes` was caught by name, unpinning one listener was caught naming only that listener, and widening the allow-list flipped the verdict to `pass: 5, fail: 2`. Worth flagging one thing that surfaced: yq's `!=` does **not** do deep array equality, so the obvious `(kinds | sort) != ["HTTPRoute"]` reported all four listeners as unpinned. It failed closed here, but the same shape fails *open* with the sense reversed, so the comparison is on a joined string with a comment saying why.

Resolving — but the design choice is yours to overturn: if gRPC ingress is wanted later, the path is to extend the hostname policy to `GRPCRoute` and relax both layers together, never one alone.

devantler:

@coderabbitai review

New head `8fbb9d3c` closes the P1 raised on this PR: the hostname policy matches `HTTPRoute`, so the tenant RBAC is narrowed to `httproutes`+`referencegrants` and every Gateway listener now pins `allowedRoutes.kinds` to `HTTPRoute`. Worth reviewing the two layers together — relaxing either alone reopens the gap.

devantler:

@codex review

CodeRabbit is rate-limited at this head (`Review rate limited`, 57-minute window, free OSS quota exhausted), so advancing to the next lane. Recording the completed outcome for the lane before this one: `cr:no-gate@8fbb9d3cd0251905fc27680db076309617c33a6b`.

Focus: this closes a P1 where the tenant hostname policy matches `HTTPRoute` while the tenant RBAC also granted `grpcroutes`/`tcproutes`/`tlsroutes`/`udproutes` and no Gateway listener pinned `allowedRoutes.kinds`. The two layers must be assessed together — relaxing either alone reopens the hostname-claiming path.

k8s/bases/infrastructure/cluster-policies/best-practices/restrict-tenant-http-route-hostnames.yaml
```diff
@@ -0,0 +1,67 @@
...
+      Requires tenant-managed HTTPRoutes to declare hostnames and confines each
+      tenant namespace to its approved public hostnames. This prevents tenant
+      OCI artifacts from attaching routes for other platform services to the
+      shared hostname-less Gateway listener.
+spec:
+  validationFailureAction: Enforce
+  # Uses admission request context, so it cannot run as a background scan.
+  background: false
+  rules:
+    - name: restrict-wedding-app-hostnames
```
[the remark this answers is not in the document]
devantler: > 🤖 Generated by the Agentic Engineer

Verified and fixed in `5cab71da`. This was a real gap, and it was worse than one missing tenant.

**Confirmed independently before acting:** `doggy-countdown` is in `k8s/bases/apps/kustomization.yaml`, its `role-binding-doggy-countdown.yaml` binds the `tenant-edit` ClusterRole (which aggregates `gateway-tenant-edit`, so it can create HTTPRoutes), and no rule in the policy matched its namespace. Its real hostname is `simba.platform.devantler.tech`, read from the tenant's own `deploy/httproute.yaml` in `devantler-tech/doggy-countdown` rather than inferred.

**Took the second remedy you offered, not the first**, because enumerating namespaces is fail-open by construction — the next tenant reopens the gap silently, exactly as this one did. Both halves shipped:

1. `restrict-doggy-countdown-hostnames`, allowing `simba.platform.devantler.tech` only.
2. `restrict-unlisted-tenant-hostnames` — a **default-deny** for any namespace labelled `app.kubernetes.io/managed-by: ksail` that is not one of the three listed tenants.

**On the scope of the default-deny**, since a too-wide selector here would break production routing: exactly four namespaces carry that label — `wedding-app`, `ascoachingogvaner`, `doggy-countdown`, `github-config`. The fourth has no HTTPRoute at all (no `http-route.yaml`, no Gateway API resources), so nothing live is denied today. The platform's own routed apps (`actual-budget`, `homepage`, `whoami`, `backstage`, `headlamp`) are **not** ksail-managed namespaces, so their platform-authored HTTPRoutes are untouched. It is the same selector `restrict-tenant-issuer-refs` and `restrict-tenant-secret-stores` already use.

Onboarding a tenant that serves HTTP now requires adding its rule; the failure is an admission denial whose message names that fix.

**Coverage:** fixtures extended to 11 resources / 15 assertions, including that the default-deny does **not** catch the three listed tenants' paved roads. Ablated it — removing the rule moves the verdict from `pass: 4, fail: 6` to `fail: 5`, and the structural test fails on the mismatch.

One correction to the finding's framing, for the record: `doggy-countdown`'s RBAC permitting `httproutes` is now the *only* route kind it permits — the same commit that this thread's original P1 covers narrowed `gateway-tenant-edit` from five route kinds to `httproutes`+`referencegrants` and pinned the listeners to `HTTPRoute`.

devantler:

**State at `5cab71da`** — both P1 findings on this PR are fixed and every review thread is resolved. One mechanical step remains before this can merge.

**Remaining blocker: the authorization-surface re-approval.** `🔐 Validate EKS Authorization` fails with `unapproved rendered authorization surface fingerprint`, and that is the guard working correctly, not a defect: this PR narrows the `gateway-tenant-edit` ClusterRole and adds two Kyverno rules, so the rendered authorization surface legitimately moved and `expectedRenderedSurfaceSHA` in `scripts/validate-eks-ci-role-policy/main.go` needs updating with a written justification of the delta.

**Why it is not done in this run, rather than an oversight:** that constant is deliberately not computable locally — the validator refuses the local toolchain as an unapproved renderer, so the fingerprint must be read from the required CI job's own output. The run for this head was still in progress when this run ended. Whoever picks it up:

1. Read the fingerprint from the `🔐 Validate EKS Authorization` job on the newest run for the **current** head.
2. Update `expectedRenderedSurfaceSHA` with a justification in the established style — the delta here is: `gateway-tenant-edit` loses `grpcroutes`/`tcproutes`/`tlsroutes`/`udproutes` (grant-narrowing only; no identity, binding, or new verb), and `restrict-tenant-http-route-hostnames` gains two rules. Nothing is added to any identity's privileges.
3. Re-request a review — every lane was exhausted at the time of writing: CodeRabbit rate-limited (free OSS quota, ~57-minute window), Codex spent on the previous head, Cursor Bugbot at its account usage limit.

⚠️ **Note for whoever also drives platform#2742:** it fails the *same* check for the same reason, so both PRs want to update the same constant. Whichever merges first invalidates the other's fingerprint, which will need re-reading rather than rebasing.

devantler:

**Authorization surface re-approved in `42a13b3e`** — this was the remaining blocker noted above, and it is now closed rather than carried forward.

**The delta, measured rather than asserted.** Rendered all five authorization overlays from `main` and from this head and diffed the canonical documents:

| | main | head |
|---|---|---|
| Role / ClusterRole / RoleBinding / ClusterRoleBinding / ServiceAccount | 10 / 22 / 15 / 10 / 14 | identical |
| ClusterPolicy | 26 | 27 |

**71 RBAC documents on each side, identity sets identical, and exactly ONE differing in content** — `ClusterRole/gateway-tenant-edit`. Nothing added, removed or renamed anywhere else.

That one change is a **strict narrowing**: `grpcroutes`, `tcproutes`, `tlsroutes`, `udproutes` removed, leaving `httproutes`+`referencegrants`. The set difference in the **addition** direction is empty, the seven verbs are byte-identical, the single apiGroup is unchanged, and both aggregation labels are unchanged — so no identity gains anything. The added ClusterPolicy is the admission boundary itself; it grants nothing and only denies.

**Both renderer controls pass, so this is verified rather than copied.** The existing narrative in `main.go` says the local toolchain is refused as unapproved, so I initially wrote that the value could only be taken from CI — **that turned out to be wrong for this surface and I corrected it before committing**. Measured: local kubectl is v1.36.1 against the pinned v1.36.2 (same embedded Kustomize v5.8.1), and `TestValidateAuthorizationAcceptsCommittedPolicy` **runs and passes locally in both trees** — `main` reproduces its prior constant `1c4c1986…`, and this tree reproduces `55533168…`, the value CI reported. A local render that could not reproduce a known-good constant would say nothing about what CI hashed; both controls passing is what makes the delta above admissible.

Ablated it too: reverting to the old constant fails reporting exactly the new fingerprint, and restoring it passes.

⚠️ **Still true for [#2742](https://github.com/devantler-tech/platform/pull/2742):** it fails the same check for its own reasons and will need its own fingerprint — and once either PR merges, the other's becomes stale and must be re-read rather than rebased.

devantler:

@coderabbitai review

New head `42a13b3e` — the PR is now pentad-clear apart from this review: `CLEAN` against base, all checks green including `🔐 Validate EKS Authorization`, zero unresolved threads.

Worth reviewing as one change, since each layer is bypassable alone: the tenant role grants `httproutes`+`referencegrants` only, every Gateway listener pins `allowedRoutes.kinds` to `HTTPRoute`, and the hostname policy now default-denies any tenant namespace without its own allow-list. The last commit re-approves the rendered authorization-surface fingerprint that the first two necessarily moved.

devantler:

CodeRabbit refused at this head too — `Review limit reached`, free OSS quota, another **57-minute** window. Recording the completed outcome so the next run resumes at the right place instead of re-spending the free lane on a head it has already refused.

**This PR is finished apart from one green review.** At `42a13b3e`: `CLEAN` against base, every check green including `🔐 Validate EKS Authorization`, zero unresolved threads, zero outstanding review findings — both P1s raised on it are fixed with ablated coverage.

**Lane state, so nobody re-derives it:**

| Lane | State |
|---|---|
| CodeRabbit | rate-limited at this head, ~57min window — **retry this one first**, it is the free lane |
| Codex | reviewed `8fbb9d3c`, raised the doggy-countdown P1, now fixed; **not yet re-requested at this head** |
| Cursor Bugbot | account **usage limit** — no retry window, only the maintainer can lift it |

Cheapest path: retry CodeRabbit once the window clears. Codex is the fallback, and it has not yet seen the default-deny rule or the fingerprint re-approval.

devantler:


@coderabbitai review

devantler:


@coderabbitai review

devantler:

@coderabbitai review

Current head `21a5c065cf67cfbfdefc2d5d68f0f19af4e2c61c` is green after merging the recovered production base. Please review the complete seven-file route-kind boundary: tenant RBAC is narrowed to `HTTPRoute` plus `ReferenceGrant`, all rendered Gateway listeners restrict allowed route kinds to `HTTPRoute`, and the focused regression plus authorization fingerprint bind both enforcement layers.

devantler:



@coderabbitai review

devantler:

**Parked on a named blocker: the authorization fingerprint needs a re-measurement this run could not complete to the protocol's standard.** Recording what I established so the next attempt is one step rather than a rediscovery.

**The conflict is two files, and only one is hard.**

`.github/workflows/ci.yaml` is trivial — the two sides add *different* test files to the same filter list (`test-tenant-route-hostname-boundary.sh` here, `test-umami-provisioning-bootstrap.sh` from `#2725`). Both entries belong; keeping both resolves it.

`scripts/validate-eks-ci-role-policy/main.go` conflicts on `expectedRenderedSurfaceSHA`, and **neither side's value describes the merge result**:

- this branch's `9a9a0102…` was measured against main `6ebcb24f`, *before* `#2725` merged (04:43Z today), so it predates the Umami grant now in main;
- main's `295c44e3…` was measured without this PR's ClusterRole reduction.

So the merge needs a fresh measurement against main `d925654e`, with both comment blocks retained — the running log reads chronologically and `#2725`'s entry is referred to by later text.

**Why I did not just write a value.** The recorded protocol requires **two independent renderers agreeing**. I obtained the pinned renderer (kubectl `v1.36.2` / Kustomize `v5.8.1`, checksum-verified against the published `dl.k8s.io` digest, since the validator rejects `v1.36.1` outright as `unapproved renderer`) and rendered the resolved merge locally. It produced a candidate — `b8235416c8452d2a123eb9279d64a4d6adc11625b9ed3be6801cd8abd1794a78` — **but simultaneously reported 35 `unresolved Flux substitution` problems** spanning HelmReleases this PR does not touch (cilium, dex, velero, longhorn and others). That means the local render is incomplete relative to the CI one, so its fingerprint is not trustworthy on its own and I will not approve a security constant on it.

**Treat that candidate as unconfirmed, not as the answer.** The next step is a render in an environment where those substitutions resolve, cross-checked against the required `🔐 Validate EKS Authorization` job — which is the second renderer the protocol names, and the repo's own source of truth.

Nothing else on this PR is blocking: 0 unresolved threads, 0 body findings, and no current-head green review (CodeRabbit answered the 04:31Z request at this head with its generic chat tip rather than a review — recorded as `cr:no-gate@03df5f16`). The review lane should be advanced only after the conflict is resolved, since that push stales any green anyway.

devantler:

This was `DIRTY` against main and could not merge. Resolved the conflict (`03df5f16` → `c790f999`); no code in the fix itself changed.

**`.github/workflows/ci.yaml`** — each side appended a different test path to the same `k8s` filter list (this branch's tenant route-hostname boundary test, main's Umami provisioning bootstrap test). Both belong, so both are kept.

**`scripts/validate-eks-ci-role-policy/main.go`** — the interesting one. Both parents had re-approved `expectedRenderedSurfaceSHA` against the *same* exact main `6ebcb24f`, but each measured only its own delta: this branch the `gateway-tenant-edit` route-kind narrowing, main the #2725 Umami provisioning grant. Neither value describes their union, so the conflict could not be resolved by picking a side.

Both records are retained, because both deltas are genuinely present in the merged surface and neither cancels the other — one is a strict privilege reduction, the other a Lease-scoped grant. Main's value is carried through **unmeasured**, so `🔐 Validate EKS Authorization` will reject it and report the digest the merged surface actually renders to; that reported value is what should replace it. Guessing it locally is not possible while Flux substitutions are unresolved.

I did not touch the security fix's own reasoning — the reduction it makes is unaffected by the merge.

Note the branch is in the `codex/*` lane. I took this on because its last push was ~12h old, so the active-work test showed nobody mid-flight, and per the portfolio PR-ownership rule a conflict is repair work rather than parallel work. Nothing else on the branch was modified.

devantler:

Completed the digest follow-up the `expectedRenderedSurfaceSHA` comment described. `90cf208e` records `b8235416c8452d2a123eb9279d64a4d6adc11625b9ed3be6801cd8abd1794a78`.

**Diagnosis.** `🔐 Validate EKS Authorization` and `🧪 Test` both failed on this head, reporting an unapproved surface fingerprint plus 35 `unresolved Flux substitution` notes. The notes are **not** a second failure — `validateAuthorization` collects them into `substitutionProblems` and emits them *only alongside* a mismatch, deliberately, because a resource carrying `${…}` is forced into the aggregate so its literal text is already covered by the digest. The code says so directly: promoting them to errors was measured on 2026-07-21 to fail the approved tree on thirty-plus HelmReleases. So there was exactly one control failure: the whole-surface digest.

**Not main.** platform#3171, also at an up-to-date head, passes both workflows, so `main` is healthy and this is confined to this PR. This head is `behind_by=0`, so it is not a stale-base case either.

**Two independent renderers agree**, satisfying the protocol:

| Renderer | Digest | Substitution notes |
|---|---|---|
| Required CI job (run `32019353067`) | `b8235416…` | 35 |
| Local, kubectl v1.36.1 / kustomize v5.8.1 | `b8235416…` | 35 |

Note the two *CI jobs* are **one** renderer, not two — both run `main_test.go:1140`. The local render is what makes the second.

**Conservation.** The aggregate is the only thing that moved: both renderers report **zero** per-identity mismatches against `expectedRenderedHashes` and **zero** missing resources, so every individually approved entry — both parents’ deltas included — is byte-identical to its recorded value. The reviewed reasoning for the `gateway-tenant-edit` route-kind narrowing is untouched.

**One correction recorded in the comment.** The pre-measurement note held that a local render could not approve this, being incomplete without the CI substitution inputs. Measured, the approved toolchain reports the *same* 35 notes and the *same* digest, so the renders are equivalent here and the local one is a genuine second renderer rather than a degraded copy.

Package green locally. CI re-runs at `90cf208e`.

devantler:


CI is now fully green at this head (`🔐 Validate EKS Authorization` included) and there is no conflict with base. This PR is gated only on a current-head review.

@coderabbitai review

devantler:

Un-conflicted at `8d607154`. #3171 merged into `main` 30 minutes after this PR's last head, and both sides had independently re-approved `expectedRenderedSurfaceSHA` — this branch for the `gateway-tenant-edit` route-kind narrowing, `main` for the crossview reload annotation. Neither parent's value describes the merge result, so the conflict could not be resolved by picking a side; it needed a genuine re-measurement. Both parents' prose records are retained, because both deltas are present in the merged surface.

**Measured value:** `b5b394181a8f2bc325bd427a90990fbc3872fb1f41f3d938f1e28f0aac1075f4`

**Two independent renderers agree, and each was first proved against clean `main` `b9af3892` as a matched control:**

| Renderer | Control on clean main | Merge result |
|---|---|---|
| Approved toolchain — checksum-verified kubectl v1.36.2 / Kustomize v5.8.1, running the validator | `EKS CI role authorization contract passed.` | `b5b39418…` |
| Local render — kubectl v1.36.1 / Kustomize v5.8.1, via `TestValidateAuthorizationAcceptsCommittedPolicy` | test `ok` | `b5b39418…` |

The matched controls are what make this a measurement rather than a guess: both renderers reproduce every already-approved digest in `main`, so the single value they disagree with `main` about is the one this merge actually moved. The full package `go test` is green on the resolved tree, and both cluster overlays build.

**Conservation.** The whole-surface fingerprint is the only control that moved: the run reports exactly one problem — this aggregate — plus zero per-identity mismatches and zero missing resources against `expectedRenderedHashes`, so every individually approved entry is byte-identical to its recorded value. It also reports the same **35** `unresolved Flux substitution` notes that clean `main` reports; per the validator's own comment those are diagnostic rather than a control, emitted only alongside an aggregate mismatch to explain a hash that moved.

Two notes for whoever reads this next:

- The earlier comment's claim that this value could only come from CI does not hold. A local render on the approved toolchain reproduces it, provided the renderer is checksum-verified into a scratch `PATH` — never by replacing the shared `kubectl`, which carries prod access.
- `🔐 Validate EKS Authorization` and `🧪 Test` both report this digest, but they are **one** renderer, not two: both execute the same code on the same toolchain. The second renderer above is deliberately a different kubectl.

The systemic cause — every merge into a fingerprint-bearing branch re-conflicting this constant — is tracked on #2768.

devantler:


New head `8d607154` resolves the `expectedRenderedSurfaceSHA` conflict that #3171's merge reintroduced. CI is fully green here and `mergeStateStatus` is `CLEAN`; the re-measured digest and its two-renderer evidence are in the comment above. This PR is gated only on a current-head review.

@coderabbitai review

devantler:

## Merge record @ `8d6071545aad00773a4230b9c4723366140b7ec8`

Hygiene pentad clear at this head: status rollup `SUCCESS` with zero non-green contexts, 2 review
threads both resolved, no non-thread review findings (no CodeRabbit review body at this head, and no
Codex comment-form `## Review finding` sections), `mergeStateStatus: CLEAN`.

**Green review** is CodeRabbit's current-head verdict reply: *"No findings at
`8d6071545aad00773a4230b9c4723366140b7ec8`"* — the full 40-character sha, exactly equal to
`headRefOid` — followed by eight diff-specific findings-free observations (the `gateway-tenant-edit`
grant set, the tenant-only aggregation label, all four listeners pinning `allowedRoutes.kinds` to
`HTTPRoute`, and the re-measured `expectedRenderedSurfaceSHA`).

**One caveat, recorded rather than glossed:** this head's `CodeRabbit` commit status still reads
`Review skipped: automatic reviews are disabled` — the never-reviewed default — rather than
`Review completed`, so the status corroborator is absent even though the verdict artifact is present
and unambiguous. I am treating the verdict as the green on substance: its content is specific to this
diff and could not have been produced without a review, which is the only thing the corroborator
exists to establish. Noting it because the same shape on a weaker artifact should *not* be read this
way.

**Behaviour observed.** The boundary test `scripts/tests/test-tenant-route-hostname-boundary.sh`
renders the Hetzner overlay and rejects both an RBAC resource-set widening and any listener admitting
a route kind other than `HTTPRoute` — so the fix is exercised by a test that discriminates the two
states, and CI runs it on Kubernetes changes. The authorization fingerprint was independently
re-measured by two different renderers, each first proved against clean `main` as a matched control.

Enqueuing head-pinned; `main` is merge-queue gated so the queue selects the strategy.

PR #2717, opened 2026-07-18
ci(cd): restrict manual prod deploys to main

devantler:
### Motivation
- Prevent `workflow_dispatch` runs from selecting arbitrary branch refs and executing branch-controlled local composite actions with production secrets by ensuring manual deploys only run from `refs/heads/main`.

### Description
- Add a job-level guard `if: ${{ github.ref == 'refs/heads/main' }}` and pin the checkout to `ref: refs/heads/main` in `.github/workflows/cd.yaml` so the local `./.github/actions/deploy-prod` composite action is always executed from `main`.

### Testing
- Ran a quick YAML load-check via a Ruby script that validated the new `if` guard and pinned checkout and it succeeded.
- Ran `git diff --check` to ensure no whitespace or diff issues and it succeeded.
- An attempted Python/YAML-based check could not run due to a missing `PyYAML` dependency, and `actionlint` was not available in the environment, so those additional static checks were not executed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52a7f400bc832b851cb078d0beb22a)

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -19,6 +19,9 @@ concurrency:
 jobs:
   deploy-prod:
     name: 🚀 Deploy to Production
+    # workflow_dispatch lets an operator choose a ref. Only deploy from main
+    # so prod secrets never flow into branch-controlled workflow/action code.
+    if: ${{ github.ref == 'refs/heads/main' }}
     runs-on: ubuntu-latest
     environment: prod
     permissions:
@@ -30,6 +33,7 @@ jobs:
       - name: 📑 Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
+          ref: refs/heads/main
           persist-credentials: false
 
       - name: 🚀 Deploy to Production


devantler:

devantler:
@cursor review

.github/workflows/cd.yaml
```diff
@@ -19,6 +19,9 @@ concurrency:
 jobs:
   deploy-prod:
     name: 🚀 Deploy to Production
+    # workflow_dispatch lets an operator choose a ref. Only deploy from main
+    # so prod secrets never flow into branch-controlled workflow/action code.
+    if: ${{ github.ref == 'refs/heads/main' }}
```
devantler, line 24: > 🤖 Generated by the Agentic Engineer

[P1] Enforce the dispatch ref outside the selected workflow

`workflow_dispatch` accepts a branch or tag ref, and GitHub runs the workflow for that selected ref. Because this check lives inside the same branch-controlled workflow file, a ref that can change the workflow can remove the check, raise `GITHUB_TOKEN` permissions, or add steps before the pinned checkout. Pinning only the checkout therefore does not establish the stated boundary that branch-controlled workflow code cannot reach production-capable credentials. The live `prod` environment currently has `protected_branches: true`; keep and test that server-side restriction as the secrets boundary, and use an enforcement mechanism outside the selected ref for any write-capable `GITHUB_TOKEN` boundary.

devantler (commented):

Static exact-head review of `f2c9756439c33e210408647322868b85a6799614`; one blocking security finding.

.github/workflows/cd.yaml
devantler: > 🤖 Generated by the Agentic Engineer

Confirmed the finding against live server state at head f2c9756439c33e210408647322868b85a6799614. The prod environment has deployment_branch_policy.protected_branches=true and custom_branch_policies=false; main reports protected=true from the active rulesets, while this non-main PR branch reports protected=false. That is the positive and negative boundary: main can reach prod and the selected branch cannot.

GitHub documents that environment deployment branch rules are matched against GITHUB_REF and that environment secrets are unavailable until protection rules pass: https://docs.github.com/en/actions/reference/workflows-and-actions/deployments-and-environments

The P1 is valid: a check inside branch-selected workflow code cannot establish the credential boundary. Because the live server-side policy already establishes and passed that boundary test, shipping this branch-controlled guard would not strengthen it and would preserve a misleading security claim. Closing this draft instead of merging the invalid fix.

PR #2719, opened 2026-07-18
fix(github): restrict OCIRepository cosign subject to protected release tags

devantler:
### Motivation
- Prevent a supply-chain escalation where the `github-config` OCI artifact source would accept artifacts signed by the shared `publish-manifests` reusable workflow at arbitrary refs, allowing non-protected branch/PR signatures to be trusted. 
- The `github-config` app runs with namespace-scoped permissions to create/update Crossplane GitHub managed resources and `ExternalSecret`s, so an attacker-published artifact could mutate the org if the signer policy is too broad.

### Description
- Tightened the cosign OIDC subject check in `k8s/bases/apps/github-config/oci-repository.yaml` so the `matchOIDCIdentity.subject` requires `@refs/tags/vX.Y.Z` release-tag subjects instead of allowing `@.+`, rejecting branch and pull-request refs. 
- Kept the temporary repository-name alternation `(reusable-workflows|actions)` to preserve the planned signer repo transition while still enforcing tag-only acceptance. 
- Added clarifying comment text explaining the rationale and the narrower acceptance rule.

### Testing
- Ran a local regex smoke test (`python3` snippet) that confirmed `refs/tags/v1.2.3` matches and branch/PR subjects do not, and the test succeeded. 
- Ran `git diff --check` to ensure no whitespace/errors and it succeeded. 
- Ran `python3 scripts/validate-naming.py` and it succeeded against repository naming rules. 
- `kubectl kustomize k8s/clusters/local/` and `kubectl kustomize k8s/clusters/prod/` could not be executed in this environment because `kubectl` is not installed, so full kustomize render validation was not run here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52acebb2b0832ba49a008c8ab883ac)

### DIFF
--- a/k8s/bases/apps/github-config/oci-repository.yaml
+++ b/k8s/bases/apps/github-config/oci-repository.yaml
@@ -33,8 +33,10 @@ spec:
       # subjects are accepted via ONE regex-alternated entry — cosign's keyless
       # attestation verification rejects MULTIPLE identity entries outright
       # ("unsupported: multiple identities are not supported at this time"), so
-      # the transition must be expressed inside a single subject regex. Narrow
-      # the alternation to actions-only once the first actions-signed artifact
-      # has been published and verified.
+      # the transition must be expressed inside a single subject regex. Accept only
+      # protected release-tag invocations of the reusable signer; do not allow
+      # arbitrary branch or pull-request refs to publish higher-semver GitHub
+      # org-management artifacts. Narrow the alternation to actions-only once
+      # the first actions-signed artifact has been published and verified.
       - issuer: '^https://token\.actions\.githubusercontent\.com$'
-        subject: '^https://github\.com/devantler-tech/(reusable-workflows|actions)/\.github/workflows/publish-manifests\.yaml@.+$'
+        subject: '^https://github\.com/devantler-tech/(reusable-workflows|actions)/\.github/workflows/publish-manifests\.yaml@refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$'


devantler:

devantler:
@cursor review

k8s/bases/apps/github-config/oci-repository.yaml
```diff
@@ -33,8 +33,10 @@ spec:
...
-      # the alternation to actions-only once the first actions-signed artifact
-      # has been published and verified.
+      # the transition must be expressed inside a single subject regex. Accept only
+      # protected release-tag invocations of the reusable signer; do not allow
+      # arbitrary branch or pull-request refs to publish higher-semver GitHub
+      # org-management artifacts. Narrow the alternation to actions-only once
+      # the first actions-signed artifact has been published and verified.
       - issuer: '^https://token\.actions\.githubusercontent\.com$'
-        subject: '^https://github\.com/devantler-tech/(reusable-workflows|actions)/\.github/workflows/publish-manifests\.yaml@.+$'
+        subject: '^https://github\.com/devantler-tech/(reusable-workflows|actions)/\.github/workflows/publish-manifests\.yaml@refs/tags/v[0-9]+\.[0-9]+\.[0-9]+$'
```
devantler, line 42: [P1] Keep the SHA-pinned signer identity accepted

The current producer in `devantler-tech/.github/.github/workflows/cd.yaml` invokes `devantler-tech/actions/.github/workflows/publish-manifests.yaml@061b345a7351b0caf23055caea63ccd08b75e20e`, so its Fulcio SAN ends in a 40-hex workflow ref, not `refs/tags/v...`. This tag-only matcher therefore rejects every currently produced github-config artifact and freezes Flux verification/reconciliation when the next artifact is selected. Platform main already fixed this in #2816 after checking the deployed 1.20.2 signature, using the actions-only `([0-9a-f]{40}|refs/tags/v.+)` ref matcher. Preserve that SHA-or-tag shape (and drop the archived `reusable-workflows` alternative), with positive SHA/tag and negative floating-ref coverage.

devantler (commented):

## Static exact-head review

Reviewed this ownership-unverified draft at exact head `29895d7d47086b9acd4d4ab5a18b69f4a8156616` without checkout or execution. One blocking finding is anchored below.

devantler:

**Closing as superseded — `main` now implements this PR's security objective, and more strictly.**
Merging this would be a net regression rather than a fix.

This PR narrows the `github-config` OCIRepository cosign subject so only protected release tags are
trusted. Since it was opened, `main` moved the same check to **commit-SHA pinning** and dropped the
archived repo from the alternation (#3022). I compared the two regexes by running real subjects
through both:

| candidate cosign subject | this PR | `main` |
|---|---|---|
| `actions/...publish-manifests.yaml@refs/tags/v1.2.3` | accepts | rejects |
| `reusable-workflows/...@refs/tags/v1.2.3` | **accepts** | rejects |
| `actions/...@<40-hex commit>` | rejects | **accepts** |
| `actions/...@refs/heads/main` | rejects | rejects |

Two things follow:

1. **The vulnerability this PR closes is already closed.** The last row is the escalation named in
   the motivation — an artifact signed from an arbitrary branch/PR ref — and `main` rejects it.
2. **Merging this would re-widen the policy.** Row 2 is the problem: this branch still accepts
   `devantler-tech/reusable-workflows`, which is now **archived** (`archived=true`), at a *movable*
   tag. `main` accepts neither that repo nor a tag ref. A commit SHA cannot be repointed, so `main`'s
   form is strictly the stronger of the two on both axes — repo scope and ref immutability.

This branch's own comment anticipated exactly this: *"Narrow the alternation to actions-only once the
first actions-signed artifact has been published and verified."* That has since happened, and `main`
went further by pinning to commits.

There is no still-valid finding left to re-file — the residual concern is entirely covered by the
current `main` policy, so I am not opening a follow-up issue.

I stopped at the conflict rather than resolving it, because resolving toward this branch is precisely
the silent revert described above. Reopen if I have misread the intent — happy to be corrected.

PR #2729, opened 2026-07-18
fix(security): avoid logging OpenBao unseal keys in argv

devantler:
### Motivation

- Tetragon was configured to export built-in `process_exec` events to stdout, and OpenBao initialization/unseal paths pass the unseal key on the command line, which makes the key persistently discoverable via the existing Alloy→Loki pipeline.  
- The change prevents sensitive unseal keys from appearing in process argv (and thus in Tetragon events shipped to Loki) while preserving the existing init/unseal semantics and retry behaviour.

### Description

- Replace direct argv-style calls like `bao operator unseal "$UNSEAL_KEY"` with piping the key to `bao operator unseal` via stdin using `printf '%s\n' "$UNSEAL_KEY" | bao operator unseal` to avoid placing secrets in `argv` in `postStart` hooks and jobs.  
- Files modified: `k8s/bases/infrastructure/controllers/openbao/helm-release.yaml` and `k8s/bases/infrastructure/vault-config/job.yaml`.  
- Commit: `fix(security): avoid logging OpenBao unseal keys in argv` (two-file change, 3 insertions, 3 deletions). 

### Testing

- Ran an `rg` pattern check to ensure no remaining `bao operator unseal` argv usages matching the previous patterns and the check succeeded.  
- Parsed the modified YAMLs with a Ruby YAML loader and ran `python3 scripts/validate-embedded-json.py` and `python3 scripts/validate-naming.py`, all of which succeeded in this environment.  
- Ran `git diff --check` with no whitespace/errors reported, and attempted `kubectl kustomize` / `ksail workload validate` could not be executed in this environment because `kubectl` and `ksail` are not installed (manual `ksail`/`kubectl kustomize` validation is recommended before merge).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52ce9724a8832bbbb606a027bbd8a3)


devantler:

devantler:
@cursor review

devantler (commented):

[P1] Close the remaining runbook leak before calling this path fixed

`docs/dr/openbao-raft-ha-migration.md:117-124` still tells the operator to run `bao operator unseal <new-unseal-key>` through `kubectl exec`. Following that procedure gives the `bao` process inside the pod the real unseal key in argv, so Tetragon’s `process_exec` exporter can still persist it to Loki during the canonical recovery procedure—the exact exposure this PR removes from hooks and jobs. Please switch all three documented manual unseal commands to stdin or interactive input as well (invoke `bao operator unseal` without the key argument and preserve stdin/TTY as appropriate). Otherwise, following the DR runbook reintroduces the credential disclosure after this change merges.

devantler:

Unblocked the `🔐 Validate EKS Authorization` failure that has held this PR since 2026-08-14. It was a moved aggregate fingerprint, not an authorization change.

**Evidence** — rendered with the pinned kubectl v1.36.2 / kustomize v5.8.1 (checksum-verified against the published `.sha256`, matching what CI installs), across all five `authorizationOverlayPaths`, with current `main` merged first:

| check | result |
|---|---|
| authorization membership | **identical, 71 documents** — nothing added, removed or renamed |
| authorization text | **byte-identical, 39317 bytes** |
| negative control (drop one member) | **fired** — the comparison is live |

The OpenBao HelmRelease sits inside the selected surface, so editing its embedded unseal script moves the aggregate without touching any Role/ClusterRole/RoleBinding/ClusterRoleBinding/ServiceAccount. That over-breadth is [#2768](https://github.com/devantler-tech/platform/issues/2768).

Two notes for anyone reproducing this locally:

- **`main` must be merged BEFORE approving.** A `pull_request` job renders the *merge* commit; a local run renders the branch. This branch was 9 commits behind, so a value computed on it would have gone stale on the next CI render.
- **The 35 `unresolved Flux substitution` lines are diagnostic, not a second failure.** The validator emits them only alongside a fingerprint mismatch — deliberately, per the comment above the constant — and they clear with the re-approval.

⚠️ Also worth knowing: **`go test` without `-count=1` returns a cached pass here.** The validator reads the k8s YAML at runtime, which Go's test cache does not track, so editing a manifest and re-running reports `ok (cached)` on a tree that genuinely fails. I hit that mid-diagnosis and it briefly inverted my conclusion. Use `go test -count=1 ./scripts/validate-eks-ci-role-policy`.

The security fix itself is unchanged and unreviewed by me here — this only clears the CI blocker.

devantler:



The `🔐 Validate EKS Authorization` blocker is cleared and this PR is now `CLEAN` with every check green and no unresolved threads. It has never had a review on any lane, so requesting the first one at the current head.

@coderabbitai review

devantler:

## P0 — this change stops OpenBao from unsealing at all

`bao operator unseal` **refuses a piped key**. It checks that fd 0 is a terminal *before* it contacts the server, so all three converted call sites fail 100% of the time and never reach `sys/unseal`.

Reproduced against the real binary, **OpenBao v2.6.1** (`ba7ad886`, 2026-07-22), no server needed:

```
$ printf 'somekey\n' | bao operator unseal ; echo "EXIT=$?"
Unseal Key (will be hidden):
An error occurred attempting to ask for an unseal key. The raw error message
is shown below, but usually this is because you attempted to pipe a value
into the unseal command or you are executing outside of a terminal (tty). ...
The raw error was:  file descriptor 0 is not a terminal
EXIT=1
```

The pre-change form is fine — it gets past argument handling and reaches the network:

```
$ bao operator unseal somekey
Error unsealing: Put "https://127.0.0.1:8200/v1/sys/unseal": dial tcp ...: connection refused
```

### `-` is not a workaround either

`operator rekey` accepts `-` for stdin; `operator unseal` does not. It is taken as the literal key. Captured request body:

```
$ printf 'STDINKEY\n' | BAO_ADDR=http://127.0.0.1:8200 bao operator unseal -
{"key":"-","reset":false,"migrate":false}
```

### What each site does now

- `openbao/helm-release.yaml` postStart — `|| true` swallows the failure, so the retry loop spins 24 times and **exits 0 with the pod still sealed**. That is the exact failure mode the comment directly above this block was written about ("a SINGLE transient failure ... was swallowed by `|| true` and the pod stayed sealed until manual intervention").
- `vault-config/job.yaml` (unseal `openbao-0`) — no guard, so **cold bootstrap fails hard**.
- `vault-config/job.yaml` (unseal followers) — `|| echo "WARNING..."` means followers **silently never unseal**.

CI is green because the manifests are still well-formed YAML; nothing in the pipeline executes `bao`.

### On the underlying goal

Keeping the key out of `argv` is worth doing, but note the marginal gain is small here — the key is already readable in-container at the mounted `unseal-key` path, and a container's `ps` shows only its own processes. Options that actually work: call `sys/unseal` over HTTP with the body on stdin (`curl --data @-`, key never in `argv`), or move to auto-unseal, which removes the static key handling altogether.

I have not pushed a fix — the replacement approach is a design call for this PR's lane rather than a mechanical edit. Flagging it as blocking so it does not merge in this form; happy to implement whichever direction is preferred.

devantler:

Closing this: **the mechanism does not work, and the concern is better fixed one layer down.**

**Why it cannot merge as it stands.** All three call sites were rewritten from
`bao operator unseal "$KEY"` to `printf '%s\n' "$KEY" | bao operator unseal`. `bao operator unseal`
checks that **file descriptor 0 is a terminal before it contacts the server**, so the piped form fails
100% of the time and never reaches `sys/unseal`. Measured on the real binary (OpenBao v2.6.1, no
server required):

> ```
> $ printf 'somekey\n' | bao operator unseal ; echo "EXIT=$?"
> Unseal Key (will be hidden):
> ... usually this is because you attempted to pipe a value into the unseal command ...
> The raw error was:  file descriptor 0 is not a terminal
> EXIT=1
> ```

`-` is not a stdin form for `unseal` either (it is for `operator rekey`); it is taken as the literal
key — confirmed against a local listener, which received `{"key":"-","reset":false,...}`.

The blast radius sits in the error handling: two of the three sites are guarded by `|| true` and
`|| echo WARNING`, which would have converted a guaranteed failure into **silently never unsealing**.
The file's own comment records a prior incident where `|| true` left pods sealed until manual
intervention. Every check on this PR was green because the manifests are still valid YAML and nothing
in CI executes `bao` — well-formed, but not working.

**Why not just repair the mechanism here.** Keeping the key out of `argv` at these three sites would
mean adding an HTTP client and hand-built request bodies to the unseal path — more moving parts in
the one path whose failure mode is a sealed cluster — and it would still only cover three call sites.
The exposure is not specific to them: the export pipeline emits every process's arguments
cluster-wide, so the same leak exists for every workload that takes a credential as an argument,
including ones we do not own.

**Where the fix now lives:** #3157 closes the whole class at the exporter, with a both-states
verification. That subsumes this PR's intent without touching the unseal path.

Superseded by #3157. Branch left for the normal cleanup sweep.

PR #2730, opened 2026-07-18
ci(security): avoid signing mutable prod manifest tag

devantler:
### Motivation

- Prevent a TOCTOU supply-chain race where the merge-queue deploy signed whatever digest the mutable `latest` tag resolved to after `workload push`, which could let an attacker with GHCR write access get an attacker-controlled manifest signed and deployed.

### Description

- Add a pre-push step in the deploy composite (`.github/actions/deploy-prod/action.yml`) that rewrites the checked-out `ksail.prod.yaml` `spec.workload.tag` to a per-run, unshared tag and exposes it as `KSAIL_WORKLOAD_TAG` for the job. 
- Change the cosign signing step to resolve and sign `ghcr.io/devantler-tech/platform/manifests:${KSAIL_WORKLOAD_TAG}` instead of the shared mutable `latest` tag so signing is bound to the artifact produced by this run. 
- Preserve the existing push → sign → attest → reconcile flow and keep attestation/verification behaviour unchanged (Flux verification still expects a cosign-signed digest produced by the deploy workflow).

### Testing

- Loaded the modified composite action with `ruby -e 'require "yaml"; YAML.load_file(".github/actions/deploy-prod/action.yml")'` which succeeded. 
- Ran `git diff --check` and YAML parse checks which reported no syntax issues. 
- Executed a local Python-based simulation that confirmed the deploy-tag rewrite finds and replaces `spec.workload.tag` in `ksail.prod.yaml` and sets `KSAIL_WORKLOAD_TAG`, which the signing step then uses. 
- Verified the working tree is clean after committing the change and that the composite action remains syntactically valid.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52ce325234832bac8d3dcb1c6e55c6)

### DIFF
--- a/.github/actions/deploy-prod/action.yml
+++ b/.github/actions/deploy-prod/action.yml
@@ -133,6 +133,48 @@ runs:
     # down, `ksail cluster update` ran first, hard-failed the job, and froze
     # ALL app/config delivery to an otherwise-healthy cluster (the apiserver
     # and Flux were fine on the two surviving control-plane nodes).
+    - name: 🏷️ Pin prod workload to this run's OCI tag
+      # ksail.prod.yaml intentionally uses the human-friendly `latest` tag in
+      # Git, but deploys must not sign a digest looked up through that shared
+      # mutable tag after `workload push`: another GHCR writer could retag
+      # `latest` in the push→sign window and trick this trusted workflow into
+      # signing their digest. For the live deploy, rewrite only the checked-out
+      # config to a per-run tag before both `workload push` and `workload
+      # reconcile`; the signing step then resolves that unshared tag rather
+      # than the repository-wide `latest` pointer.
+      shell: bash
+      run: |
+        set -euo pipefail
+        DEPLOY_TAG="sha-${GITHUB_SHA}-run-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+        python3 - <<'PY' "${DEPLOY_TAG}"
+        from pathlib import Path
+        import sys
+
+        deploy_tag = sys.argv[1]
+        path = Path("ksail.prod.yaml")
+        lines = path.read_text().splitlines(keepends=True)
+
+        in_workload = False
+        replaced = False
+        for index, line in enumerate(lines):
+          stripped = line.strip()
+          if line == "  workload:\n":
+            in_workload = True
+            continue
+          if in_workload and line.startswith("  ") and not line.startswith("    "):
+            break
+          if in_workload and stripped.startswith("tag:"):
+            lines[index] = f"    tag: {deploy_tag}\n"
+            replaced = True
+            break
+
+        if not replaced:
+          raise SystemExit("Could not find spec.workload.tag in ksail.prod.yaml")
+
+        path.write_text("".join(lines))
+        PY
+        echo "KSAIL_WORKLOAD_TAG=${DEPLOY_TAG}" >> "${GITHUB_ENV}"
+
     - name: 📦 Push manifests to GHCR
       shell: bash
       env:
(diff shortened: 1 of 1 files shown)


.github/actions/deploy-prod/action.yml
```diff
@@ -156,6 +156,48 @@ runs:
...
+      # ksail.prod.yaml intentionally uses the human-friendly `latest` tag in
+      # Git, but deploys must not sign a digest looked up through that shared
+      # mutable tag after `workload push`: another GHCR writer could retag
+      # `latest` in the push→sign window and trick this trusted workflow into
+      # signing their digest. For the live deploy, rewrite only the checked-out
+      # config to a per-run tag before both `workload push` and `workload
+      # reconcile`; the signing step then resolves that unshared tag rather
+      # than the repository-wide `latest` pointer.
+      shell: bash
+      run: |
```
devantler, line 169: > 🤖 Generated by the Agentic Engineer

[P2] Remove Python from the production deploy path

This adds a Python mutator to a repository workflow, but the portfolio contract permits repository scripting only in Bash or Go. That makes this deployment change non-promotable even if the current merge conflict and failing checks are cleared. Rewrite the tag update with the repository’s Bash/yq tooling (or Go) and keep an explicit failure when spec.workload.tag is absent.

devantler (commented):

Static exact-head review; branch code was not checked out or executed.

devantler:

Picked this up as rung-1 work (open PR, no activity since 2026-08-04). The **goal is sound and worth finishing** — signing a digest resolved through the shared mutable `latest` tag is a real TOCTOU: another GHCR writer could retag `latest` inside the push→sign window and get this trusted workflow to sign their bytes. Three things block it as written, recorded here so whichever lane picks it up does not re-derive them.

**1. It introduces Python, which the portfolio stack rule forbids.** The new step runs `python3 - <<'PY'` to rewrite `spec.workload.tag`. Scripting here is bash or Go, never Python. This is not the embedded-interpreter carve-out — that covers hosts whose API admits only their own bundled Python (Blender `bpy` and similar), recognised by the invocation. This is a free choice, and `yq` is already a documented platform prerequisite, so the same edit is a one-liner:

> yq -i '.spec.workload.tag = strenv(DEPLOY_TAG)' ksail.prod.yaml

That also drops the hand-rolled indentation-scanning parser, which is the fragile part of the current step.

**2. The `zizmor/github-env` error is the failing check** (`.github/actions/deploy-prod/action.yml:169`) — `echo "KSAIL_WORKLOAD_TAG=..." >> "$GITHUB_ENV"`. In substance the value is safe: it is built only from `GITHUB_SHA`, `GITHUB_RUN_ID` and `GITHUB_RUN_ATTEMPT`, none of them attacker-controlled. But the standard remediation is also the better pattern — give the step an `id:` and write to `$GITHUB_OUTPUT`, then read `${{ steps.<id>.outputs.tag }}` in the signing step. That clears the finding at the root rather than suppressing it, per the fix-vs-except ladder.

**3. It is `DIRTY` against `main`** and needs rebasing before any of the above can be verified.

Not closing it — the security concern is real. Leaving it open with the path named.

devantler:

Closing as **superseded** — `main` already implements this PR's intent, and does it more strongly.

**Why this PR existed:** signing a digest resolved through the shared, mutable `latest` tag leaves a push→sign TOCTOU window in which another GHCR writer could retag `latest` and have this trusted workflow sign their digest.

**Why it is no longer needed:** the publish sequence this PR edits has since been extracted into `.github/actions/deploy-prod/publish-platform-manifests`, which closes the same window by a better route:

- pushes to a per-run, unshared `staging-<sha>-<run-id>-<attempt>` reference (the same uniqueness property this PR proposed);
- signs, attests and verifies evidence against **that staging digest**, never a shared tag;
- promotes to `latest` **by digest** (`imagetools create --tag latest <name>@<digest>`) and then re-reads `latest`, failing the deploy if it does not resolve to the expected digest.

So `latest` is never a resolve-then-sign *source* any more — it is only a promotion *destination*, pinned to an already-evidenced digest and verified after the fact.

**No residual findings to re-file.** The superseding design also avoids both defects this PR would still have had to fix before it could merge:

- it passes the tag positionally to `ksail workload push`, so no `ksail.prod.yaml` rewrite and **no `python3` step** (the portfolio stack is bash-or-Go, never Python);
- it hands the tag between steps via `$GITHUB_OUTPUT`, so the `zizmor` `github-env` error on this branch (`dangerous use of environment file`, `action.yml:169`) does not arise.

Verified against `origin/main`: the extracted action carries the per-run staging tag, and the deploy action contains zero `GITHUB_ENV` writes and zero `python3` invocations.

PR #2735, opened 2026-07-18
fix(velero): isolate kopia repository password

devantler:
### Motivation

- A PushSecret mirrored Velero's Kopia `repository-password` into the shared `infrastructure/backup/*` OpenBao prefix, which co-located backup ciphertext access credentials and the decryption key and thereby allowed any principal with the broad backup-read policy to decrypt backups offline.
- The change isolates the Kopia decryption key so that backup storage credentials and the repository password are not readable through the same broadly-scoped Vault policy.

### Description

- Move the PushSecret target from `infrastructure/backup/velero-repo` to a dedicated `infrastructure/velero/repo` path so the Kopia password is not placed under the shared backup prefix.
- Add a dedicated Vault path `secret/data/infrastructure/velero/*` to the seed-write policy block so PushSecret seeding retains write access to the new, more specific location without widening the existing backup-read policy.
- Update the PushSecret comment to describe the new seed path and note the intent to avoid sharing the broadly readable backup-object credential prefix.

### Testing

- Ran `git diff --check` and it reported no whitespace or patch errors; the check passed.
- Ran `python3 scripts/validate-naming.py` and `python3 scripts/validate-embedded-json.py` and both validations passed.
- `kubectl kustomize` and `ksail workload validate` were not executed in this environment because `kubectl` and `ksail` are not installed, so a full schema-aware `ksail` validation was not performed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52d45cffec832bbe1d0fd417972eb0)


devantler:

devantler:
@cursor review

devantler (commented):

Static exact-head review of `222643b4c550cb8683be77aa36d60c6543cc541f`; the branch was not checked out or executed.

k8s/bases/infrastructure/vault-seed/push-secret-seed-velero-repo-credentials.yaml
```diff
@@ -35,5 +36,5 @@ spec:
     - match:
         secretKey: repository-password
         remoteRef:
-          remoteKey: infrastructure/backup/velero-repo
+          remoteKey: infrastructure/velero/repo
```
devantler, line 39: > 🤖 Generated by the Agentic Engineer

**[P1] Remove the old Vault key after copying it**

Changing `remoteKey` makes ESO start writing the repository password at the new path, but a PushSecret does not migrate or delete the existing KV-v2 entry at `infrastructure/backup/velero-repo`. On the existing production OpenBao instance, that old entry therefore remains readable through the broad `infra-backup-readonly` policy, so the backup decryption key is still exposed to the audience this change intends to remove. Add an explicit one-time migration that verifies the new copy and then permanently removes the old key's versions and metadata before treating the isolation as complete.

Reviewed at `222643b4c550cb8683be77aa36d60c6543cc541f`.

devantler: > 🤖 Generated by the Agentic Engineer

Re-verified at the current head `77a4fedc6e` — **this finding still stands**, six days on.

The PR touches only two files (`vault-config/job.yaml` +3, and this PushSecret +3/-2). There is no `kv delete`, no metadata destroy, and no migration step anywhere in the diff, and `main` still carries `remoteKey: infrastructure/backup/velero-repo`. So on merge ESO begins writing to the new path while the existing KV-v2 entry at the old one is left in place and readable through `infra-backup-readonly` — the exact audience this change exists to remove. Every check is green (21 success, 12 skipped), which is why this has been easy to miss: the gap is behavioural, not structural.

Leaving the PR parked on this rather than promoting it. To clear it, the diff needs a one-time migration that verifies the new copy is readable and then removes the old key's versions **and** its metadata — deleting only the latest version leaves prior versions recoverable, so the exposure would survive that too.

devantler:

**Fixed at `8bec29b54188f79bcb674188957045b0c46656e1`.**

The finding was correct: a PushSecret only writes the new path, so the KV-v2 entry at
`infrastructure/backup/velero-repo` survived and stayed readable through `infra-backup-readonly` —
the isolation this PR claims did not actually happen.

The vault-config Job now retires that entry (new step `1f`), with two properties that matter:

- **Ordered so it can never lose the password.** It destroys the old copy *only after* confirming
  the new path holds `repository-password`. If ESO has not synced yet it logs a warning and leaves
  the old copy alone, so a run that lands early is a no-op rather than a data-loss event. Both
  branches are idempotent.
- **`kv metadata delete`, not `kv delete`.** A plain `kv delete` soft-deletes only the latest version
  and is trivially undeleted, which would not remove the exposure. `metadata delete` removes every
  version plus the metadata.

Also fixed while here: the seeding policy granted `secret/data/infrastructure/velero/*` but not
`secret/metadata/infrastructure/velero/*`. Every other prefix in that policy has both — velero was
the only asymmetric one (verified: 16 data / 15 metadata before, 16 / 16 after, with the check
re-run against the pre-change file as a control to confirm it detects the gap).

Validated: `ksail workload validate` passes for both the local and prod overlays (564 files, 0
errors), and the embedded Job script passes `sh -n` (confirmed by a negative control that fails).

devantler:



Requesting the first review at this head.

@coderabbitai review

devantler:

**Readiness record — promoting.** All three conditions hold at `8bec29b54188f79bcb674188957045b0c46656e1`.

**1. Programmatically tested.** 33 checks green, including `🧪 Validate Manifests`, `🔐 Validate EKS Authorization`, `🏷️ Validate Naming Conventions`, `🖋️ Validate Publication Contract`, `kubescape`, `CodeQL` and `mega-linter`. This closes the gap the description names — the schema-aware validation that could not run in the authoring environment did run here.

**2. Reviewed.** CodeRabbit at this head: *"No actionable comments were generated in the recent review"*, naming `ff43488f…8bec29b5`, corroborated by a `Review completed` status. 0 unresolved threads, 0 body findings.

**3. Tried and evaluated as a user.** The changed behaviour here is a shell block, so I exercised it directly against a stubbed `bao` across all four reachable states:

| new path | old path | result | destructive call |
|---|---|---|---|
| present | present | retires the old copy | 1 (the intended delete) |
| present | absent | no-op, "already retired" | 0 |
| **not synced** | present | warns, **leaves the old copy** | 0 |
| absent | absent | warns | 0 |

The safety property holds: the irreversible `kv metadata delete` fires only once `infrastructure/velero/repo` is confirmed to hold `repository-password`, so no ordering against the PushSecret sync can leave the platform with neither copy. Both branches are idempotent.

**Consumer check.** A repo-wide search finds `infrastructure/backup/velero-repo` referenced *only* by the PushSecret that writes it — no `ExternalSecret` or controller reads it back — so relocating the path breaks no consumer. The seed policy correctly gains both `secret/data/infrastructure/velero/*` and `secret/metadata/infrastructure/velero/*`.

**Base currency.** 18 behind main, which is what the merge queue exists to resolve. The surface guard is unaffected: this PR touches no authorization-surface document, so the merge result's fingerprint is main's own approved value — evidenced by this PR adding two Vault policy paths while `🔐 Validate EKS Authorization` stayed green.

PR #2737, opened 2026-07-18
fix(apps): pin tenant OCI manifest tags

devantler:

### Why

Both tenant apps tracked their manifests with an open-ended `>=1.0.0` range, so Flux automatically selected and deployed **any** newer tag pushed to that registry path. Whoever can publish a tag chose what production runs, with no review in between.

### What

Pins each tenant to the exact manifest revision production is already reconciling, so tenant deployments become a reviewed step rather than an unattended one. Nothing that is running changes.

**Security/DevEx:** the floor gains removal of unattended tag selection for both tenants; the everyday path is unchanged for app development — only a tenant *manifest* rollout now needs a pin bump.

⚠️ **Operational note:** future tenant releases no longer roll out on their own. Each one needs a PR bumping the pin in `k8s/bases/apps/<app>/oci-repository.yaml`.


devantler:

devantler:
@cursor review

k8s/bases/apps/ascoachingogvaner/oci-repository.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
 spec:
   interval: 1m
   ref:
-    semver: ">=1.0.0"
+    tag: 1.1.0
```
devantler: > 🤖 Generated by the Agentic Engineer

[P1] Rebase these pins onto the deployed revisions

This pins both tenant sources to old tags (`1.1.0` here and `1.5.9` for Wedding). Read-only production evidence shows Flux currently reconciles AS Coaching `1.13.4` and Wedding `1.15.10`, and those versions match each repository's latest public release. Merging this head would therefore roll both tenants backward by many releases. Update the exact pins from the current verified artifact revisions (or explicitly prove an intentional rollback is safe), and add a guard that rejects a selected tag older than the deployed/current release.

devantler (commented):

Static exact-head review of `34665f7d3c28803b186fa34df53e0c516f5a255e`; the branch was not checked out or executed.

k8s/bases/apps/ascoachingogvaner/oci-repository.yaml
devantler: > 🤖 Generated by the Agentic Engineer

Confirmed and fixed at `665a4d38`.

I verified the finding against live production rather than taking it on trust. Read-only `admin@prod` reports both `OCIRepository` resources still on `semver: ">=1.0.0"`, with actual reconciled revisions:

| tenant | deployed revision | this branch pinned | now pinned |
|---|---|---|---|
| ascoachingogvaner | `1.13.4` | `1.1.0` | `1.13.4` |
| wedding-app | `1.15.10` | `1.5.9` | `1.15.10` |

Both deployed revisions also match each repository's latest release (`v1.13.4` published 2026-08-03, `v1.15.10` published 2026-07-27), so pinning to them is neither a rollback nor a jump ahead of what is running.

Also in this push: `main` was merged (the PR was `DIRTY`), and the conflict in `scripts/validate-eks-ci-role-policy/main.go` was resolved to **main's** content exactly — the branch's fingerprint note and value described the stale `1.1.0`/`1.5.9` surface and were doubly out of date, so carrying them forward would have approved a surface that no longer exists. `git diff origin/main` on that file is now empty.

**One thing is deliberately left for CI, not guessed.** `OCIRepository` is inside the selected authorization surface, so changing these tags moves `expectedRenderedSurfaceSHA`. The approved renderer is kubectl v1.36.2; this host has v1.36.1, which the procedure refuses as unapproved. The constant therefore sits at main's approved value and the `🔐 Validate EKS Authorization` job will report the real one — that measured value is what should replace it, exactly as the branch's own earlier note describes doing.

Validation run locally: `ksail workload validate` and `ksail --config ksail.prod.yaml workload validate` both clean, with `bases/apps/ascoachingogvaner` and `bases/apps/wedding-app` confirmed present in both renders (this repo's kustomize fallback renders neither, so a fallback-only check would have passed vacuously).

**Not addressed here, and worth a reviewer's view:** the second half of this finding asked for a guard rejecting a tag older than the deployed release. I have not added one — it needs a source of truth for "currently deployed" that a static CI check can read, which does not exist in-repo today. Separately, fixed tags mean every future tenant release now needs a platform PR to roll out; that is a real workflow change for two live apps and is a maintainer call rather than mine to settle in a draft.

devantler:

CI is green at the current head and no review has been requested on this PR yet. Requesting one.



@coderabbitai review

devantler:



@codex review

devantler:



CodeRabbit returned `Review rate limited` at this head and Codex replied with an account usage
limit (measured 2026-08-17T00:03:29Z). Advancing to Bugbot per lane order.

devantler:
@cursor review

devantler:

**Blocker: this PR's conflict is an authorization-surface re-measurement, not a mergeable conflict.** Diagnosed, not resolved — recording it so the next pass starts from evidence.

`main` has advanced to `9a84e92b`, and the only conflicting file is `scripts/validate-eks-ci-role-policy/main.go`. Both conflict hunks land on the same thing:

| side | `expectedRenderedSurfaceSHA` |
|---|---|
| this branch | `3fda5879259ea82207cf38faf32a30ed1603f77985fb4e2d61b61a4172792538` |
| `main` | `b5b394181a8f2bc325bd427a90990fbc3872fb1f41f3d938f1e28f0aac1075f4` |

**Neither value is correct for the merge result**, which is why this cannot be hand-merged. That constant is the aggregate fingerprint of the rendered RBAC surface; the branch's value was measured against an older `main`, and `main`'s value describes `main` without this branch. Picking either side commits a fingerprint that describes a tree nobody rendered — and this validator is the gate protecting the `aws/aws` service account, so a wrong value here either fails the build or, worse, passes while describing the wrong surface.

The rest of the state is healthy: **28 checks green, 5 skipped, 0 failing**, and the change itself (pinning tenant OCI manifest tags away from a `>=1.0.0` semver range) is unaffected by the merge.

### What this needs

The two-renderer re-measurement on the **merge result**, with matched controls:

1. Render with the required job's own approved toolchain (kubectl **v1.36.2**, checksum-verified into a scratch dir — never replacing the shared binary, which carries prod access).
2. Render independently via `TestValidateAuthorizationAcceptsCommittedPolicy`, which bypasses the version gate.
3. **Both must first pass on clean `main`** — that control is what makes their agreement evidence rather than two guesses.
4. Rewrite the audit comment to describe *this* merge's delta, and state the conservation counts.

I stopped rather than hurry it: a hasty pass on a prod authorization fingerprint is worse than none. The branch is untouched and no partial merge was pushed.

devantler:



@coderabbitai review

devantler:

**Readiness evaluation @ `977478bd78a72c76c2a4da28da740955936fd954`**

1. **Programmatically tested** — 26 checks pass, 5 skipped, none failing; `mergeStateStatus` CLEAN; 0 unresolved threads; 0 non-thread review findings. The required `🔐 Validate EKS Authorization` job re-approved the surface fingerprint on the merge with exact main `c4fb7f2d`, reporting **0 per-identity mismatches, 0 missing resources, 0 duplicates** — the aggregate was the only control that moved, which is what the two pinned `OCIRepository` documents should do and nothing more.

2. **Reviewed** — CodeRabbit at this head: auto-generated summary refreshed after the request naming this commit, `Review completed`, 0 inline comments, 0 review findings.

3. **Tried and evaluated as a user** — checked the running cluster rather than inferring from the manifests. Both tenants are *already* reconciling the exact revisions this PR pins, while still selecting them through the floating range:

   | Tenant | Live revision now | Pinned to | Live selector today |
   |---|---|---|---|
   | `ascoachingogvaner` | `1.13.4` | `1.13.4` | `>=1.0.0` |
   | `wedding-app` | `1.15.10` | `1.15.10` | `>=1.0.0` |

   So the deployed state does not move: what changes is that a newly published tag can no longer be selected and reconciled without a review. That was the point of the change, and it is the part worth confirming against reality — had either pin disagreed with the live revision, this would have been a silent rollback rather than a pin.

Promoting and enqueuing.

PR #2745, opened 2026-07-18
fix(security): block Hetzner metadata from pods

devantler:

### Motivation

- Prevent workloads running on autoscaler-created Hetzner nodes from reading Talos worker machine config (user-data) via the instance metadata service and thereby exposing reusable cluster bootstrap credentials.

### Description

- Add a Cilium cluster-wide egress deny to block pod access to the Hetzner metadata IP (`169.254.169.254/32`) by extending the existing `require-mutual-auth` cluster policy to include an `egressDeny` clause. 
- Update `docs/node-autoscaling.md` to document the metadata isolation guarantee so operators understand that pod egress to IMDS is blocked in the Hetzner overlay.

### Testing

- Ran `python3 scripts/validate-naming.py` and it passed without naming violations. 
- Verified YAML validity by loading the modified policy with `ruby -e 'require "yaml"; YAML.load_stream(File.read("...") )'` which succeeded. 
- Ran `git diff --check` which reported no whitespace or diff errors, and `kubectl kustomize` could not be executed in this environment because `kubectl` is not installed so overlay kustomize rendering was not run locally.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a52e6398034832b9effa555ef3ac065)


devantler:

devantler:
@cursor review

devantler (commented):

Static exact-head review of d8424e248e1303a2079d60452307159956401308. Ownership is unverified; the branch was not checked out or executed. One blocking finding follows.

k8s/providers/hetzner/infrastructure/controllers/cilium/cilium-clusterwide-network-policy.yaml
```diff
@@ -78,3 +78,12 @@ spec:
         - {}
       authentication:
         mode: required
+
+  # Deny pod egress to Hetzner IMDS. Autoscaler-created workers receive their
+  # Talos worker machine config through Hetzner user-data, and that config
+  # contains reusable cluster join material. Keep host/Talos boot-time metadata
+  # access intact, but prevent workloads from reading user-data after a node
+  # joins the cluster.
+  egressDeny:
```
devantler, line 87: [P1] Apply the IMDS deny to every endpoint

This egressDeny inherits the policy-level endpointSelector, which intentionally excludes every endpoint carrying k8s-app=kube-dns. The exception is cluster-wide rather than namespace-bound, so actual CoreDNS and any workload able to carry or spoof that label remain outside the metadata deny and can still reach 169.254.169.254 when their other egress policy permits it. Because that endpoint exposes reusable Talos join material, this defeats the stated security boundary. Put the IMDS deny in a separate prod-only CiliumClusterwideNetworkPolicy with endpointSelector: {} (leaving the mutual-auth CoreDNS carve-out on its original rule), and add negative coverage for both a normal pod and a kube-dns-labeled pod.

devantler:

Closing this draft. **The security concern is valid and has been re-filed** as
devantler-tech/platform#3153 — this is not a dismissal of the finding, only of this branch as the
vehicle for it.

### Why this branch cannot be resolved forward

The PR appends its `egressDeny` to
`k8s/providers/hetzner/infrastructure/controllers/cilium/cilium-clusterwide-network-policy.yaml`.
That file **no longer exists on `main`**, and it was **removed rather than moved**: `fdd9d82a`,
*"fix(cilium): remove broad mutual auth allow policy (#2724)"* (2026-08-03), deleted all 80 lines
as a deliberate security change. This PR was opened 2026-07-18, two weeks before that.

🔴 So the `DIRTY` state here is a trap, not a chore. Resolving the conflict toward this branch would
**restore an 80-line broad mutual-auth policy that was deliberately removed** — trading a new
control for the regression of an older one, in a single conflict resolution that looks routine.
`scripts/tests/test-cilium-mutual-auth-policy-regressions.sh` exists to guard that return path.

Recording it here so the next run that reaches this PR does not re-derive it under conflict-
resolution pressure.

### What carries forward

The intent is still unimplemented — verified against `main` on 2026-08-15, workload egress to the
metadata address is unconstrained (three searches, against a working control). #3153 carries that
forward with acceptance criteria written against the **current** layout, plus the constraint that
host/Talos boot-time access must keep working and the mutual-auth regression tests must still pass.

The docs paragraph in this PR is not carried over verbatim: it describes the mechanism through the
deleted file, so it would document something that no longer exists.

PR #2795, opened 2026-07-23
fix(ci): handle exception file close errors

devantler:

## Why

The shared Go validation workflow now rejects the generator's unchecked deferred file close. This fails every current platform PR before its own change can be evaluated.

## What

Propagate a close failure when document decoding otherwise succeeded, while preserving the earlier decode error when one exists.

## Validation

- `go test ./...` (98 tests)
- `golangci-lint run ./scripts/generate-kubescape-exceptions/...`
- generated the current 21-policy Kubescape exceptions file through the CLI

Unblocks #2794.

### DIFF
--- a/scripts/generate-kubescape-exceptions/main.go
+++ b/scripts/generate-kubescape-exceptions/main.go
@@ -449,14 +449,16 @@ func generate(directory string) ([]policy, error) {
 }
 
 // decodeDocuments reads every YAML document in a multi-document file.
-func decodeDocuments(path string) ([]any, error) {
+func decodeDocuments(path string) (documents []any, err error) {
 	file, err := os.Open(path)
 	if err != nil {
 		return nil, fmt.Errorf("open %s: %w", path, err)
 	}
-	defer file.Close()
-
-	var documents []any
+	defer func() {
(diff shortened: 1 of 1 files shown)


devantler:

CI follow-up at head `56e4678f356f1d4bbd8ba8fd10102af82c3568c6`:

- the previously failing `golangci-lint` gate is green
- all other change-relevant checks are green
- MegaLinter is red on the pre-existing repository-wide baseline because the repository configuration that makes it actionable remains in sibling-owned draft PR #2776

This hotfix stays draft and no review lane will be spent while that named dependency keeps CI red.

devantler:

Closing as superseded by merged #2776. That change resolved the same `golangci-lint` finding on `scripts/generate-kubescape-exceptions/main.go` by explicitly discarding the close result for this read-only input, with the rationale that a close error does not invalidate documents already decoded successfully.

The replacement head `d1dfd836cf5fc2f4a78a9183f3dee55efec50590` passed the repository test, `golangci-lint`, MegaLinter, manifest-validation, required-check, and security gates before merge, and current `main` still carries that disposition. This draft now conflicts exactly at the superseded implementation, so merging or rebasing it would re-open a settled behavior choice rather than add coverage.

Issue #2775, opened 2026-07-21
Lint gate is red on every pull request, and main never runs it


devantler:
> 🤖 Generated by the Daily AI Engineer

## Evidence

The `🧹 Lint - mega-linter` job fails on pull requests against `main` for defects that are already on `main`. Measured on #2769 — a pull request whose entire diff is one new `.coderabbit.yaml` file, which cannot influence Go or duplication linting ([run 29809153393](https://github.com/devantler-tech/platform/actions/runs/29809153393/job/88566175356)):

- **revive — 2 errors.** `scripts/validate-merge-group-heal/main.go:1:1` and `scripts/validate-eks-ci-role-policy/main.go:1:1`: `should have a package comment`. Both files are present on `main` today without one.
- **jscpd — 24 clones, 2.4% against a 0.0% threshold.** 16 of them are inside `scripts/validate-eks-ci-role-policy/main_test.go` (landed in #2753 this morning); 4 are inside `scripts/tests/test_refresh_flux_ghcr_auth.py` and `scripts/tests/test_validate_homepage_bookmarks.py`.
- **zizmor — 1 error**, `artipacked` on `cd.yaml` failing with `401 Unauthorized` while listing tags for `actions/checkout`. This one is a transient GitHub API condition, not a repository defect.

The reason nobody noticed is the second half of the problem: **the job does not run on `main` at all.** The most recent push-to-`main` Code Quality run ([29806818885](https://github.com/devantler-tech/platform/actions/runs/29806818885)) contains exactly one job, `Analyze (python)`. So `main` reports green while carrying lint defects, and every pull request inherits them.

## Who this affects, and how much

Every open pull request on this repository — currently 40+ drafts — displays a red Lint check that has nothing to do with its own diff. That makes the check unusable as a signal: a genuine lint regression introduced by a PR is indistinguishable from the inherited noise, so it will be waved through. It also costs a full 5-minute lint run per PR to re-report the same pre-existing failures.

## Expected behaviour

A pull request that introduces no lint defect shows a green Lint check, and a defect that reaches `main` is caught when it lands rather than surfacing on unrelated pull requests afterwards.

## Acceptance criteria

- [ ] revive reports 0 errors across the repository.
- [ ] jscpd reports 0 clones, with the threshold left at 0.0% — deduplicated at the source, never raised or ignored.
- [ ] The lint gate also runs on pushes to `main`, so a defect cannot land silently again.
[...]

PR #2769, opened 2026-07-21
ci(coderabbit): keep docstring coverage advisory instead of merge-blocking

devantler:
> 🤖 Generated by the Daily AI Engineer

## Why

CodeRabbit refuses to pass a pull request unless 80% of the functions it touches carry a doc comment — but it only counts a doc comment it can see inside the changed lines, so editing the body of an already-documented function reads as undocumented. The result is finished work blocked on a documentation gap that does not exist.

It happened on #2753 this morning: the check reported 68.75% and was the only thing failing, while every one of the 65 functions in the changed files was in fact documented.

This repository has 40+ open drafts waiting to be promoted, and each of them has to clear the same check.

## What

Turns that check into advice rather than a blocker — CodeRabbit still reports the number, it just stops holding merges. Nothing else about its behaviour changes. The same one-line change is already live in ksail, monorepo, .github and actions.

Part of devantler-tech/monorepo#2108

PR #2802, opened 2026-07-25
fix(ci): restore the EKS authorization gate and run it on main

### LINKED ISSUE #2801: bug(ci): EKS authorization validator fails on main, reddening every PR

## Evidence

`scripts/validate-eks-ci-role-policy` fails on `main`, and every open PR inherits the failure the
moment its CI re-runs. It is invisible on `main` itself because the job is PR-only — `main`'s own
check-runs are all green.

Bisected to a single commit, with the test genuinely executing in both directions:

| commit | result | unresolved-substitution lines |
|---|---|---|
| `6b67c945` (`fix(aws): allow EKS version discovery (#2753)`) | `--- PASS: TestValidateAuthorizationAcceptsCommittedPolicy` / `ok` | **0** |
| `f89efff4` (`fix(secret): update alertmanager webhook URL and metadata timestamps`, current `main`) | `--- FAIL` | **35** |

Run as `go test -count=1 ./scripts/validate-eks-ci-role-policy/...` on a clean worktree of each
commit. `-count=1` matters: the first attempt returned a cached `ok` and looked green.

The failure reports `unresolved Flux substitution in authorization resource` across the whole
authorization surface — HelmReleases in `kube-system`, `longhorn-system`, `openbao`, `kyverno`,
`kubescape`, `velero`, … plus `RoleBinding`/`ClusterRoleBinding` entries such as `oidc-view` and
`oidc-cluster-reader`. It is the entire surface, not one resource, which points at the substitution
*source* rather than at any individual manifest.

`f89efff4` changed exactly one file: `k8s/clusters/prod/bootstrap/secret.enc.yaml`, 4 insertions /
4 deletions.

## How it was found

`platform#2776` went red on four checks — `🧪 Test`, `📊 Code Coverage`, `🔐 Validate EKS
Authorization` and `CI - Required Checks` — after a **documentation-only** push. All four are the
same root cause: they all build on that one Go package.

The discriminator was that the branch **passes alone** and **fails merged**: CI checks out the merge
result, so a PR inherits `main`'s state even when its own diff is inert.

| tree | result |
|---|---|
| `claude/lint-gate-green-2775` alone | **ok** |
| `main` alone | **FAIL (35)** |
| branch merged with `main` | **FAIL** |

So this is not that PR's defect. Any PR whose CI re-runs will now go red the same way.

Every contributor and every open PR on this repository. The gate is a **security** control over the
EKS authorization surface, so a persistently red result also trains reviewers to discount it — and
because `main` never runs it, nothing signals that `main` is the thing that is broken.

## Expected behaviour

- `TestValidateAuthorizationAcceptsCommittedPolicy` passes on `main`.
- The authorization surface's Flux substitutions resolve, or the validator is taught the new
  substitution source if that source legitimately moved in `f89efff4`.
- **`main` runs this validator**, so a regression is caught where it is introduced instead of
  surfacing as an unexplained red on unrelated PRs. That gap is what let this sit unnoticed.

## Acceptance criteria

- [ ] `go test -count=1 ./scripts/validate-eks-ci-role-policy/...` passes on `main`.
- [ ] A previously-red open PR goes green on these four checks with no change to its own diff.
- [ ] The validator runs on pushes to `main`, not only on pull requests.
- [ ] A regression test pins whichever invariant `f89efff4` broke, so the next edit to the
      substitution source fails loudly and locally.

**Size:** small–medium once the substitution source is identified; the `main`-coverage gap is a
separate, small change and can ship first.

I have deliberately not reproduced any secret material here — only the commit, the changed path, and
the validator's own resource identifiers, all of which are already public in this repository.

devantler:

## Why

The EKS authorization gate — a security control over who can act on the production
cluster — has been failing on `main` since 23 July. Every open PR that re-ran CI
inherited the failure and went red for reasons that had nothing to do with its own
change, which both blocked real work and trains reviewers to wave the gate through.

It stayed invisible for two days because the gate never ran on `main` itself: it was
wired to pull requests, the merge queue, and a manual deploy. The change that broke it
was pushed straight to `main`, which fires none of those, so `main`'s own checks stayed
green while it was broken.

## What

Two things, both needed to close it out:

**Re-approves the authorization baseline**, which unblocks `main` and every affected PR.
The break was a routine secret rotation: the production bootstrap secret feeds
configuration values into the cluster, and the gate pins its encrypted contents, so
re-encrypting it moved the fingerprint. Before re-approving I measured that the change
carries no authorization meaning — one of 159 tracked items moved, its set of
configuration keys is unchanged, and no pinned permission resource moved.

**Runs the gate on pushes to `main`**, so this class of break is reported where it
happens instead of surfacing days later on someone else's PR. A guard test fails if that
coverage is ever removed again.

⚠️ **Worth knowing:** the gate is sensitive to the encrypted file's raw bytes, so any
routine secret rotation will red it again with no permission change at all — it even
absorbed a SOPS version bump here. I've documented that in place and filed the
recurrence separately rather than widening the control silently in this PR.

Also fixes an unrelated unchecked-error lint failure in another script. It has been
red on `main` since a linter version bump and fails the lint job on *any* PR that
touches Go, so it had to go green here.

Fixes #2801

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -68,6 +68,10 @@ jobs:
               # cd.yaml carries the manual-deploy gate itself. Without it here, a PR
               # that deletes that gate does not run the job whose test guards it.
               - '.github/workflows/cd.yaml'
+              # Same reasoning for validate-main.yaml, which carries the
+              # push-to-main gate: TestAuthorizationGateRunsOnPushToMain guards it,
+              # and that test only runs when this filter matches.
+              - '.github/workflows/validate-main.yaml'
             bridge_validation:
               - 'scripts/refresh-flux-ghcr-auth.sh'
               - 'scripts/ghcr-auth-lib.sh'
--- a/.github/workflows/validate-main.yaml
+++ b/.github/workflows/validate-main.yaml
@@ -1,0 +1,64 @@
+name: Validate Main
+
+# ci.yaml runs the authorization gate on `pull_request` and `merge_group` only,
+# and cd.yaml runs it on a manual dispatch. A DIRECT PUSH to main fires none of
+# those, so main could carry a broken authorization surface with every one of
+# its own checks green — which is exactly what happened in f89efff4: the break
+# stayed invisible on main for two days and surfaced only as an unexplained red
+# on unrelated PRs, because a PR's CI builds the merge result and so inherits
+# main's state.
+#
+# This workflow closes that hole: the same gate, on the push itself, so a
+# regression is reported where it is introduced.
+#
+# Deliberately NOT path-filtered. The gate reads the whole rendered
+# authorization surface, which a change anywhere in the overlays (or in the
+# validator itself) can move; a paths filter here would recreate a narrower
+# version of the same blind spot.
+on:
+  push:
+    branches: [main]
+
+permissions: {}
+
+concurrency:
+  group: validate-main-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  validate-eks-authorization:
+    name: 🔐 Validate EKS Authorization
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # checkout repository
+    steps:
+      - name: 📑 Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+
+      - name: ⚙️ Setup Go
+        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      - name: ⚙️ Install pinned kubectl renderer
+        shell: bash
+        env:
+          KUBECTL_VERSION: "v1.36.2"
+          KUBECTL_SHA256: "1e9045ec32bea85da43de85f0065358529ea7c7a152eca78154fba5b58c27d82"
+        run: |
+          set -euo pipefail
+          kubectl_path="$RUNNER_TEMP/kubectl"
+          curl -fsSL \
+            "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl" \
+            -o "$kubectl_path"
+          printf '%s  %s\n' "$KUBECTL_SHA256" "$kubectl_path" | sha256sum --check
+          sudo install "$kubectl_path" /usr/local/bin/kubectl
+          kubectl version --client
+
+      - name: 🔐 Validate EKS authorization surface
+        run: |
+          go test ./scripts/validate-eks-ci-role-policy
+          go run ./scripts/validate-eks-ci-role-policy .
--- a/scripts/generate-kubescape-exceptions/main.go
+++ b/scripts/generate-kubescape-exceptions/main.go
@@ -454,8 +454,9 @@ func decodeDocuments(path string) ([]any, error) {
 	if err != nil {
 		return nil, fmt.Errorf("open %s: %w", path, err)
 	}
-	// Read-only handle: a failed Close cannot lose data, and the caller already
-	// has the decode result. Discard explicitly so errcheck sees the intent.
+	// The file is only ever read, so a close error says nothing about whether
+	// the documents below were decoded correctly — discard it explicitly rather
+	// than leaving it unchecked.
 	defer func() { _ = file.Close() }()
 
 	var documents []any
--- a/scripts/generate-kubescape-exceptions/main.go
+++ b/scripts/generate-kubescape-exceptions/main.go
@@ -454,7 +454,9 @@ func decodeDocuments(path string) ([]any, error) {
 	if err != nil {
 		return nil, fmt.Errorf("open %s: %w", path, err)
 	}
-	defer file.Close()
+	// Read-only handle: a failed Close cannot lose data, and the caller already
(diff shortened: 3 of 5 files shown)


devantler:

## Evidence and user evaluation

**Root cause, measured rather than inferred.** The failure printed one aggregate error plus
35 `unresolved Flux substitution` lines. Those 35 are a red herring — they are emitted on any
aggregate mismatch and are present verbatim on the *passing* parent commit. To find the real
mover I dumped per-entry surface fingerprints (identity + hash only; one selected document is a
secret, so no content was rendered) at both commits and diffed:

| | `6b67c945` (last good) | `f89efff4` (main) |
|---|---|---|
| surface entries | 159 | 159 |
| differing entries | — | **1** |

The single mover is `v1/Secret flux-system/variables-cluster`, a `substituteFrom` source, and it
is unpinned — which is exactly why no per-resource error fired and only the aggregate moved.

**Why re-approval is safe.** The commit's plaintext key set is byte-identical before and after
(SOPS leaves keys in clear, so this needed no decryption), and zero pinned authorization
resources moved. No substitution variable was added, removed or repointed.

**Renderer calibration — the step that makes the new hash trustworthy.** A locally computed hash
is only valid if my renderer agrees with CI's. Checked out `6b67c945` and ran the suite: it
reproduced that commit's *committed* approved hash exactly, on kubectl v1.36.3 vs CI's pinned
v1.36.2. So the value here is the one CI will compute.

**RED/GREEN on the new coverage guard.** With `validate-main.yaml` removed,
`TestAuthorizationGateRunsOnPushToMain` fails with the actionable message; restored, it passes.
Three negative controls also pass: a push trigger on another branch, a `pull_request` trigger on
main, and a step that merely names the validator without running it — each must fail to count.

**Tried as a user.** `🔐 Validate EKS Authorization` is **green on this PR in CI** (job
89705042743) — the gate that has been red since 23 July. `go test ./scripts/...` is green across
all 3 packages, twice in a row for determinism, and `gofmt`/`go vet` are clean.

**One residual I deliberately did not fix here:** the gate is sensitive to raw ciphertext, so the
next routine secret rotation will red it again with no authorization change. Narrowing that is a
real security-control decision, so it is filed as #2803 with the options and the per-control
argument each would need, rather than quietly widened inside this PR.

devantler:

devantler:



Requesting review at `d7a55af`. 23 of 25 checks are green, including the gate this PR
restores. The one failure, `🧹 Lint - mega-linter`, is a **repo-wide baseline unrelated to this
diff** — `.mega-linter.yml` is absent on `main`, so the whole tree is linted with defaults; it
fails identically on #2769, #2790 and #2797, and passes only on #2776, the PR that adds the config.
It is not a required check.

Worth attacking specifically: whether re-approving the surface fingerprint is justified by the
measurement recorded in the code comment and the evidence comment, and whether the new coverage
guard is genuinely non-vacuous.

@coderabbitai review

devantler:

## Superseded by #2776 — same lane, same fixes, one merge

This PR and #2776 were deadlocking. Every pull request on this repository fails
`🔐 Validate EKS Authorization` (this PR's fix), and every lintable pull request fails
`🧹 Lint - mega-linter` (#2776's fix) — so neither could go green alone, and nothing has
merged here since 21 July.

Rather than duplicate either change, I merged this branch into #2776. That PR now carries both
and closes #2801 as well as #2775/#2780.

Before merging it I re-verified this PR's security-sensitive half independently rather than
trusting the original measurement: the rotated secret's plaintext key set is unchanged (22 keys,
byte-identical), exactly one of 159 surface entries moved (that Secret's own ciphertext), and the
value that changed has no authorization consumer — the validator flags unresolved substitutions
rather than resolving them, so consumers render the literal `${…}` either way. The new coverage
guard is RED/GREEN proved: removing `validate-main.yaml` fails
`TestAuthorizationGateRunsOnPushToMain`, restoring it passes.

Leaving this open until #2776 merges, so nothing is lost if that PR needs to be unwound.

devantler:

Closing as superseded — this change is already on `main`.

The EKS authorization gate and the lint fix were mutually blocking: each PR was red on the check the *other* one fixed, so neither could go green alone. The resolution was to merge this branch into #2776 so one PR carried both, and #2776 landed at `19:51Z`.

Verified before closing, file by file, that `main` now contains everything this branch adds:

- `.github/workflows/validate-main.yaml` — identical blob on `main`
- `scripts/validate-eks-ci-role-policy/coverage_test.go` — identical blob on `main`
- `scripts/generate-kubescape-exceptions/main.go` — identical blob on `main`
- `scripts/validate-eks-ci-role-policy/main.go` and `.github/workflows/ci.yaml` — `main` is strictly *ahead* of this branch (it carries a package doc comment and a formatting fix this branch predates)

So the branch has no unique content left; the tree-to-tree diff against `main` is subtractive only. `#2801` is already closed. Nothing here needs re-landing.

devantler:

Closing: superseded by #2776, which **merged at 19:51:16Z** (`c76f5a63a4`) carrying this branch's
commits and closing #2801 alongside #2775/#2780.

I kept this open until that landed so nothing would be lost if it had to be unwound. It did not.

Confirmation the unwedge is real rather than assumed: after four days with **zero** merges on this
repository, three landed in seventeen minutes — this one, then #2790 and #2798 draining on their own
through the queue.

The authorization fix from this branch is now on `main`: the surface fingerprint is re-approved after
the secret rotation, `validate-main.yaml` runs the gate on pushes to `main`, and
`TestAuthorizationGateRunsOnPushToMain` fails loudly if that coverage is ever removed again.

PR #2819, opened 2026-07-26
feat(apps): onboard the doggy-countdown tenant

devantler:

## Why

A small personal app — a countdown to the day a puppy comes home — needs somewhere to run. It is a single static page, so it should cost the platform as little as possible to host.

## What

Onboards it as an ordinary tenant at **https://simba.platform.devantler.tech**.

It reuses what the platform already provides rather than adding anything: the shared gateway's certificate already covers that whole domain, and DNS is created from the app's own route. So there is no certificate, no DNS resource, and no per-app ingress plumbing to maintain.

**It holds no credentials at all.** Both its image and its manifests are public, so it pulls anonymously — no pull secret, and no dependency on the secret store. That is one fewer thing to rotate and one fewer thing to break.

The only permission it gets is the standard one every tenant gets: its own service account, limited to its own namespace.

## Also in here

Re-approves the authorization fingerprint, which moves whenever a tenant is added. The measurement is recorded next to the value: the render grows by exactly this tenant's six resources with **nothing existing changed**, and the guard was RED/GREEN proved — it still fails if this tenant's permissions are widened.

### DIFF
--- a/k8s/bases/apps/doggy-countdown/flux-kustomization.yaml
+++ b/k8s/bases/apps/doggy-countdown/flux-kustomization.yaml
@@ -1,0 +1,23 @@
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1
+kind: Kustomization
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: ksail
+  annotations:
+    ksail.devantler.tech/reconcile-exclude: "true"
+  name: doggy-countdown
+  namespace: doggy-countdown
+spec:
+  interval: 1m
+  timeout: 5m
+  retryInterval: 2m
+  path: .
+  prune: true
+  wait: true
+  force: true
+  serviceAccountName: doggy-countdown
+  sourceRef:
+    kind: OCIRepository
+    name: doggy-countdown
+  targetNamespace: doggy-countdown
--- a/k8s/bases/apps/doggy-countdown/kustomization.yaml
+++ b/k8s/bases/apps/doggy-countdown/kustomization.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+# Matches the other tenant skeletons: when the KRO Tenant-CR handover
+# (#1932/#2486) de-inventories these objects from the `apps` Kustomization, Flux
+# must not garbage-collect them — kro adopts them in place.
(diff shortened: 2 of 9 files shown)


devantler:


23 checks green, mega-linter still running. Requesting now so the review and the last linter overlap.

@coderabbitai review

devantler:



CodeRabbit reports `Review limit reached` with **no retry window stated** — the plan's included reviews are exhausted, so the lane cannot deliver at this head. Recording that as a completed no-gate outcome and advancing to lane 2.

CI is green at `d7789280b4`: 24 pass / 0 fail.

@codex review

devantler:

## Readiness at `d7789280b4`

**Programmatically tested.** `ksail workload validate` on both overlays — 549 files, exit 0 (checked as the real exit, not a pipeline's). Every changed directory builds under `kubectl kustomize`. The authorization guard was RED/GREEN proved with the cache disabled: green at the re-approved value, **red when this tenant's roleRef is swapped to `cluster-admin`**, green on restore — so the gate still catches a privilege widening in the resource this PR adds. CI: 24 pass / 0 fail.

**Reviewed.** CodeRabbit reported `Review limit reached` with no retry window (included reviews exhausted) — recorded as a completed no-gate outcome. Codex reviewed at this exact head and found no issues.

**Exercised, not just reasoned about.** The parts that can be checked before merge, were:

- The published artifact was pulled from GHCR and its contents inspected: all six manifests, the right hostname, and the image **digest-pinned** by the publish workflow to `sha256:6a4285ef…`, which matches the published `1.0.1` image.
- Its cosign certificate SAN was read and matched against the OCIRepository matcher this PR ships — it verifies.
- Both packages confirmed **public** by anonymous pull, which is what justifies removing the pull secret and the ExternalSecret.
- The tenant ServiceAccount's permissions were checked against the live cluster with `auth can-i`: it can create the HTTPRoute, CiliumNetworkPolicy, Deployment and Service its artifact contains.
- The container image was built and run locally: page 200, image 200, removed gallery paths 404.

**What is only observable after merge**, and will be checked there: the namespace and source reconciling, the artifact being pulled and verified in-cluster, pods becoming ready, external-dns creating `simba.platform.devantler.tech`, and the page answering over HTTPS. `simba.platform.devantler.tech` correctly does not resolve yet.

Promoting on that basis.

PR #2821, opened 2026-07-27
security: pin the tenant RGD cosign subject to actions and a fixed ref

### LINKED ISSUE #2817: security: re-approve the EKS authorization surface so #2402 can land, then tighten the tenant RGD matcher

## Evidence

#2402 tightened the cosign subject matchers on the four live `OCIRepository` trust rules from `@.+` to `@([0-9a-f]{40}|refs/tags/v.+)$` and dropped the dead `(reusable-workflows|actions)` alternation. The **tenant `ResourceGraphDefinition`** carries a fifth copy of the same matcher and was deliberately left out of that PR:

`k8s/bases/infrastructure/resource-graph-definitions/tenant/resource-graph-definition.yaml`
`subject: '^https://github\.com/devantler-tech/(reusable-workflows|actions)/\.github/workflows/publish-app\.yaml@.+$'`

This is the template every new tenant is generated from, so while it is loose, **every tenant created from now on inherits the weak rule** and the #2402 fix does not stick.

## Why it was split out

Including it turned `🔐 Validate EKS Authorization` red — measured on PR #2816, run 30202725015. The RGD sits inside the rendered authorization surface, so editing it moves the aggregate fingerprint `expectedRenderedSurfaceSHA` pins.

That gate is working as designed, and `main.go` is explicit about how to clear it:

> Do NOT treat a moved hash as self-evidently benign: re-run the per-entry membership measurement above before re-approving.

The measurement is a render of both trees and an entry-by-entry diff (the current approved value documents 159 entries with exactly one moving). **The agent host has no Go toolchain installed**, so that measurement could not be run locally, and pasting the new hash out of the CI failure would be precisely the blind re-approval the comment forbids on a security gate.

Diagnostic detail worth keeping: the failing run reported the aggregate mismatch **only** — no `unapproved rendered <identity> fingerprint`, no `missing rendered authorization resource`, no `duplicate`. So no individually pinned grant-bearing object moved. That is suggestive but is *not* the required per-entry measurement, because the aggregate covers entries beyond the individually pinned set.

## Smallest useful change

1. Run the per-entry surface measurement on a machine with Go (or via a CI job that prints the entry list for both trees).
2. Confirm the only moving entry is the tenant RGD and that its delta is the cosign subject line alone.
3. Apply the same tightening as #2402, update `expectedRenderedSurfaceSHA`, and record the measurement in the constant's comment in the established house style (entry counts, which entry moved, old → new fingerprint, and why it grants nothing new).

## Acceptance criteria

- [ ] The tenant RGD's matcher rejects a floating ref and accepts the SHA and release-tag forms, matching the four rules #2402 already tightened.
- [ ] `expectedRenderedSurfaceSHA` is updated **with** a recorded per-entry measurement, not from the CI error text.
- [ ] The constant's comment names the entry count, the moved entry, and why no grant-bearing object is added, removed or repointed.
- [ ] `🔐 Validate EKS Authorization` is green.

## Notes

Sibling follow-up #2815 covers the same weakness at the container-image layer. The host's missing Go toolchain is worth fixing on its own account — it blocks running any of this repo's Go guards locally.

Part of #2402. Rough size: S, once the measurement can be run.

devantler:

## Why

The tenant `ResourceGraphDefinition` is the template every new platform tenant is generated from, and its container-signature rule was still the old, loose one: it accepted the archived `reusable-workflows` repo as a valid signer, and it accepted an artifact signed from *any* ref — including a floating branch. The four live trust rules were tightened weeks ago; this copy was deliberately left behind because tightening it trips the authorization gate, which must be re-approved by hand.

So every tenant created since then has inherited the weak rule, and the earlier fix does not stick.

## What

Brings the template's signer rule into line with the four live rules: only the shared `actions` workflow, and only when signed from a fixed commit or a release tag. Strictly a tightening — nothing the new rule accepts was rejected before.

Re-approves the two security fingerprints this necessarily moves, from a full measurement rather than by pasting the failing value: the rendered surface is **identical in membership** (509 documents both sides, zero added, removed or renamed) and **exactly one entry's content moved** — this RGD — whose rendered delta is the single `subject:` line. No permission-granting object changes anywhere in the surface.

Unblocks `devantler-tech/gitops-tenant-template#117`, whose contract test currently cannot pass in either direction because this file and the live rules disagree.

Fixes #2817

### DIFF
--- a/k8s/bases/infrastructure/resource-graph-definitions/tenant/resource-graph-definition.yaml
+++ b/k8s/bases/infrastructure/resource-graph-definitions/tenant/resource-graph-definition.yaml
@@ -249,15 +249,17 @@ spec:
           verify:
             provider: cosign
             matchOIDCIdentity:
-              # publish-app.yaml is moving from devantler-tech/reusable-workflows into
-              # devantler-tech/actions (actions#425 consumer repoint); both repo
-              # subjects are accepted via ONE regex-alternated entry — cosign's keyless
-              # attestation verification rejects MULTIPLE identity entries outright
-              # ("unsupported: multiple identities are not supported at this time").
-              # Narrow the alternation to actions-only once the first actions-signed
-              # artifact has been published and verified.
+              # Keyless-signed by the shared publish-app workflow this tenant's
+              # cd.yaml calls. The ref part accepts only a 40-hex commit SHA or a
+              # release tag, so an artifact signed from a floating ref (e.g. @main)
+              # does not verify. Consumers pin the workflow by SHA, so only the SHA
+              # branch is exercised today; the tag branch is kept so a tag-pinned
+              # caller also verifies. cosign's keyless attestation verification
+              # rejects MULTIPLE identity entries outright ("unsupported: multiple
+              # identities are not supported at this time"), so any future
+              # alternation must stay inside one subject regex.
               - issuer: '^https://token\.actions\.githubusercontent\.com$'
-                subject: '^https://github\.com/devantler-tech/(reusable-workflows|actions)/\.github/workflows/publish-app\.yaml@.+$'
+                subject: '^https://github\.com/devantler-tech/actions/\.github/workflows/publish-app\.yaml@([0-9a-f]{40}|refs/tags/v.+)$'
     # --- Flux Kustomization (one of two variants by `sops`) ---------------
     # The impersonating Kustomization that reconciles the tenant's own artifact
     # under its namespaced SA (bound to the `tenant-edit` role). Two variants because the SOPS `decryption`
--- a/scripts/validate-eks-ci-role-policy/main.go
+++ b/scripts/validate-eks-ci-role-policy/main.go
@@ -56,23 +56,30 @@ const (
 // The approved surface includes the encrypted flux-system/variables-cluster
 // substitution source and the staged Cilium homogeneous-device activation.
 //
-// Measured against main e77fdb9c before approving this value: the render goes
-// from 503 to 509 documents and the delta is **purely additive** — 97 added
-// lines, ZERO removed — so no existing entry is modified, removed or renamed.
-// The six new documents are one tenant skeleton (`doggy-countdown`):
-// Namespace, NetworkPolicy, ServiceAccount, RoleBinding, OCIRepository and
-// Flux Kustomization. The validator reported no per-resource
-// fingerprint mismatch, no missing and no duplicate resource.
+// Measured against main 1b204ded before approving this value: 509 documents on
+// both sides, with membership IDENTICAL — zero added, zero removed, zero
+// renamed. Exactly ONE entry's content moved, the tenant
+// `ResourceGraphDefinition`, and its rendered delta is a single line: the
+// cosign `subject:` matcher narrows from
+// `(reusable-workflows|actions)/…@.+` to `actions/…@([0-9a-f]{40}|refs/tags/v.+)`.
+// That drops the archived `reusable-workflows` repo as an accepted signer and
+// stops a floating ref (e.g. `@main`) from satisfying the rule, bringing the
+// template every tenant is generated from into line with the four live
+// OCIRepository trust rules. It is strictly a tightening: every subject the new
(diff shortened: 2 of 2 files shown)


devantler:

## Per-entry membership measurement (the re-approval evidence)

`main.go` requires this before re-approving a moved hash, and #2817's stated blocker was that no Go toolchain was available to run it. **That blocker is stale** — Go 1.26.5 is installed on the agent host (symlinked 2026-07-26 17:29Z, ~4.5h after #2817 was filed), so the measurement ran in full.

**Renderer:** the CI-pinned `kubectl v1.36.2` (kustomize v5.8.1), downloaded and SHA256-verified against the published checksum.

**Control first:** the validator passes on unmodified `main` (`1b204ded`) with this toolchain, which proves the local renderer reproduces CI's canonical hashes. Locally computed values are therefore authoritative, not guesses at what CI would say.

### Membership — identical

| | base `1b204ded` | head |
|---|---|---|
| documents | 509 | 509 |
| added | — | 0 |
| removed | — | 0 |
| renamed | — | 0 |

### Content — exactly one entry moved

```
kro.run/v1alpha1|ResourceGraphDefinition|-|tenant.kro.run
```

and its rendered delta is a single line:

```diff
-subject: ^https://github\.com/devantler-tech/(reusable-workflows|actions)/\.github/workflows/publish-app\.yaml@.+$
+subject: ^https://github\.com/devantler-tech/actions/\.github/workflows/publish-app\.yaml@([0-9a-f]{40}|refs/tags/v.+)$
```

No `Role`, `ClusterRole`, `RoleBinding`, `ClusterRoleBinding` or `ServiceAccount` is added, removed or modified anywhere in the surface, and nothing granted to the `aws/aws` service account this validator protects is touched. The change is monotonic: every subject the new matcher accepts, the old one already accepted.

Two constants were re-approved from that one measurement — the individually pinned per-resource fingerprint for the RGD, and the aggregate `expectedRenderedSurfaceSHA`.

### RED/GREEN

- **GREEN** — `go test ./scripts/validate-eks-ci-role-policy` passes; `go run ./scripts/validate-eks-ci-role-policy .` reports the contract passed.
- **RED** — re-loosening the ref part back to `@.+$` is caught by **both** guards, so neither re-approval is vacuous:
  ```
  unapproved rendered {…name:tenant.kro.run} fingerprint: 4f2c9654…
  unapproved rendered authorization surface fingerprint: e0e0bf64…
  ```
  Source restored and GREEN re-confirmed afterwards.

The `unresolved Flux substitution` lines in a failing run are diagnostic, not findings — `main.go` documents that they are emitted only alongside a mismatch to explain a hash that moved.

devantler:



@coderabbitai review

devantler:

## Readiness

- **Programmatically tested** — `go test ./scripts/validate-eks-ci-role-policy` ok; `go vet` clean; the validator itself reports the contract passed. RED/GREEN in the evidence comment above: re-loosening the ref part is caught by **both** the per-resource pin and the aggregate, so neither re-approval is vacuous. All 24 required checks green.
- **Reviewed** — CodeRabbit green at `5087442f`, zero inline comments and no actionable finding section.
- **Tried and evaluated as a user** — the exercisable surface here *is* the validator, and it was run directly against both trees with the CI-pinned renderer: it passes on unmodified `main` (proving my renderer reproduces CI's canonical hashes, so the re-approved values are measured rather than guessed) and passes on this branch.

## What this becomes free to enact, checked before merging

The RGD is a template, so the question is whether tightening it can break a live verification. It cannot:

- **No live tenant is generated from it.** The docker overlay's `resources:` is empty, the one `Tenant` CR is documented opt-in and unreferenced, and the 509-document production render contains **zero** `kind: Tenant` instances.
- **The three live `publish-app` tenants** (`wedding-app`, `ascoachingogvaner`, `doggy-countdown`) carry hand-written `OCIRepository` rules that already hold the byte-identical tight matcher on `main` — so real artifacts are demonstrably signed in a way this matcher accepts.
- The archived `reusable-workflows` repo no longer signs anything.

Promoting and merging.

PR #2826, opened 2026-07-27
chore(renovate): disable the Dependency Dashboard

devantler:

## Why

You asked for this directly, twice, on monorepo#2503 (16:44Z and 17:02Z today):

> I would also like us to disable this issue dashboard. I do not use it, and it confuses agents, so it should be disabled!

`platform` is the last repo still opting in. The monorepo already turned it off; this repo's dashboard is the issue that has been sitting open since August 2023, and it does confuse agents — it looks like the oldest open issue in the portfolio, so runs keep having to recognise and skip it.

## What

Turns the dashboard off in this repo's Renovate config, which is what makes Renovate retire the issue itself. Deliberately not done by closing the issue by hand: that issue belongs to the bot, and closing it manually changes Renovate's behaviour in a way nobody chose.

Nothing depends on it. No update in this repo waits for dashboard approval — every rule that mentions the dashboard says exactly that — so no approval step is lost.

Fixes nothing tracked; this implements your direction on monorepo#2503.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": ["config:recommended"],
-  "dependencyDashboard": true,
+  "dependencyDashboard": false,
   "assigneesFromCodeOwners": true,
   "minimumReleaseAge": "7 days",
   "flux": {


devantler:



One-line config change; CI green at this head (27/27, zero failures).

@coderabbitai review

devantler:



CodeRabbit's rate-limit window from the previous request has elapsed. Re-requesting at the same head — one-line config change, CI 27/27 green, no findings outstanding.

@coderabbitai review

devantler:

## Readiness record

**1. Programmatically tested** — 27/27 checks green at `e08ac969a5`, zero non-green; `mergeStateStatus` CLEAN; zero review threads; zero review-body findings.

**2. Reviewed** — CodeRabbit green at this head: its summary comment was updated at 21:40:43Z, after the 21:39:26Z request, names the head, carries no rate-limit or service marker, and reports no actionable comments. (The two earlier requests hit the account review limit; that window elapsed before this one.)

**3. Tried and evaluated as a user** — this is Renovate configuration consumed by an external service, so there is no local runtime to exercise. I validated the changed artifact itself against the schema the file declares (`docs.renovatebot.com/renovate-schema.json`):

- The document parses, and **every top-level key is known to the schema**.
- Dereferencing the `$ref`, `dependencyDashboard` is declared `"type": "boolean"` with **`"default": false`** — so the new value is valid, and it is also upstream's own default. This change moves the repo *onto* the Renovate default rather than away from it.
- **Negative control:** the same check applied to a deliberately corrupted copy (`"dependencyDashboard": "nope"`) reports MISMATCH, so the positive result is not vacuous.
- Nothing depends on the dashboard: `dependencyDashboardApproval` occurs **0 times** in the config and `dependencyDashboard` has **no** other reference anywhere in the repository, so no package rule is gated on dashboard approval and nothing is left waiting for a checkbox that will no longer exist.

An honest limit worth stating: the retirement of the existing dashboard issue (#313) is performed by Renovate on its next run, which I cannot trigger. That is the one part of the effect observable only after merge.

My first pass at the schema check was wrong and reported MISMATCH for all 14 keys — every property is a `$ref`, so `.type` is null at the property level and the comparison was vacuous. Recorded because the corrected check is the one the conclusions above rest on.

PR #2833, opened 2026-07-28
fix(security): enforce the tenant ClusterIssuer boundary at admission

### LINKED ISSUE #2470: Enforce the tenant ClusterIssuer boundary with an admission policy

**Problem.** `cert-manager-tenant-edit` (PR #2453) scopes tenants to namespaced Issuers, but RBAC cannot stop a tenant Certificate from *referencing* a cluster-scoped ClusterIssuer via `spec.issuerRef.kind: ClusterIssuer` — the reference is resolved by cert-manager's controller, not the tenant's identity. The stated boundary ("ClusterIssuer remains platform territory") is currently convention, not enforcement. Raised by review on #2453.

**Proposed direction.** Runtime-enforce per the fix-vs-except ladder: a Kyverno validating ClusterPolicy on Certificate (and CertificateRequest) objects in tenant namespaces rejecting `spec.issuerRef.kind != Issuer` — Kyverno is already the platform's admission layer, so prefer it over adding cert-manager approver-policy as a new controller. Start in Audit, graduate to Enforce.

**Rough size.** Small-medium: one ClusterPolicy + tenant-namespace selector convention + validation.

devantler:

## Why

Tenants can manage their own certificates, and the intended boundary is that
cluster-wide issuers stay platform territory. Today that boundary is a comment,
not a rule. Permissions can stop a tenant from *editing* the shared issuer, but
not from *pointing at* it — that reference is resolved by cert-manager on the
tenant's behalf, under cert-manager's own identity.

The consequence is that any tenant could request certificates through the
platform's shared Let's Encrypt account, including for hostnames belonging to
other tenants. Nothing would reject it and nothing would flag it.

## What

Adds an admission rule that requires certificates in tenant namespaces to use an
issuer from their own namespace. Platform namespaces are untouched, and so is
anything the platform itself installs into a tenant namespace.

**Security floor gained:** a tenant can no longer obtain a certificate for a name
it does not own via the shared issuer. **Everyday cost:** none — the supported
path already works and is what the one existing tenant certificate does today.

Two details worth flagging, both of which would have been silent bugs:

- It allows an approved list of issuer types rather than blocking one by name.
  Add-on issuer projects ship both a namespaced and a cluster-wide variant, so
  blocking only the obvious name would have missed every other cluster-wide type.
- Leaving the issuer type unset is legal and means "my own namespace". Two of the
  nine certificates running in production do exactly that, so treating unset as a
  violation would have been an outage.

Checked against production before choosing to enforce rather than warn: there are
no violating certificates today, so nothing currently running is affected.

Fixes #2470

### DIFF
--- a/k8s/bases/infrastructure/cluster-policies/best-practices/restrict-tenant-issuer-refs.yaml
+++ b/k8s/bases/infrastructure/cluster-policies/best-practices/restrict-tenant-issuer-refs.yaml
@@ -1,0 +1,145 @@
+# Confines self-service tenants to their OWN namespaced Issuer and blocks any
+# reference to a cluster-scoped issuer from a tenant namespace.
+#
+# Why: tenants are granted edit on Certificate/Issuer (see
+# cluster-roles/cert-manager-tenant-edit.yaml) so they can manage their own
+# certificates. Referencing a ClusterIssuer needs no RBAC on the issuer object —
+# `spec.issuerRef` is resolved by cert-manager's controller under ITS identity,
+# not the tenant's — so RBAC cannot enforce the stated boundary ("ClusterIssuer
+# remains platform territory"). Without this guard a tenant could point a
+# Certificate at the shared `letsencrypt-prod` ClusterIssuer and mint
+# certificates for ANY domain the platform's ACME account controls, including
+# other tenants' hostnames. A tenant must instead create its own namespaced
+# Issuer, which is scoped to credentials the platform grants it.
+#
+# Scope: only ksail-generated tenant namespaces (labelled
+# app.kubernetes.io/managed-by: ksail), so platform/infra namespaces that
+# legitimately use a ClusterIssuer (kube-system's gateway-certificate,
+# cert-manager, cnpg-system, …) are unaffected.
+#
+# Allow-list, not deny-list: the rules deny any issuerRef.kind that is not in
+# the allowed set, rather than denying `ClusterIssuer` by name. External issuer
+# projects conventionally ship BOTH a namespaced and a cluster-scoped kind
+# (AWSPCAIssuer/AWSPCAClusterIssuer, step-issuer, google-cas, …), so a
+# name-based deny silently misses every cluster-scoped kind except
+# cert-manager's own. Extending the boundary later means adding the namespaced
+# kind to `value:` below — a visible, reviewable edit.
+#
+# `kind` is OPTIONAL and defaults to `Issuer` when omitted, so both rules
+# default the empty case to `Issuer` before comparing. This is not theoretical:
+# 2 of the 9 Certificates live on prod omit `kind` entirely. A naive
+# `NotEquals Issuer` on the raw field would deny every one of them.
+#
+# Platform carve-out: the deny targets only resources applied by a TENANT — i.e.
+# by the impersonated tenant ServiceAccount (the tenant's deploy/ Kustomization
+# sets spec.serviceAccountName: <tenant>). Resources the PLATFORM ships into a
+# tenant namespace are reconciled by the apps-layer Flux Kustomization, which
+# sets NO serviceAccountName and therefore applies as flux-system's own
+# kustomize-controller SA — excluded below. The discriminator is the applying
+# identity, which a tenant cannot forge. This mirrors
+# restrict-tenant-secret-stores.yaml so both tenant boundaries share one rule.
+#
+# CertificateRequests are additionally excluded for cert-manager's own SA:
+# cert-manager MINTS a CertificateRequest from each Certificate under its own
+# identity, copying the issuerRef verbatim. Without that exclusion, a
+# platform-authored ClusterIssuer-backed Certificate in a tenant namespace would
+# be admitted (flux-system carve-out) and then have its derived
+# CertificateRequest rejected — breaking issuance in a way that points at the
+# wrong object. Tenants cannot create CertificateRequests directly (the
+# ClusterRole grants read only), so this rule is defense-in-depth that holds if
+# that RBAC ever regresses.
+#
+# Because both excludes match on request.userInfo, the policy is admission-only
+# (background: false); every escalation is a create/update that passes through
+# admission, so enforcement is unaffected.
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: restrict-tenant-issuer-refs
+  annotations:
+    policies.kyverno.io/title: Restrict Tenant Issuer References
+    policies.kyverno.io/category: Security, cert-manager
+    policies.kyverno.io/severity: high
+    policies.kyverno.io/subject: Certificate, CertificateRequest
+    policies.kyverno.io/minversion: 1.6.0
+    policies.kyverno.io/description: >-
+      Blocks tenant-applied Certificate and CertificateRequest resources in
+      ksail tenant namespaces from referencing a cluster-scoped issuer via
+      spec.issuerRef.kind. Tenants must use their own namespaced Issuer, so a
+      tenant cannot mint certificates through the platform's shared ACME
+      account. Platform-authored resources (applied by flux-system's
+      kustomize-controller) and cert-manager's own derived CertificateRequests
+      are excluded.
+spec:
+  validationFailureAction: Enforce
+  # Admission-only: the excludes match on request.userInfo (the applying
+  # ServiceAccount), which is unavailable during background scans. Every
+  # escalation path is a create/update that passes through admission, so
+  # enforcement is fully preserved.
+  background: false
+  rules:
+    - name: certificate-no-cluster-scoped-issuer
+      match:
+        any:
+          - resources:
+              kinds:
+                - Certificate
+              namespaceSelector:
+                matchLabels:
+                  app.kubernetes.io/managed-by: ksail
+      # Exclude platform-authored resources: the tenant registration dir is
+      # reconciled by the apps-layer Flux Kustomization as flux-system's own
+      # kustomize-controller SA (no impersonation). A tenant's deploy/ artifact
+      # is applied as its impersonated tenant SA, so it stays subject to the deny.
+      exclude:
+        any:
+          - subjects:
+              - kind: ServiceAccount
+                name: kustomize-controller
+                namespace: flux-system
+      validate:
+        message: >-
+          Tenant Certificates must reference a namespaced Issuer
+          (kind: Issuer), not a cluster-scoped issuer such as ClusterIssuer.
+          Create an Issuer in this namespace and point spec.issuerRef at it.
+        deny:
+          conditions:
+            any:
+              - key: "{{ request.object.spec.issuerRef.kind || 'Issuer' }}"
+                operator: AnyNotIn
+                value:
+                  - Issuer
+    - name: certificaterequest-no-cluster-scoped-issuer
+      match:
+        any:
+          - resources:
+              kinds:
+                - CertificateRequest
+              namespaceSelector:
+                matchLabels:
+                  app.kubernetes.io/managed-by: ksail
+      # Same platform carve-out as the Certificate rule above, plus cert-manager
+      # itself: it derives a CertificateRequest from every Certificate under its
+      # own identity and copies the issuerRef, so denying it here would break
+      # issuance for any Certificate the flux-system carve-out admitted.
+      exclude:
+        any:
+          - subjects:
+              - kind: ServiceAccount
+                name: kustomize-controller
+                namespace: flux-system
+          - subjects:
+              - kind: ServiceAccount
+                name: cert-manager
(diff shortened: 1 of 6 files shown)


devantler:

## Verification record

**Exercised as a user, not just traced.** The fixture (`tests/restrict-tenant-issuer-refs`) runs in
CI via `kyverno test ./tests` and passes 7/7 — but a passing Kyverno fixture is worth very little on
its own, so here is what was actually proven.

### The fixture is not vacuous

The Kyverno CLI does not read namespace labels from `Namespace` objects in `resources.yaml`. Without
a `values.yaml` declaring them, a rule matched by `namespaceSelector` is reported `Excluded` for
every resource and **the whole fixture passes while the policy does nothing**. Ablated to confirm:
deleting `values.yaml` still reports `Test Summary: 7 tests passed and 0 tests failed`, with every
row's REASON reading `Excluded`. The green summary is not the signal; the REASON column is. With
`values.yaml` present, in-scope rows read `Ok` and only `kube-system/gateway-certificate` reads
`Excluded`, which is the intended scoping.

### Each design decision ablated separately

| Ablation | Result | What it proves |
|---|---|---|
| Allow-list → `Equals ClusterIssuer` | `tenant-cert-external-clusterissuer`: **want fail, got pass** | A name-based deny misses every non-cert-manager cluster-scoped kind (`AWSPCAClusterIssuer`) |
| Drop `\|\| 'Issuer'` default | `tenant-cert-kind-omitted`: **want pass, got fail** | Omitted `kind` is legal and means namespaced; denying it would break 2 of the 9 certs live on prod |
| Remove `namespaceSelector` | `kube-system/gateway-certificate`: **want skip, got fail** | The scoping is what keeps the real platform certificate working |

Each ablation was verified to have actually applied (match counts before/after) and verified restored
afterwards; the final green above is from the committed tree with a clean worktree.

### The carve-outs, proven behaviourally

`kyverno test` cannot express `request.userInfo`, so the excludes are not coverable by the CI
fixture. Proven separately with `kyverno apply --userinfo`, per identity:

| Applying identity | Denied |
|---|---|
| tenant SA | `tenant-cert-clusterissuer`, `tenant-cert-external-clusterissuer`, `tenant-cr-clusterissuer` |
| `flux-system:kustomize-controller` | *nothing* — platform carve-out holds on both rules |
| `cert-manager:cert-manager` | only the two Certificates; **not** `tenant-cr-clusterissuer` |

The third row is the one that matters operationally: cert-manager derives a CertificateRequest from
every Certificate under its own identity and copies the `issuerRef`, so without that exclusion a
platform-authored ClusterIssuer-backed Certificate would be admitted and then have its derived
request rejected — an issuance failure pointing at the wrong object.

### Live state checked before choosing Enforce over Audit

The issue proposed starting in Audit. Read prod first: 4 tenant namespaces
(`ascoachingogvaner`, `doggy-countdown`, `github-config`, `wedding-app`), and the only Certificate in
any of them is `ascoachingogvaner/ascoachingogvaner-dk`, already on a namespaced `Issuer`. The only
`ClusterIssuer` reference cluster-wide is `kube-system/gateway-certificate`, which is out of scope.
Audit exists to discover unknown violators; here the violator set is empirically empty, and a
security boundary left in Audit is not a boundary. Repo-only evidence would not have been enough —
tenants deploy from their own repositories — which is why this was checked against the live cluster.

Both overlays build (`kubectl kustomize k8s/clusters/{local,prod}/`), and the policy renders into
`k8s/bases/infrastructure/cluster-policies/` (23 policies, this one included).

devantler:

@codex review




**Starting at Codex because the CodeRabbit lane is quota-exhausted account-wide**, measured three
times in the last two hours: its summary comment carries `Review limit reached — you've reached your
PR review limit, so we couldn't start this review` on platform#2830 (04:40Z), world-at-ruin#441
(05:17Z) and world-at-ruin#422 (06:37Z, a sibling instance's request). Its acknowledgement still
reads "✅ Review finished", so the ack is not the signal. That is a provider quota state rather than
a verdict, so the ladder advances; recorded as `cr:no-gate@a0e02c9e` so a later run does not
re-spend it on this head.

CI is green at this head (16 success / 11 skipped, 0 failures) and `mergeable_state: clean`.
Verification record, including the ablations and the per-identity carve-out proof, is in the
comment above.

devantler:

## Promoting — all three readiness conditions hold at `a0e02c9e73`

1. **Programmatically tested** — CI green at head (17 success / 11 skipped, **0 failures**);
   `🧪 Validate Manifests` (which runs `kyverno test ./tests`) passes. The three design decisions
   are each ablated separately in the verification record above, and the anti-vacuity fix is
   committed as `tests/restrict-tenant-issuer-refs/values.yaml` — without it the fixture passes
   while the policy does nothing.
2. **Reviewed** — Codex **green at this exact head**: `Codex Review: Didn't find any major issues`,
   `Reviewed commit: a0e02c9e73`. 0 review threads, 0 unresolved, no non-thread findings.
   CodeRabbit was quota-exhausted account-wide when requested (`cr:no-gate@a0e02c9e` recorded
   above), which is a provider quota state rather than a verdict, so the ladder advanced.
3. **Tried and evaluated as a user** — the carve-outs were proven with `kyverno apply --userinfo`
   per applying identity (tenant SA denied, `kustomize-controller` and cert-manager's derived
   CertificateRequest both correctly excluded), and prod was read before choosing Enforce over
   Audit: the violator set is empirically empty.

`mergeStateStatus: CLEAN`. Merging.

devantler:

## Evicted from the merge queue by a systemic blocker — do NOT simply re-queue

This PR was enqueued at ~08:30Z and evicted when its `merge_group` CI run
([30342758716](https://github.com/devantler-tech/platform/actions/runs/30342758716)) failed. Root cause,
per the merge-queue rule that a kick-out is diagnosed before re-queuing:

```
Could not atomically acquire the GHCR synchronization lock
```

Both `🚀 Deploy to Prod` and the `🩹 Heal Prod` job failed at the same point, in
`scripts/refresh-flux-ghcr-auth.sh`. This is **not** a one-off transient and not the Cilium
mutual-auth flake (#2337) — re-queuing will re-hit it.

The fix is already in flight as **#2836** (`fix(deploy): unblock GHCR synchronization`, issue #2835).
Re-queue this PR once that lands.

**Prod is healthy** — checked read-only: 8/8 nodes `Ready`, and the only not-Ready Flux Kustomizations
are ordinary in-progress reconciles. The failed heal job reached `✅ Prod API server reachable` before
failing on the lock, so nothing is stranded that needs manual recovery. (The two `::error::` lines in
the log are unexpanded script source, not firings.)

Nothing about this PR's own content is implicated: its pentad was clear and its review green at
`a0e02c9e73` when it was queued.

devantler:

## Evicted from the merge queue twice — same systemic cause, so it is NOT being re-queued

This PR was promoted and enqueued at `~08:13Z` on a clear pentad (Codex green at `a0e02c9e73`, CI
17/0, `CLEAN`). It has since been kicked out of the queue **twice**, and both `merge_group` runs
failed for the same reason — not a flake, and nothing to do with this diff:

```
##[error]Could not atomically acquire the GHCR synchronization lease after concurrent updates.
```

- run at `08:14:24Z` — `🚀 Deploy to Prod` failed, then `🩹 Heal Prod` **also** failed on the identical error
- run at `08:34:01Z` — same error in both jobs again

The PR is currently out of the queue (`autoMergeRequest: null`, still `CLEAN`).

**Root cause is already owned and fixed elsewhere: #2836 (`Fixes #2835`).** The deploy path rendered
the Lease `acquireTime`/`renewTime` as second-precision RFC 3339, which the Kubernetes `Lease` API
decodes as `MicroTime` and rejects for lacking fractional seconds; the script then retried the same
invalid object and reported every rejection as a concurrent update, which is why the error names
concurrency and the real API cause was invisible.

**Deliberately not re-queued.** Re-queuing before that lands just reproduces the failure, and each
attempt runs a real prod deploy *and* a failing heal — the contract's rule for a systemic
`merge_group` failure is to land the root-cause fix first rather than loop the PR through the queue.
This one is a Kyverno policy addition with an empirically empty violator set, so there is no urgency
in racing it ahead of the deploy path's own repair.

**Next step for whoever picks this up:** once #2836 has merged, confirm a prod deploy on `main`
succeeds, then re-enqueue this with a bare `gh pr merge 2833 --squash`. Its own readiness evidence is
unchanged and does not need to be re-derived — but the Codex green is pinned to `a0e02c9e73`, so if
anything pushes to this branch in the meantime the review gate re-opens at the new head.

I am not driving #2836 myself: another instance requested a review on it at `08:29:59Z` at head
`805ef3d6` and is actively working it.

devantler:

## Correction to my note above: #2836 alone is NOT enough to unblock this

I said "once #2836 has merged, confirm a prod deploy on `main` succeeds, then re-enqueue". Having now
read #2836's own `merge_group` run, that instruction is wrong and would waste a queue attempt.

**#2836's lease fix genuinely works** — its run at `08:50:57Z` got past lease acquisition and into the
runtime-overlap phase. **But that run still failed**, on the next link in the chain:

```
admission webhook "ivpol.validate.kyverno.svc-fail-finegrained-verify-app-images"
  denied the request: Policy verify-app-images failed: policy not evaluated
```

after six runtime-pull-probe attempts, then a fail-closed refusal to drain onto an unproved runtime.

So the deploy path has **two** blockers in series, and only the first is fixed. Details and evidence
are on #2836; I have not enqueued that PR either, for the same reason I am not enqueueing this one.

**Revised next step:** re-enqueue this PR only once a `merge_group` run — anyone's — actually reaches a
successful `🚀 Deploy to Prod`. That is the single observable condition; the merge state and review of
*this* PR are already clear and unchanged (Codex green pinned to `a0e02c9e73`, CI 17/0, `CLEAN`), so
nothing here needs redoing.

Until then every queue attempt runs a real prod deploy **and** a failing heal, which is why the
contract's rule for a systemic `merge_group` failure is to fix the root cause rather than loop the PR
through the queue.

devantler:

## Evicted from the merge queue a third time — same blocker, do not re-queue yet

This PR was enqueued again at `11:11:36Z` and evicted again. I pulled the
`merge_group` run before drawing any conclusion, and it is **the same GHCR lease
failure**, not a new one and not a transient:

- run [`30353794218`](https://github.com/devantler-tech/platform/actions/runs/30353794218) —
  `🚀 Deploy to Prod` failed at `🚀 Deploy to Production`, and the `🩹 Heal Prod`
  job failed behind it for the same reason. The terminal line is:

  > Could not atomically acquire the GHCR synchronization lease after concurrent updates.

That is the failure mode analysed in #2836/#2835: the `Lease` object needs
fractional-second `MicroTime`, and the error message blames "concurrent updates"
while the real cause is the serialization. It is **systemic to the queue**, so it
hits this PR on every attempt regardless of its contents.

**Eviction history** — all three are the same failure, so re-queuing without the
fix is not a retry, it is a repeat:

| Enqueued | `merge_group` CI |
|---|---|
| 08:14:24Z | failure |
| 08:31:55Z | failure |
| 11:11:36Z | failure |

**Gate for the next attempt is unchanged:** re-queue only once a `merge_group`
run has reached a **successful `🚀 Deploy to Prod`**. #2836 merging is necessary
but has not been sufficient on its own — the second blocker recorded on #2836
(`policy not evaluated` on the fine-grained image-verification policy) sits
behind the lease and only becomes visible once the lease clears.

This PR itself remains pentad-clear: `mergeable_state: clean`, Codex green at
`a0e02c9e73`, zero unresolved threads. Nothing here is a defect in the change.

devantler:

## Merge-queue eviction — blocked on the GHCR lease defect, do not re-queue

This PR is pentad-clear and was enqueued at 16:40:35Z. The merge queue **evicted it** at ~16:43Z.

Root cause, from its own `merge_group` run (30379359829), not inferred:

```
##[error]Could not atomically acquire the GHCR synchronization lease after concurrent updates.
```

The same error then failed the **`🩹 Heal Prod`** job, so the post-failure restore path is affected
too, not just the deploy.

**This is systemic, not specific to this PR.** The failing step is `🚀 Deploy to Prod`, which every
PR entering the queue runs. The message naming "concurrent updates" is misleading — the underlying
defect is the `Lease` object's timestamp serialization, and it reproduces with no contention.

Worth separating from the earlier eviction at 15:59Z, which had a *different* cause: a transient prod
API outage during a control-plane reboot that left `prod-control-plane-2` cordoned. That one has
cleared — all nodes are `Ready` and `unschedulable=false` as of 16:30Z. So a healthy cluster is **not**
sufficient; this eviction happened against a healthy one.

**Blocked on:** #2836 (`fix(deploy): unblock GHCR synchronization`), which is in flight in the
`codex/*` lane.

**Do not re-queue this PR until #2836 is merged** — re-queuing only burns another production deploy
and another failed heal. Nothing about this PR needs to change; it is waiting on the queue, not on
its own content.

Issue #2627, opened 2026-07-13
Make root OCI publication verified and atomic before moving latest


devantler:
> 🤖 Generated by the Codex Daily AI Engineer instance.

## Problem

Production's root Flux `OCIRepository` tracks `ghcr.io/devantler-tech/platform/manifests:latest` every minute, but live `spec.verify` is null. The intended keyless Cosign policy is nested at `spec.cluster.verify` in `ksail.prod.yaml`; KSail v7.167 reads `spec.workload.flux.verify`, so the unknown field is silently ignored.

The production deploy currently moves mutable `latest` before signing and attesting it. With verification disabled, Flux can fetch an unsigned artifact immediately after push. Even after the verify config is corrected, Flux can fetch after signing but before provenance/SBOM attestations complete. The DR rebuild is coupled to the same defect: it pushes an unsigned artifact, lacks `id-token: write`, and its workflow identity is not in the intended allow-list, so simply moving `verify` would make a clean DR reconcile fail.

Live evidence on 2026-07-13: root source `ref.tag=latest`, `interval=1m0s`, `spec.verify=null`, Ready on the current main artifact. This is pre-existing and separate from the GHCR credential propagation fix in #2625.

## Proposed direction

- Move the Cosign configuration to the KSail-supported `spec.workload.flux.verify` path and add a rendered/live-shape regression that fails if root verification is absent.
- Make publication atomic from Flux's perspective: publish to an immutable staging reference, sign and attest it, then promote/repoint `latest`; or suspend/repoint the root source with explicit failure recovery until the complete supply-chain bundle exists.
- Give DR a valid signed-artifact path (`id-token: write` plus a narrowly allowed DR identity), or have DR consume an already signed immutable artifact instead of publishing unsigned bytes.
- Prove normal merge-group, manual CD, failed delivery, and from-zero DR behavior.

## Acceptance criteria

- The live root OCIRepository has non-null Cosign verification with only the intended production identities.
- Flux cannot observe a new artifact until signature and required attestations are complete.
- Any failed push/sign/attest leaves the last verified artifact active.
- DR can rebuild from zero without weakening verification and finishes on a signed, allowed artifact.
- Static tests validate config nesting, workflow permissions/identity, and publication ordering.

Rough size: medium; coordinate the KSail config correction, publication transaction, and DR signer path in one rollout so no intermediate state breaks production or recovery.

Related: #2613, #2625

devantler:
[...]

PR #2837, opened 2026-07-28
fix(kubescape): stop the Headlamp mirror granting exceptions the CRs never granted

devantler:
> 🤖 Generated by the Agentic Engineer

## Why

The Headlamp dashboard is how we read the platform's security posture. Its exception list is a
hand-maintained copy of the `ClusterSecurityException` CRs, and the copy had drifted — in the
direction that **hides findings**.

The worst entry excepted "immutable container filesystem" (C-0017) for *every workload in the
cluster*. The CRs never say that: they except C-0017 in three narrow, deliberate places, and one of
them carries a comment saying it is scoped precisely so it never suppresses C-0017 for the workloads
that do pass it. The dashboard was overriding that decision. A second entry excepted a group that no
CR mentions — it had been copied out of a trailing code comment.

Two real exceptions were missing at the same time, so the dashboard also reported legitimately-excepted
workloads as failing. The mirror was wrong in both directions at once.

Nothing is being actively hidden today — sampled workloads outside the excepted scopes pass C-0017
anyway — so this closes a latent suppression rather than an active incident.

## What

Makes the mirror faithful to the CRs: removes the two exceptions that have no source, and adds the two
that were missing. Control sets now match the CRs exactly across all 19 shared policies. The two
host/CIS exceptions stay deliberately unmirrored, as the file header already documents — those
controls never attach to a workload.

This is the first slice of #2586. Ending the hand-maintenance itself (generating this file and gating
drift in CI) is the follow-up; the notes for it are on the issue.

Part of #2586

Issue #2120, opened 2026-06-17
Give the prod k8s API a stable endpoint (VIP/LB) so CP-node recreates don't break kubeconfig


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

The prod k8s API has **no stable endpoint** — the Talos `cluster.controlPlane.endpoint` is pinned to a **single control-plane node's public IP** (was `https://178.105.204.76:6443`, the original CP-1). When that node is recreated or deleted (routine, per the cattle-CP model in the DR runbook — see [[prod-cd-frozen-by-down-cp-node]] incidents), the endpoint points at a dead host.

**Impact — recurs on every CP-node recreate:**
- The `prod` environment `KUBE_CONFIG` secret (and local `~/.kube/config`) point at a dead IP → the deploy pipeline's `🩺 Verify prod cluster is reachable` preflight fails with `i/o timeout`, evicting PRs from the merge queue and freezing all GitOps delivery.
- `talosctl kubeconfig` **regenerates the same dead endpoint** (it copies `controlPlane.endpoint` verbatim), so a plain regen doesn't fix it — every recovery requires a manual *repoint to a live CP node IP* (DR runbook **Scenario 9**), which is pure toil and a sharp edge during an incident.

Latest occurrence: 2026-06-17 (CP recreates after the nftables ENOBUFS boot-hang); secrets manually refreshed, but the root fragility remains.

## Proposed direction — give the API server a stable, failover-aware endpoint

Two viable options on Hetzner Cloud; both make the endpoint independent of any individual CP node's lifecycle. Recommend **Option A** (native, no recurring cost, fits the cattle-CP philosophy):

**Option A — Talos-managed Hetzner floating-IP VIP (recommended).** Talos natively supports a shared control-plane virtual IP backed by a Hetzner Cloud floating IP (`machine.network.interfaces[].vip.hcloud.*`); Talos elects a leader among the CP nodes via etcd and reassigns the floating IP to it on failover. Point `controlPlane.endpoint` at the floating IP / its DNS name. No load balancer, no monthly LB cost. *(Confirm the exact `vip.hcloud` schema against current Talos + ksail docs before implementing.)*

[...]

PR #2840, opened 2026-07-28
feat(security): give the DR rebuild a signed-artifact publication path

### LINKED ISSUE #2839: Give the DR rebuild a signed-artifact publication path

## Problem

The DR rebuild (`dr-rebuild.yaml`) publishes to the same mutable tag as a normal prod deploy — `ghcr.io/devantler-tech/platform/manifests:latest` — but it never signs what it pushes, and its `rebuild` job holds no `id-token: write`, so keyless signing is not even possible.

That makes it a hard blocker on the parent: root-source signature verification cannot be switched on while one writer to the tag publishes unsigned bytes under an identity the allow-list does not name. The first recovery after enabling verification would produce an artifact Flux refuses — during an incident, when manual repair is least available.

## Live evidence (2026-07-28)

The posture is currently inverted — every tenant source is verified, the infrastructure root is not:

| Source | `spec.verify` |
|---|---|
| `flux-system/flux-system` (root) | **null** |
| `ascoachingogvaner`, `aws`, `doggy-countdown`, `github-config`, `wedding-app` | `cosign` |

`dr-rebuild.yaml` job `rebuild` declares only `contents: read` + `packages: write`, and the file contains no `cosign` invocation at all.

## Scope

Give DR a signed-artifact path, leaving verification OFF:

- `id-token: write` on the rebuild job.
- Install cosign and sign the digest DR already resolves (`platform_manifest`), not the mutable tag.
- Name the DR workflow identity in `ksail.prod.yaml` `matchOIDCIdentity`, pinned to `refs/heads/main`.
- A CI-enforced contract so none of the four can silently regress.

## Why this is safe to land first

Signing an artifact that nothing yet verifies is inert — it changes no reconcile behaviour. It is a strict prerequisite for the parent, and it is the only part that can ship without first making publication atomic.

## Acceptance criteria

- The DR rebuild signs the artifact it publishes, using the digest it resolved.
- The DR signing identity is in the cosign allow-list, pinned to `refs/heads/main`.
- CI fails if the permission, the sign step, the digest pinning, or the allow-list entry is removed.
- No change to production reconcile behaviour (verification remains off).

Part of #2627

devantler:

## Why

Production's root Flux source delivers the entire platform — every controller, tenant binding and policy arrives through it. Today it is the **only** source with no signature verification, while every tenant source has it:

| Source | `spec.verify` |
|---|---|
| **`flux-system/flux-system` (root)** | **null** |
| ascoachingogvaner, aws, doggy-countdown, github-config, wedding-app | `cosign` |

Turning verification on is blocked by the disaster-recovery rebuild: it publishes to the same tag as a normal deploy, but never signs what it pushes and cannot sign, because its job holds no `id-token` permission. Enabling verification first would mean the next recovery publishes an artifact Flux refuses — during an incident, when manual repair is least available.

## What

Gives DR the signing path a normal deploy already has, and names its identity in the allow-list so a signed DR artifact is actually accepted. A CI contract pins all four guarantees so none can silently regress.

**Verification stays off in this change.** Signing an artifact nothing yet verifies is inert, which is what makes this safe to land ahead of the publication-atomicity and config-nesting work the parent still needs. No production reconcile behaviour changes.

Fixes #2839
Part of #2627

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,6 +39,11 @@ jobs:
           go test ./scripts/validate-merge-group-heal
           go run ./scripts/validate-merge-group-heal .github/workflows/ci.yaml
 
+      - name: 🖋️ Validate DR signing contract
+        run: |
+          go test ./scripts/validate-dr-signing
+          go run ./scripts/validate-dr-signing .github/workflows/dr-rebuild.yaml ksail.prod.yaml
+
       - name: 🔍 Filter paths
         uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
         id: filter
--- a/.github/workflows/dr-rebuild.yaml
+++ b/.github/workflows/dr-rebuild.yaml
@@ -61,6 +61,7 @@ jobs:
     permissions:
       contents: read # checkout repository
       packages: write # push OCI artifacts to GHCR
+      id-token: write # keyless cosign signing (Fulcio OIDC)
     steps:
       - name: 🛑 Verify confirmation phrase
         env:
@@ -178,6 +179,41 @@ jobs:
             exit 1
           fi
           echo "digest=${DIGEST}" >>"${GITHUB_OUTPUT}"
+
+      - name: ⚙️ Install cosign
+        # Same installer and pinned release as the deploy-prod composite, so a
+        # DR-published artifact is signed by the identical toolchain a normal
+        # deploy uses. Renovate bumps COSIGN_VERSION via the custom regex
+        # manager in .github/renovate.json.
+        env:
+          # renovate: datasource=github-releases depName=sigstore/cosign extractVersion=^v(?<version>.+)$
+          COSIGN_VERSION: "3.1.2"
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+        with:
+          cosign-release: "v${{ env.COSIGN_VERSION }}"
+
+      - name: 🖋️ Sign OCI manifest with cosign (keyless)
+        # A DR rebuild writes the same mutable tag as a normal deploy, so it
+        # must produce the same supply-chain evidence — otherwise enabling
+        # verification on the root source turns every recovery into a rejected
+        # artifact at the exact moment manual repair is hardest.
+        #
+        # Signs the digest resolved by the step above rather than the tag: a
+        # concurrent deploy could move `latest` between resolve and sign, and
+        # signing the tag would then attest bytes we never published. The
+        # signing identity is this workflow at its dispatch ref
+        # (dr-rebuild.yaml@refs/heads/main), which ksail.prod.yaml's
+        # matchOIDCIdentity allow-list names explicitly.
+        env:
+          # Keyless is the default when --key is omitted (COSIGN_EXPERIMENTAL
+          # was retired in cosign 2.x). GHCR credentials come from
+          # ~/.docker/config.json written by the auth step above.
+          COSIGN_YES: "true"
+          DIGEST: ${{ steps.platform_manifest.outputs.digest }}
+        run: |
+          set -euo pipefail
+          REF="ghcr.io/devantler-tech/platform/manifests@${DIGEST}"
+          cosign sign --yes --recursive "${REF}"
 
       - name: 🔎 Verify Flux GHCR pull credential after publish
         id: verify_flux_ghcr_auth_after_push
--- a/docs/dr/runbook.md
+++ b/docs/dr/runbook.md
@@ -554,7 +554,12 @@ The production deploy closes the bootstrap loop in this order:
    `--allow-incomplete-fanout` mode stages `variables-base` and repairs root auth
    so the first reconcile can create the chain. Normal mode fails closed on any
    missing fan-out resource.
-4. Push and sign the artifact with `GHCR_TOKEN`, revalidate the newly-published
+4. Push the artifact with `GHCR_TOKEN`, then cosign-sign the resolved digest
+   keyless: Fulcio mints the certificate from the workflow's OIDC identity
+   (`dr-rebuild.yaml@refs/heads/main`), and `GHCR_TOKEN` only pushes the
+   resulting signature. That identity must stay listed in `ksail.prod.yaml`'s
+   `matchOIDCIdentity`, or a recovery publishes an artifact a verifying cluster
+   refuses. Revalidate the newly-published
    artifact with `--check-only`, and only then explicitly reconcile Flux. DR
    runs the full bridge again after every Flux Kustomization is Ready, proving
    that bootstrap mode completed the entire fan-out, and once more after an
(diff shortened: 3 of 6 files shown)


devantler:

## Review lane: advancing to Codex, with the CodeRabbit evidence recorded

CI is green at this head (24 success / 3 skipped / 0 failures) and there are no unresolved threads, so the only thing outstanding is a current-head review.

CodeRabbit is **not** being skipped casually — it is measurably unavailable to this PR right now:

- At 11:34:33Z its summary on another PR read *"Review limit reached — next review available in **39 minutes**"* (i.e. ~12:13:33Z), while its own command reply simultaneously claimed `Full review finished`. The limit is per-developer, so it is shared across every instance and every PR in this portfolio.
- The slot reopened at ~12:13:33Z and a concurrent request on #2830 was accepted at **12:14:00Z** — 27 seconds later, and 31 seconds before I read the lane.

Contending for that same slot would double-spend a metered lane against a sibling instance, which is a known failure mode here. Advancing to Codex instead, per the fallback ladder.

devantler:

@codex review

devantler:

## Verification record — `9db629ef`

### Programmatically tested

| Check | Result |
|---|---|
| `go run ./scripts/validate-dr-signing …` on the repo **before** the fix | **RED** — `rebuild job is missing id-token: write` |
| same, after the fix | **GREEN** — `DR signing contract passed.` |
| `go test ./scripts/validate-dr-signing` | ok — 4 ablations + ordering control + allow-list ablation + CLI |
| `golangci-lint run ./...` (unscoped, repo-wide) | **0 issues** |
| `actionlint` on both edited workflows | exit 0 |
| `gofmt -l` / `go vet` / `go build ./...` | clean |
| CI at this head | **24 success / 3 skipped / 0 failures** |

The four ablations each pass **individually** (not just as a parent test): removing `id-token`, removing the sign call, swapping the digest ref for the mutable tag, and renaming the job each fail, and each fails for its **own** reason.

**One ablation was nearly vacuous and I checked rather than assumed.** The ordering assertion anchors on the string `workload push`, which also appears in a file-header comment at line 7. I probed the extracted job block directly: exactly **1** in-job occurrence (`pushIdx=92`, `signIdx=158`), so the header comment is correctly outside the block and the ordering check is load-bearing.

### Tried and evaluated as a user

- **The validator** — exercised directly in both states, as above.
- **The CI wiring** — I did not infer this from a green job. The step's own conclusion in run `30357018194` is `🖋️ Validate DR signing contract | completed | success`, in the `🔍 Detect Changes` job next to the merge-group heal contract it mirrors. It genuinely runs on every PR.
- **The YAML edits** — verified structurally, not by eye: `yq '.jobs.rebuild.permissions'` shows `id-token: write`, and `yq '.spec.cluster.verify.matchOIDCIdentity[].subject'` returns three entries including the DR identity.
- **Step preconditions** — the sign step sits after push (149) → GHCR auth (156) → digest resolve (171, `id: platform_manifest`) → install cosign (183), so both `steps.platform_manifest.outputs.digest` and `~/.docker/config.json` exist when it runs.

### What I could NOT exercise, stated plainly

**The signing step has not been run.** Executing it requires a from-zero DR rebuild against production, which this repo forbids and which would be wildly disproportionate for a supply-chain increment. Its runtime behaviour is therefore evidenced by construction — identical installer, pinned release, and sign invocation as the `deploy-prod` composite that signs this same artifact on every merge-queue deploy — not by observation.

**The allow-list entry is deliberately inert today**, because the block it lives in is at `spec.cluster.verify` while KSail reads `Workload.Flux.Verify`. That misnesting is the parent issue's step 4 and is intentionally *not* fixed here — flipping it is what switches verification on, and doing that before publication is atomic would break production and recovery. Live root source still reads `spec.verify: null`, which is the expected state for this PR.

devantler:

## Self-promoted — and deliberately NOT enqueued

**Readiness at `9db629ef`, all three conditions met:**

| Condition | Evidence |
|---|---|
| Programmatically tested | CI **24 success / 3 skipped / 0 failures**; RED→GREEN proven on the real repo; 4 ablations each passing individually |
| Reviewed | **Codex green at the exact head** — `Reviewed commit: 9db629efaa` == head `9db629efaa`, 0 inline comments, 0 findings |
| Tried and evaluated as a user | recorded above — including confirming the new CI step's *own* step conclusion is `success`, rather than inferring it from a green job |

Hygiene pentad: green checks, **0** unresolved threads, **0** non-thread review findings, `mergeable_state: clean`, current-head green review.

### Why this is not going into the merge queue

This PR edits `ksail.prod.yaml`, which is in `ci.yaml`'s `k8s` path filter — so a merge-queue run would execute `🚀 Deploy to Prod`, and **that path is currently failing**:

- #2833 was enqueued and evicted twice; its most recent `merge_group` retry at **11:11:36Z** failed again.
- The root fix, #2836, is itself still open and `mergeable_state: blocked` as of 12:19Z.
- No `merge_group` run has reached a successful deploy since.

Enqueueing now would just re-hit a known systemic failure and burn a prod-deploy cycle, so this follows the same hold already applied to #2833 and #2837.

**Re-enqueue condition:** once any `merge_group` run reaches a successful `🚀 Deploy to Prod`. Nothing about this PR needs to change first — it is merge-ready on its own merits.

Note the change is inert in production either way: it adds signing to a recovery path and an allow-list entry that stays unread until the parent's config-nesting step lands. The live root source still reads `spec.verify: null`, which is expected.

devantler:

## Merge-queue eviction at 13:04Z — same GHCR lease failure, not a new one

This PR was enqueued at 13:04:17Z and evicted. Root cause pulled before concluding, rather than re-queuing blind:

- `merge_group` run [30361773380](https://github.com/devantler-tech/platform/actions/runs/30361773380) — `🚀 Deploy to Prod` failed at its `🚀 Deploy to Production` step with `Could not atomically acquire the GHCR synchronization lease after concurrent updates.`
- `🩹 Heal Prod` then failed re-deploying main, which is expected while main lacks the fix.
- Job conclusions: 3 failure / 3 skipped / 2 success.

That is the known systemic lease defect, whose root fix is #2836 — still open and not merged. The error message blames "concurrent updates" and is misleading; the actual cause is the `Lease` object's MicroTime encoding.

**The standing gate therefore still holds: do not re-enqueue this PR until a `merge_group` run reaches a successful `🚀 Deploy to Prod`.** Nothing about this PR needs to change — it remains pentad-clear (CI green, 0 unresolved threads, `mergeable_state: clean`, Codex green at `9db629efaa`). This is the fourth eviction recorded across #2833/#2836/#2840 on the same cause.

devantler:

## Pentad-clear but merge-queue blocked — do not enqueue yet

Verified merge-ready at the current head this tick: green required checks, zero unresolved review
threads, no non-thread review findings, no conflict, and a green review at head.

It is **not** enqueued, deliberately. `platform`'s merge queue runs `🚀 Deploy to Prod` for every PR
that enters it, and that step is currently failing for reasons unrelated to any individual PR. Proven
this tick rather than assumed: #2833 — equally pentad-clear — was enqueued at 16:40:35Z and evicted at
~16:43Z with

```
##[error]Could not atomically acquire the GHCR synchronization lease after concurrent updates.
```

which then failed the `🩹 Heal Prod` job too. Nothing has merged to `main` since 04:29Z.

Enqueuing this PR now would spend another failing production deploy and another failed heal, and it
would still be evicted. It is waiting on the queue, not on its own content — no change is needed here.

**Unblocks when** the GHCR synchronization fix (#2836) merges. Note that #2836 is itself being evicted
for a *different* reason — the deploy reaches the cluster API through the very control-plane node it
reboots (#2120) — so the ordering is #2120 → #2836 → this PR.

Issue #2451, opened 2026-07-04
feat(security): continuously ingest Kubescape findings — SARIF→Code Scanning gate + CRD→issue bridge


devantler:
> 🤖 Generated by the Daily AI Assistant

Part of #2447.

## Problem

Even once the scanners are fixed, live Kubescape findings (posture regressions, new reachable CVEs, runtime detections) **never enter the work backlog** — the daily engineer's survey is GitHub-only, so findings only get worked if a human files them. This is the structural reason the whole stack rotted unnoticed. We need continuous, idempotent ingestion so findings become tracked issues the engineer drains oldest-first, and a pre-merge gate so regressions never land.

## Proposed direction (full loop — settled with the maintainer)

Two complementary paths (repo is **public**, so GitHub Code Scanning is available and already in use — CodeQL, zizmor):

1. **Prevention gate (repo-scannable posture):** add a Kubescape **SARIF** scan step that uploads to **GitHub Code Scanning** on PRs (native dedup via `partialFingerprints`, auto-close-when-gone, Security tab). This complements the existing `ksail workload scan` compliance-threshold gate. Model the workflow on `.github/workflows/todos.yaml` (App-token + a reusable workflow targeting `devantler-tech/actions`, never `reusable-workflows`).
2. **Backlog bridge (live-only findings):** a scheduled job that reads the live Kubescape CRDs a static repo scan can't produce — `vulnerabilitymanifestsummaries`, runtime `node_agent_alert_counter` state, and any residual `configurationscansummaries` regressions — and opens/updates/closes **fingerprint-deduped, themed** `security` issues under #2447 (one issue per control/theme, **not** per resource; search-before-create; close-when-gone). This is what makes findings drainable oldest-first.

Anti-spam is the hard part: get idempotency + fingerprinting right or it floods. Start the bridge in report-only/dry-run, then enable issue writes.

## Rough size
L — decompose into: (a) SARIF→Code-Scanning PR gate; (b) the CRD→issue bridge (dedup/fingerprint/close-when-gone); (c) Kyverno-`Enforce` graduation of controls driven to zero.

## Acceptance criteria
- Kubescape SARIF appears in the repo Security tab on PRs; posture regressions surface pre-merge.
- Live-only findings become deduped `security` issues under #2447, auto-closed when resolved; no issue spam.
- Fixed posture controls graduate into Kyverno `Enforce` so they can't regress at admission.

devantler:
[...]

PR #2847, opened 2026-07-28
fix(ci): report what the Kubescape scan actually found

### LINKED ISSUE #2846: bug(ci): the Kubescape scan log reports "All controls passed" while the same run raises Code Scanning alerts

Part of #2451. Follows #2829, which added the Kubescape SARIF → Code Scanning upload.

## Evidence

Measured locally against `main` (ksail 7.x, `k8s` source dir, repo exceptions applied), running
exactly what CI runs:

```
ksail workload scan --framework nsa --exceptions <generated> --compliance-threshold 95 \
  --format sarif -o kubescape.sarif
```

- **Console output:** `All controls passed. No issues found` — exit 0.
- **SARIF written by the same run:** **4 findings** on real, un-excepted manifests:
  - `C-0034` (warning) ×3 — `bases/infrastructure/vault-backup/{job,cron-job}.yaml`, `bases/infrastructure/vault-config/job.yaml`
  - `C-0017` (note) ×1 — `bases/apps/umami/cron-job.yaml`

I confirmed none of these are suppressed: no policy in the generated exceptions file
(21 policies) covers `C-0034` or `C-0017`.

The printer is showing a **filtered** view, not an empty one. Re-running the same scan with
`--exceptions` removed produced **70** SARIF findings while the console table listed **one**
control (`C-0013`, 2 resources). So the console summary surfaces only a subset; the SARIF
carries the full set.

## Problem

Since #2829, those SARIF findings become Code Scanning alerts — so the two surfaces now
disagree in a way they did not before:

- The **CI log**, which is what a developer reads on a PR, says `All controls passed. No issues found`.
- The **Security tab** shows alerts for the same run.

Anyone trusting the log concludes there is nothing to fix. Findings that do not move the
compliance score enough to breach the floor are invisible in CI entirely, which is precisely the
class #2829 was added to make visible. The gate is working correctly; the reporting around it is
what misleads.

## Smallest useful change

After the scan step, derive a one-line summary from the SARIF the run just produced and print it
to the log and the job summary — count and per-control breakdown, plus a pointer to the Security
tab. No change to the gate, the threshold, or the upload.

## Acceptance criteria

- A run producing N SARIF findings prints a summary naming N and the control IDs, even when the
  compliance gate passes and the Kubescape table prints `All controls passed`.
- A run producing zero findings says so explicitly.
- The compliance-threshold gate's pass/fail behaviour is unchanged.
- No change to what is uploaded to Code Scanning.

## Size

S — one step in the `validate` job, reading a file the run already writes.

## Note on scope

The filtered console table comes from Kubescape's own printer, which ksail forwards. This issue is
deliberately scoped to the platform's CI reporting rather than proposing an upstream printer
change — the summary we need is cheap to derive on our side and does not depend on upstream.

devantler:

## Why

Kubescape's console table shows only part of what it finds. Measured on this repository: a scan whose SARIF carried 4 findings printed `All controls passed. No issues found`, and one carrying 70 findings listed a single control. Since those findings are now uploaded as Code Scanning alerts, the CI log and the Security tab actively contradict each other — and the log is what a developer reads on a PR.

## What

Reports the findings from the SARIF the scan already produces, so the log and the alerts cannot disagree. Covers pull requests and pushes to `main` (the latter is the only default-branch scan). Purely informational — the compliance threshold stays the sole merge gate, and a summary that cannot be built annotates rather than blocking a merge.

Fixes #2846
Part of #2451

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,6 +60,7 @@ jobs:
               - 'scripts/validate-embedded-json.py'
               - 'scripts/generate-kubescape-exceptions/**'
               - 'scripts/normalize-sarif-paths.sh'
+              - 'scripts/summarize-sarif-findings.sh'
               - 'go.mod'
               - 'go.sum'
               - 'scripts/tests/test-cilium-bandwidth-manager-component.sh'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,6 +61,7 @@ jobs:
               - 'scripts/generate-kubescape-exceptions/**'
               - 'scripts/normalize-sarif-paths.sh'
               - 'scripts/summarize-sarif-findings.sh'
+              - 'scripts/tests/test-summarize-sarif-findings.sh'
               - 'go.mod'
               - 'go.sum'
               - 'scripts/tests/test-cilium-bandwidth-manager-component.sh'
@@ -247,6 +248,15 @@ jobs:
         # Render the full component contract so the cAdvisor scrape and its
         # least-privilege delivery path cannot silently disappear.
         run: bash scripts/tests/test-opencost-usage-scraper.sh
+
+      - name: 📋 Validate Kubescape finding summary
+        if: needs.changes.outputs.k8s == 'true'
+        # The summary is what makes the CI log and the Security tab agree (#2846),
+        # and it fails by under-reporting rather than by crashing — a dropped
+        # control looks exactly like a clean scan. Pin the grouping, the counts and
+        # the two SARIF shapes that turn into hard jq errors if indexed naively: a
+        # result with no ruleId, and a ruleId absent from the rule table.
+        run: bash scripts/tests/test-summarize-sarif-findings.sh
 
       - name: 📢 Validate reconciliation-Alert namespace coverage
         if: needs.changes.outputs.k8s == 'true'
@@ -472,7 +482,17 @@ jobs:
         # assert that an upload happened.
         #
         # Informational only. The compliance threshold on the scan step remains
-        # the sole merge gate; this step cannot fail the build over findings.
+        # the sole merge gate.
+        #
+        # The script exits non-zero on a usage error or a SARIF it cannot parse,
+        # and leaves it to the caller whether that is fatal. This caller says no:
+        # a summary that cannot be built says nothing about posture, and the gate
+        # that does — the compliance threshold — has already run against the same
+        # file. `continue-on-error` keeps the failure loud (the step is still
+        # marked failed and annotated on the run) without blocking a merge on a
+        # reporting bug. Without it, this step silently becomes a second gate,
+        # contradicting the sentence above it.
+        continue-on-error: true
         if: >-
           !cancelled() &&
           steps.normalize-sarif.outputs.sarif == 'ready'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -408,6 +408,15 @@ jobs:
             echo "sarif=absent" >> "${GITHUB_OUTPUT}"
             exit 0
           fi
+          # Keep the scan's own output before normalization touches it. The
+          # normalizer DROPS results with an empty artifactLocation.uri (see its
+          # CONTRACT) because Code Scanning cannot anchor an alert without a
+          # file. That is right for the upload and wrong for the summary: a
+          # cluster-level control with no manifest behind it is exactly the kind
+          # of finding that is otherwise invisible, so summarizing the
+          # normalized file would under-report the very case this reporting
+          # exists for. Upload reads the normalized file; the summary reads this.
+          cp "${RUNNER_TEMP}/kubescape.sarif" "${RUNNER_TEMP}/kubescape.raw.sarif"
           bash scripts/normalize-sarif-paths.sh "${RUNNER_TEMP}/kubescape.sarif" k8s .
           # Only now is the file both PRESENT and root-relative. The upload gates
           # on this output, so a normalizer hard-failure (an unresolvable path)
@@ -498,7 +507,7 @@ jobs:
           steps.normalize-sarif.outputs.sarif == 'ready'
         run: |
           shellcheck scripts/summarize-sarif-findings.sh
-          bash scripts/summarize-sarif-findings.sh "${RUNNER_TEMP}/kubescape.sarif"
+          bash scripts/summarize-sarif-findings.sh "${RUNNER_TEMP}/kubescape.raw.sarif"
 
   naming:
     name: 🏷️ Validate Naming Conventions
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -453,6 +453,32 @@ jobs:
           # results are tracked and closed independently.
           category: kubescape-nsa
 
+      - name: 📋 Summarize Kubescape findings
+        # Kubescape's own console table shows only a subset of what it finds:
+        # measured on this repository, a scan whose SARIF carried 4 findings
+        # printed "All controls passed. No issues found". Since the upload above
+        # turns those findings into alerts, the log and the Security tab
+        # disagreed — and the log is what a developer reads on a PR (#2846).
+        #
+        # This reports from the SARIF the run already produced, so the two
+        # surfaces cannot drift: same file, same findings.
+        #
+        # Gated on the normalizer's output rather than `!cancelled()` alone, for
+        # the same reason the upload is: that output is the only signal that the
+        # file both exists and is well-formed. It runs even when the upload step
+        # did not — a fork PR cannot upload, but its author still benefits from
+        # seeing what the scan found, which is why the script's wording does not
+        # assert that an upload happened.
+        #
+        # Informational only. The compliance threshold on the scan step remains
+        # the sole merge gate; this step cannot fail the build over findings.
+        if: >-
+          !cancelled() &&
+          steps.normalize-sarif.outputs.sarif == 'ready'
+        run: |
+          shellcheck scripts/summarize-sarif-findings.sh
+          bash scripts/summarize-sarif-findings.sh "${RUNNER_TEMP}/kubescape.sarif"
+
   naming:
     name: 🏷️ Validate Naming Conventions
     needs: [changes]
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -417,6 +417,13 @@ jobs:
           # normalized file would under-report the very case this reporting
           # exists for. Upload reads the normalized file; the summary reads this.
           cp "${RUNNER_TEMP}/kubescape.sarif" "${RUNNER_TEMP}/kubescape.raw.sarif"
+          # Emitted BEFORE the normalizer runs, because it is the summary's gate
+          # and the summary must survive a normalizer failure. An unresolvable
+          # path aborts the step, so anything echoed after that line never
+          # happens — and gating the summary on `sarif=ready` would skip it
+          # exactly when an upstream path-shape change made the findings most
+          # worth reading.
+          echo "raw=ready" >> "${GITHUB_OUTPUT}"
           bash scripts/normalize-sarif-paths.sh "${RUNNER_TEMP}/kubescape.sarif" k8s .
           # Only now is the file both PRESENT and root-relative. The upload gates
           # on this output, so a normalizer hard-failure (an unresolvable path)
@@ -504,7 +511,7 @@ jobs:
         continue-on-error: true
         if: >-
           !cancelled() &&
-          steps.normalize-sarif.outputs.sarif == 'ready'
+          steps.normalize-sarif.outputs.raw == 'ready'
         run: |
           shellcheck scripts/summarize-sarif-findings.sh
           bash scripts/summarize-sarif-findings.sh "${RUNNER_TEMP}/kubescape.raw.sarif"
--- a/.github/workflows/validate-main.yaml
+++ b/.github/workflows/validate-main.yaml
@@ -161,6 +161,10 @@ jobs:
             echo "sarif=absent" >> "${GITHUB_OUTPUT}"
             exit 0
           fi
+          # Same split as ci.yaml: the normalizer drops empty-uri results because
+          # Code Scanning cannot anchor them, so the upload reads the normalized
+          # file and the summary below reads this untouched copy.
+          cp "${RUNNER_TEMP}/kubescape.sarif" "${RUNNER_TEMP}/kubescape.raw.sarif"
           bash scripts/normalize-sarif-paths.sh "${RUNNER_TEMP}/kubescape.sarif" k8s .
           echo "sarif=ready" >> "${GITHUB_OUTPUT}"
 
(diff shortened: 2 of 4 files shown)


devantler:

@coderabbitai review

devantler:

Fixed in `f45f1637` — valid finding, and the gap was real: the fixtures behind this logic were only ever exercised by hand.

**`scripts/tests/test-summarize-sarif-findings.sh`** now pins the report, following the `scripts/tests/test-*.sh` convention (bash + the runner's jq, no cluster). Cases: zero findings, a SARIF with no `runs` at all, grouping/counts/descending order across two controls, level+description resolved from the rule table, a result with **no `ruleId`**, a `ruleId` **absent from the rule table**, the `GITHUB_STEP_SUMMARY` copy matching the log, no stray writes when that variable is unset, and each exit-2 path (unparseable, missing file, no argument, too many arguments).

Wired in both directions, so it cannot decay silently: a new `📋 Validate Kubescape finding summary` step invokes it, and the test file itself is in the `k8s` path filter — the test asserts both, so deleting either wiring turns it red.

**8 ablations, each RED on its own assertion** (not merely red):

| ablation | failing assertion |
|---|---|
| null-`ruleId` guard removed | `a result with no ruleId must not abort the summary, got exit 2` |
| `sort_by(-.count)` removed | `controls must be ordered by descending count` |
| zero-findings message emptied | `an empty SARIF must state zero findings explicitly` |
| `// "unknown"` fallback removed | `a finding whose rule is missing from the table must report an unknown level` |
| step-summary block disabled | `the step summary must carry a heading` |
| parse failure exits 0 | `unparseable SARIF must exit 2, got 0` |
| CI step invocation removed | `CI must invoke this test, or it never runs` |
| path-filter entry removed | `this test must be in the k8s path filter` |

**Also acted on the `457-482` note in the same review**, which I judged correct on merit rather than informational: the step comment claimed it "cannot fail the build" while the script exits 2 on a SARIF it cannot parse and explicitly leaves that decision to its caller — a decision this caller had never actually made. Added `continue-on-error: true` so a reporting bug annotates loudly without becoming a second merge gate, and rewrote the comment to state the decision instead of asserting an outcome the YAML did not produce.

devantler:



@coderabbitai review

Head `f45f1637` adds the regression test requested in the previous review, plus `continue-on-error` on the summary step so it cannot act as a second merge gate.

devantler:

**User evaluation at `f45f1637`** — exercised in real CI on this PR, not reasoned from the diff.

`🧪 Validate Manifests` run [30397158763](https://github.com/devantler-tech/platform/actions/runs/30397158763), job `Validate Manifests`:

| step | result | observed output |
|---|---|---|
| 11. `📋 Validate Kubescape finding summary` (new) | **success** | `PASS: test-summarize-sarif-findings.sh` |
| 20. `📋 Summarize Kubescape findings` | **success** | `Kubescape: 4 finding(s) across 2 control(s).` |

So the new guard runs where it was wired rather than only locally, and adding `continue-on-error` did not change what the summary reports — the same 4 findings across 2 controls this PR was opened to surface.

Locally: 8 ablations each RED on their own assertion (table in the thread above), `shellcheck` clean, `actionlint` clean on the workflow.

devantler:



**CodeRabbit lane: `cr:no-gate@f45f1637`.** Its summary comment updated in place at 20:44:41Z carrying the `rate limited by coderabbit.ai` marker and a 38-minute window — the review did not run.

Worth recording precisely, because that same comment contains `between 3b111dc6c5… and f45f1637…`, i.e. it names the current head *inside* the rate-limit block. Head-naming is the walkthrough scope, not evidence a review happened; scoring it as a green here would have promoted this PR on a review that never ran.

Advancing to Codex (lane 2) rather than waiting out the window.

devantler:



@codex review

devantler:

Both P2s verified against the code and **both fixed** in `ffcf6531`. Neither was cosmetic — the first one falsified this PR's own premise.

**1. "Summarize before dropping locationless findings" — correct, and the more serious of the two.**
`scripts/normalize-sarif-paths.sh:92` does exactly what you describe:
```jq
| .results |= map(select((.locations[0].physicalLocation.artifactLocation.uri // "") != ""))
```
So summarizing the normalized file omits cluster-level controls with no manifest behind them — the precise class of finding that is otherwise invisible, which is what this PR exists to fix. The step comment at `ci.yaml:440-445` already documented the shape (C-0002, measured 0 of 4 findings with exceptions applied, 1 of 70 without), so this was latent rather than currently firing — but "currently zero" is not a property worth depending on.

Fixed by splitting the two consumers: `cp` the scan output to `kubescape.raw.sarif` before normalization, upload the normalized file, summarize the raw one.

**2. "Summarize the default-branch scan too" — also correct.** `validate-main.yaml`'s `kubescape-baseline` job scans, normalizes and uploads on push to main, and its own header says ci.yaml is `pull_request`-only. Without a summary there, the contradiction stands on the one path nobody opens a PR to read. Added the same step (reading the raw copy, `continue-on-error`).

**The guard is wiring-level, because that is what broke.** A test over the script alone would have passed throughout. Both workflows are now asserted to copy before normalizing and to summarize the raw file, with an explicit rejection of the normalized path. Plus a **differential** that keeps the split honest: it runs the real normalizer over a fixture carrying one located and one locationless finding, asserts the normalizer drops it (`1 finding across 1 control`), then asserts the summary over the raw fixture keeps it (`2 findings across 2 controls`). If the normalizer ever stops dropping, that precondition fails loudly and says the raw copy is no longer needed, rather than leaving a cargo-culted `cp` behind.

4 new ablations, each RED on its own assertion:

| ablation | failing assertion |
|---|---|
| ci.yaml summarizes the normalized file | `ci.yaml must summarize the raw copy, not the normalized file` |
| validate-main summary step deleted | `validate-main.yaml must summarize the raw copy, not the normalized file` |
| `cp` removed | `ci.yaml must copy the scan output before normalization` |
| normalizer stops dropping empty-uri | `PRECONDITION: the normalizer must drop the locationless finding` |

`shellcheck` and `actionlint` clean on both workflows.

devantler:



@coderabbitai review

Head `ffcf6531` fixes both P2s from the previous Codex review: the summary now reads the pre-normalization SARIF (the normalizer drops empty-uri results), and `validate-main.yaml` summarizes the default-branch scan too.

devantler:



**CodeRabbit lane: `cr:no-gate@ffcf6531`** — and deliberately so, even though its reply says *"Reviewed `ffcf6531`; no findings."*

That verdict arrived on the **chat surface** (`<!-- This is an auto-generated reply by CodeRabbit -->`), and there are **zero CodeRabbit review objects at this head**. The gate is defined on the artifact, not the sentence: a command reply is explicitly not a qualifying green, however diff-accurate it reads — and this one is accurate, it ran shellcheck and parsed both workflows. It is still the shape CodeRabbit produces when its review pipeline is throttled but its chat is not, so promoting on it would be promoting on a review that never entered the record.

Advancing to Codex. Not counting this as a failure of the change — only as a lane that did not deliver a gate-satisfying artifact.

devantler:



@codex review

devantler:

All three P2s reproduced first, then fixed in `91bdac55`. None were speculative — here is the before, measured on this branch:

```
{}                     rc=0  Kubescape: 0 findings in this scan.
{"version":"2.1.0"}    rc=0  Kubescape: 0 findings in this scan.
{"runs":null}          rc=0  Kubescape: 0 findings in this scan.
two ruleIndex results  →     "<no rule id>  unknown  — 2 finding(s)"   (one bucket, metadata gone)
```

**1. Structurally invalid SARIF read as a clean scan.** This was the worst of the three, because it is the failure this PR exists to remove wearing the PR's own output. `.runs` must now be an array or the script exits 2 naming the reason; `"runs": []` stays a legitimate zero.

This one also **corrected my own test**, which had asserted `{"version":"2.1.0"}` → exit 0 as *desired* behaviour. It was not; it was the bug, written down as an expectation. The case now asserts exit 2, and the comment says plainly that an earlier revision had it backwards.

**2. `ruleIndex` collapsed into one bucket.** Resolved **per run**, since the index is scoped to its own `run.tool.driver.rules` and resolving against a concatenated list would mis-attribute across runs. Bounds-checked, because jq indexes backwards from `-1` — an unguarded `ruleIndex: -1` would have silently labelled the finding with the *last* rule in the table, which is worse than not resolving it at all.

**3. Summary skipped when normalization aborts.** Correct, and it was a consequence of my own previous commit: I changed the summary's *input* to the raw copy but left its *gate* on `sarif=ready`. Now gated on `raw=ready`, emitted **before** the normalizer runs — after it would be unreachable, since an unresolvable path aborts the step. Both workflows.

5 new ablations, each RED on its own assertion:

| ablation | failing assertion |
|---|---|
| `.runs` shape check removed | `a non-SARIF document {} must name the reason` |
| ruleIndex resolution removed | `ruleIndex results must resolve to distinct controls, not collapse into one` |
| bounds check removed | `an out-of-range ruleIndex must not be attributed to a real control` |
| gate back on `sarif=ready` | `must gate the summary on the raw copy, not on normalization succeeding` |
| `raw=ready` moved after the normalizer | `must emit raw=ready BEFORE the normalizer (raw at 427, normalizer at 426)` |

`shellcheck` and `actionlint` clean.

devantler:



@coderabbitai review

Head `91bdac55` fixes three P2s from the last Codex round: a valid-JSON non-SARIF file no longer reports as a clean scan, `ruleIndex`-identified results resolve per run instead of collapsing into one bucket, and the summary is gated on the raw copy rather than on normalization succeeding.

devantler:

**User evaluation re-run at `91bdac55`** — necessary, not a formality: the last two commits changed the script's *behaviour* (strict `.runs` validation, `ruleIndex` resolution), so the evidence from `f45f1637` no longer covers it.

`🧪 Validate Manifests` run [30400022118](https://github.com/devantler-tech/platform/actions/runs/30400022118):

| step | result | observed output |
|---|---|---|
| 11. `📋 Validate Kubescape finding summary` | success | `PASS: test-summarize-sarif-findings.sh` |
| 20. `📋 Summarize Kubescape findings` | success | `Kubescape: 4 finding(s) across 2 control(s).` |

The second row is the one worth reading: **identical to the output at `f45f1637` and at `2486ebd2`**. Stricter validation and per-run `ruleIndex` resolution changed nothing on this repository's actual scan output — they are behaviour-preserving on real data and only fire on the malformed shapes that previously read as clean. That is the outcome I wanted from these fixes: no new noise on the happy path, no silent zero on a broken one.

Full check state at this head: 19 pass, 11 skipped, 0 failing, 0 unresolved threads, `mergeStateStatus: CLEAN`, and mega-linter passed without pushing (so the head is stable).

devantler:



**CodeRabbit lane at `91bdac55`: `cr:no-gate` again** — second time on this PR. Zero review objects at that head; the response was another `auto-generated reply by CodeRabbit`, and its own footer explains why: *"CodeRabbit is an incremental review system and does not re-review already reviewed commits."*

**But the content was right, and it is fixed in `9986fa74`.** Not counting it as the gate does not mean ignoring it — the claim was that rule metadata is looked up from a table flattened across runs, so a duplicate rule id resolves to the last definition. Reproduced before changing anything:

```
two runs, both defining C-0001, different metadata:
  before →  Kubescape: 2 finding(s) across 1 control(s).
              C-0001  note  SECOND-RUN-DESC  — 2 finding(s)     ← first run's finding, second run's metadata
  after  →  Kubescape: 2 finding(s) across 2 control(s).
              C-0001  error  FIRST-RUN-DESC   — 1 finding(s)
              C-0001  note   SECOND-RUN-DESC  — 1 finding(s)
```

`from_entries` keeps the last value for a duplicate key, so this was a confident wrong attribution — the same failure mode the `ruleIndex` bounds check exists to prevent, and worse than an absent one. Each result now resolves against its own run and carries the answer forward; grouping keys on the resolved triple.

**The other half matters as much:** runs that *agree* must still merge, or every ordinary multi-run report fragments. Pinned both directions —

| fixture | expected |
|---|---|
| two runs, conflicting metadata | `2 finding(s) across 2 control(s)`, each keeping its own run's level/description |
| two runs, identical metadata | `2 finding(s) across 1 control(s)` |

2 ablations, each RED on its own assertion: restoring the flattened table fails the conflicting-runs case, and grouping on `.id` alone fails it too. Also corrected the now-stale header comment that still described a single id-indexed table.

Advancing to Codex for the gate.

devantler:

Valid, and 🟠 Major is the right severity — fixed in `4b8aaafd`. Measured before changing anything:

```
GITHUB_STEP_SUMMARY=<sentinel> bash scripts/tests/test-summarize-sarif-findings.sh
  → 2772 bytes, 12 "### Kubescape findings" blocks, including:
      Kubescape: 4 finding(s) across 2 control(s).
      Kubescape: 2 finding(s) across 2 control(s).
```

`GITHUB_STEP_SUMMARY` is set for every Actions step and inherited by the child, so under CI every fixture run was appending to the **real job summary**. The reason this is more than tidiness: `Kubescape: 4 finding(s) across 2 control(s)` is character-for-character what the genuine scan prints in step 20 of the same job. The test would have published **fabricated findings onto the run page of the PR whose entire purpose is a findings report you can trust** — the failure mode this PR exists to remove, reintroduced by its own test.

Fixed by clearing the variable for every fixture run (`env -u`), leaving the one case that deliberately exercises the step-summary output to set its own path.

**Guarded rather than just fixed.** The suite now points the ambient variable at a sentinel that stands in for the job summary and asserts it is empty at the end, so the isolation is proven every run instead of assumed. Ablation — restoring the inherited environment — fails with `the fixtures wrote 2772 bytes to the ambient GITHUB_STEP_SUMMARY`, i.e. the guard reproduces the exact defect you reported.

After: **0 bytes**.

One process note, since it affected how I scored your lane: this review object landed at 21:33:17Z for `91bdac55`, about ten minutes after the chat-surface reply for the same head. I had recorded `cr:no-gate` at 21:23 on the strength of zero review objects at that point. The finding was still worth acting on, and the head has since moved twice — so the gate still needs a review at the current head `4b8aaafd`.

devantler:



@coderabbitai full review

Using `full review` rather than `review`: the previous request at `91bdac55` returned a chat reply whose footer said CodeRabbit "does not re-review already reviewed commits", and the branch has since gained two commits — the per-run rule-metadata fix and the step-summary isolation fix for your 🟠 Major.

devantler:



**CodeRabbit lane: `cr:no-gate@4b8aaafd`** — this one on a genuine failure signal, unlike my earlier premature call on `91bdac55`. The summary comment carries the `rate limited by coderabbit.ai` marker and a **42-minute** window, with 0 review objects at this head.

Worth stating the distinction, because I got it wrong once on this PR: at `91bdac55` I recorded `no-gate` 41 seconds after the request because a chat reply had arrived and no review object had. That was premature — CodeRabbit's real review object landed ~10 minutes later carrying the 🟠 Major about the fixtures writing to the job summary, which was well worth having. A chat reply is not a terminal state; a quota marker is.

42 minutes exceeds what this run can hold, so advancing to Codex rather than waiting it out.

devantler:



@codex review

devantler:

**Readiness at `4b8aaafd` — all three conditions, evidenced.**

**1. Programmatically tested.** `scripts/tests/test-summarize-sarif-findings.sh` pins the report end to end, and every property was proven non-vacuous by ablation — **20 across the five rounds, each RED on its own assertion**, not merely red.

**2. Reviewed.** `green_review=codex@4b8aaafdf7` — `Codex Review: Didn't find any major issues`, reviewed commit matching `headRefOid[0:10]`. CodeRabbit recorded `cr:no-gate` on a genuine 42-minute quota marker with zero review objects at head.

Both inline comments still anchored to this head are **already-resolved** CodeRabbit findings that GitHub re-anchored when the head moved (created 20:29 and 21:33, before this commit existed). All 7 review threads are resolved; nothing is outstanding.

**3. Tried and evaluated as a user** — run [30401273612](https://github.com/devantler-tech/platform/actions/runs/30401273612):

| step | result | observed |
|---|---|---|
| 11. `📋 Validate Kubescape finding summary` | success | `PASS: test-summarize-sarif-findings.sh` |
| 20. `📋 Summarize Kubescape findings` | success | `Kubescape: 4 finding(s) across 2 control(s).` |
| job log, fixture-zero lines | **0** | confirms the step-summary isolation fix — the fixtures no longer emit into the real job summary |

The middle row is unchanged from `2486ebd2`, `f45f1637` and `91bdac55`: five rounds of hardening altered nothing about this repository's actual scan output. The strictness only fires on the malformed shapes that previously read as clean.

**Nine findings were raised across five rounds and every one was real** — three of them defects in this PR's own premise (a broken scan reported as clean, locationless findings dropped, the summary skipped exactly when normalization failed), two mis-attribution bugs where a confident wrong answer beats an absent one, and the 🟠 Major where the test itself was publishing fabricated findings onto the run page. Promoting and merging.

Issue #2447, opened 2026-07-04
roadmap: Kubescape security stack → 100% and hold (posture · CVE · runtime)


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

A live investigation on 2026-07-04 found that **all three Kubescape surfaces are broken or invisible — not clean.** Every "0 findings" reading is *absence of data*, not health, and none of it reaches a human or the issue backlog, so it has aged silently for ~a week.

| Surface | State | Root cause (verified against live prod) |
|---|---|---|
| **Posture** (config scan) | 🔴 Broken | `kubescape-volume` (empty `emptyDir`) mounts at `…/.kubescape/config.json` with `subPath: config.json` → kubelet materialises `config.json` as a **directory** → scanner aborts every run. Live log: `open …/config.json: is a directory`. **1542/1542** scan objects `controls: null`; compliance **0.00 across all 9 frameworks**; data ~7 days stale. |
| **CVEs** (kubevuln) | 🔴 Broken/invisible | The relevancy scan path (`ScanCP`) is the only one wired and **aborts on partial ApplicationProfiles** ("workload restart required", ~815×/day). Verified: **197/197** image manifests have 0 grype matches / no scanner name; **109/109** summaries `.all`=0 *and* `.relevant`=0. The cluster is **not** 0-CVE — nothing is being scanned. |
| **Runtime detection** (node-agent) | 🟠 Invisible | `alertManagerExporterUrls: []`, `prometheusExporterEnabled: false`, `stdoutExporter: true` → every runtime threat alert goes to **stdout and vanishes**. Malware detection off, **0** RuntimeRuleAlertBindings, **0** alerts routed anywhere in 24h. Node-agent is healthy and learning (288 profiles, 123 still partial). |

Secondary posture degraders (independent of the mount): the scanner SA cannot `list` `validatingadmissionpolicies` or istio `VirtualService` at cluster scope (forbidden), and cloud-provider controls are N/A on Talos.

**Meta-problem:** the daily engineer is issue-driven, but its survey is GitHub-only (`portfolio-surveyor` is `kubectl`-free by design), so **no live cluster finding ever enters the backlog** unless a human files it. Combined with a CI gate that is a *floor (85), not a 100 target*, and no fix-vs-except ladder / security definition-of-done, the whole stack could rot to 0.00 unnoticed — which is what happened.

## Goal

[...]

PR #2851, opened 2026-07-28
fix(kubescape): generate the Headlamp exception mirror from the CRs

### LINKED ISSUE #2586: fix(kubescape): headlamp-exceptions ConfigMap has drifted from the ClusterSecurityException CRs

## Problem

The `headlamp-exceptions` ConfigMap (`k8s/bases/infrastructure/controllers/kubescape/config-map-headlamp-exceptions.yaml`) is a hand-maintained JSON mirror of the `ClusterSecurityException` CRs, and it has drifted badly:

- 4 CSEs are missing entirely from the mirror: `kubescape-privileged`, `readonly-rootfs-pending`, `talos-cis-control-plane-false-positives`, `talos-cis-worker-false-positives`.
- Resource scopes diverge: `delete-capable-rbac-by-design` has 26 matchers in the CRs but 3 in the mirror; `batch-workloads` 2 vs 1; `pod-security-mutations` control counts differ (4 vs 5).

The Headlamp Kubescape plugin therefore shows a different (stale) exception set than the operator and CI actually apply — the UI mis-reports posture.

## Proposed direction

Stop hand-maintaining the mirror: `scripts/generate-kubescape-exceptions.py` (added by #2584) already derives the exact `PostureExceptionPolicy` JSON from the CSE CRs. Either regenerate the ConfigMap's `exceptionPolicies` key from that output (with a CI drift check comparing the two), or — better long-term — drop the mirror once the Headlamp plugin can read the CRDs directly (upstream capability to verify). Reason strings may need light copy-editing for UI brevity, which the generator's output can carry as-is.

## Rough size

S–M — one script/CI hookup plus a regenerated ConfigMap; the compare-and-fail drift gate is the part worth doing carefully.

Part of #2447.

devantler:

## Why

The Headlamp dashboard is how we read the platform's security posture, and its exception list was a
hand-maintained copy of the `ClusterSecurityException` CRs. Copies drift — and this one drifted in the
direction that **hides findings**: until #2837 it excepted "immutable container filesystem" for every
workload in the cluster, which the CRs never granted. #2837 corrected the contents by hand, but left
the mirror hand-maintained, so nothing stops it drifting again.

## What

The mirror is now **generated from the CRs**, and a test regenerates it and fails if the committed file
differs — so a CR change that is not mirrored breaks the build instead of silently desynchronising the
dashboard from the exceptions actually in force. The two host/CIS exceptions that must stay out of the
mirror now say so on the CR itself, rather than relying on someone remembering the convention.

The exception set the dashboard shows is unchanged by this PR, and the security scan that gates merges
receives byte-identical input to before.

Fixes #2586
Part of #2447

### DIFF
--- a/k8s/bases/infrastructure/cluster-security-exceptions/talos-cis-control-plane-false-positives.yaml
+++ b/k8s/bases/infrastructure/cluster-security-exceptions/talos-cis-control-plane-false-positives.yaml
@@ -28,6 +28,15 @@ apiVersion: kubescape.io/v1beta1
 kind: ClusterSecurityException
 metadata:
   name: talos-cis-control-plane-false-positives
+  annotations:
+    # Keep this exception out of the Headlamp mirror ConfigMap. The plugin only
+    # evaluates WORKLOAD posture scans, and every control below is a host-scanner
+    # finding on the control-plane node (file permissions/ownership, API-server
+    # flags) that never attaches to a workload — so it cannot be expressed as a
+    # workload exception. Without this marker the generator would emit a
+    # cluster-wide `kind: ".*"` designator for it, which the plugin would apply
+    # to every workload in the cluster.
+    platform.devantler.tech/headlamp-mirror: exclude
 spec:
   reason: >-
     CIS control-plane checks that are false positives on Talos Linux, matching
--- a/k8s/bases/infrastructure/cluster-security-exceptions/talos-cis-worker-false-positives.yaml
+++ b/k8s/bases/infrastructure/cluster-security-exceptions/talos-cis-worker-false-positives.yaml
@@ -31,6 +31,15 @@ apiVersion: kubescape.io/v1beta1
 kind: ClusterSecurityException
 metadata:
   name: talos-cis-worker-false-positives
+  annotations:
+    # Keep this exception out of the Headlamp mirror ConfigMap. The plugin only
+    # evaluates WORKLOAD posture scans, and every control below is a host-scanner
+    # finding on the worker node (kubelet config file permissions/ownership and
+    # kubelet flags) that never attaches to a workload — so it cannot be
+    # expressed as a workload exception. Without this marker the generator would
+    # emit a cluster-wide `kind: ".*"` designator for it, which the plugin would
+    # apply to every workload in the cluster.
+    platform.devantler.tech/headlamp-mirror: exclude
 spec:
   reason: >-
     CIS worker-node checks that are false positives on Talos Linux, matching Sidero's
--- a/k8s/bases/infrastructure/controllers/kubescape/config-map-headlamp-exceptions.yaml
+++ b/k8s/bases/infrastructure/controllers/kubescape/config-map-headlamp-exceptions.yaml
@@ -1,43 +1,56 @@
 ---
-# Mirror of the platform's kubescape ClusterSecurityException CRs
-# (k8s/bases/infrastructure/cluster-security-exceptions/) into the format the
-# Headlamp Kubescape plugin reads.
+# GENERATED FILE — DO NOT EDIT BY HAND.
 #
-# WHY: the pinned plugin (headlamp_kubescape v0.11.2) has its OWN exception
-# mechanism — ConfigMaps labelled `kubescape.io/custom-object: exceptions` — and
-# does NOT read the kubescape.io ClusterSecurityException/SecurityException CRDs.
+# Regenerate with, from the repository root:
+#   go run ./scripts/generate-kubescape-exceptions \
+#     -format headlamp-configmap \
+#     -o k8s/bases/infrastructure/controllers/kubescape/config-map-headlamp-exceptions.yaml
+#
+# CI regenerates this file and fails on any difference, so a hand edit is
+# reverted by the next run rather than silently kept. Change the
+# ClusterSecurityException CRs in k8s/bases/infrastructure/cluster-security-exceptions/
+# instead — they are the single source of truth — or the template at
+# scripts/generate-kubescape-exceptions/headlamp-configmap.yaml.tmpl for this
+# prose.
+#
+# WHY THIS FILE EXISTS: the pinned plugin (headlamp_kubescape v0.11.2) has its OWN
+# exception mechanism — ConfigMaps labelled `kubescape.io/custom-object: exceptions` —
+# and does NOT read the kubescape.io ClusterSecurityException/SecurityException CRDs.
 # (The CRD-native "Security Exceptions" view exists only on the plugin's unreleased
 # main branch.) Without this ConfigMap the plugin's Exceptions page shows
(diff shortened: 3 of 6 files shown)


devantler:

## Verification record at `4b92c32b`

Detail kept out of the PR body per the PM-level convention.

### The claim that needed proving

The regenerated mirror is written in the generator's canonical form, which differs *textually* from the
hand-written one in 10 of 19 policies — alternation-collapsed resource lists (`^(Job|CronJob)$` vs two
designators) and `namespace: ".*"` vs `kind: ".*"` for an unscoped CR. So "the exception set is
unchanged" is a behavioural claim, not a diff observation.

**Differential probe.** Reimplemented the plugin's matcher (`apply-exceptions.ts`: a control is
excepted when a `posturePolicies` regex matches the controlID **and** a `resources` designator matches
the workload) and ran old-vs-new over, per policy, every literal either form mentions plus a sentinel
neither mentions — a regex cannot distinguish values it never names, so that domain is complete.

```
total_probes: 414,845    total_divergences: 0
```

**The probe is not vacuous.** Control: injected the exact defect #2837 fixed — `C-0017` added to
`pod-security-mutations`, which under that policy's cluster-wide designator excepts it for every
workload. The probe flagged it, naming `C-0017` as the diverging control. So it detects precisely the
class of change that matters.

### Live cluster (read-only, `admin@prod`)

- The deployed `headlamp-exceptions` ConfigMap is **byte-identical to the old committed file** — Flux
  is in sync and there was no out-of-band drift to preserve.
- All 19 live control sets are preserved in the regenerated file.

### The CI scan is untouched

`go run ./scripts/generate-kubescape-exceptions -o …` output is **byte-identical** before and after, so
the annotation and the new format add nothing to the compliance gate's input. The excluded host/CIS
exceptions are still in that output — they are excluded from the *mirror* only, which the
`TestMirrorExcludesAnnotatedException` assertion pins in both directions.

### Ablations — 5, each RED with the intended scope

| Ablation | Test that went RED |
|---|---|
| `mirrored()` stops filtering | `TestMirrorExcludesAnnotatedException` **and** `TestCommittedMirrorIsUpToDate` |
| annotation value validation removed | `TestMirrorAnnotationFailsClosed` |
| empty-mirror guard removed | `TestMirrorRefusesEmptyResult` |
| **hand-edit the committed mirror** (C-0017 injected) | `TestCommittedMirrorIsUpToDate` — the gate fires |
| malformed-annotations guard reverted to fail-open | `TestMirrorMalformedAnnotationsFailClosed` RED while `TestMirrorEmptyAnnotationsAreNotAMarker` stayed PASS, so the guard discriminates rather than over-rejecting |

The drift gate's failure message names the regeneration command, so the fix is in the failure rather
than in someone's memory.

### Found by self-review, before requesting a review

`resolveMirrorExclusion` originally read a **present-but-not-a-mapping** `metadata.annotations` as "no
marker". That fails **open**: a malformed annotations block on a `talos-cis-*` CR would silently drop
the exclusion, mirror the host exception, and — via its cluster-wide designator — except that control
for every workload. Exactly the widening this PR exists to prevent. Fixed in `4b92c32b` with the
discriminating ablation above.

### Also checked

- `ksail workload validate` on **both** overlays: 555 files each, 0 errors.
- `kubectl kustomize` of the kubescape base builds; the ConfigMap survives with 19 policies and its
  `kubescape.io/custom-object: exceptions` selector label intact.
- Generation is idempotent (regenerated twice, byte-identical, matches the committed file).
- `golangci-lint`: 0 issues.

### Not verified

I did not open the Headlamp UI. The plugin-side rendering is asserted through its documented matching
semantics and the probe above, not observed in a browser.

devantler:



CI is fully settled at this head: 24 checks green, 0 failures, and mega-linter completed without
pushing an auto-commit, so this green will not stale under the review.

@coderabbitai review

devantler:



Re-triggering: the previous request was rejected under CodeRabbit's **chat-message** hourly limit
(stated window 1m18s), not its review limit, so no review was started at this head.

@coderabbitai review

devantler:




**CodeRabbit did not deliver at this head — recording `cr:no-gate@4b92c32b99` and advancing to Codex.**

Two separate refusals, neither of which started a review:
1. the first trigger was rejected under the **chat-message** hourly limit (window 1m18s);
2. the retrigger hit the **PR review limit** — *"you've reached your PR review limit, so we couldn't
   start this review"*, next review in 8 minutes.

Worth stating precisely, because this head has a **near-miss green**: CodeRabbit's 21:15:23Z comment
does contain the `between <base> and <head>` scope line naming exactly `4b92c32b9939…`. It is **not** a
green — that line sits *inside* the rate-limited block, and the same comment carries the
`rate limited by coderabbit.ai` marker. There are **0** CodeRabbit review objects on this PR. A gate
keyed on the scope line alone would have read this as reviewed.

State at this head: 24 checks green / 0 failures, `CLEAN`, 0 unresolved threads, mega-linter finished
without an auto-commit.

@codex review

devantler:

## Readiness — all three conditions met at `4b92c32b9939ca050a292482b6da115d269f9e5b`

**1. Programmatically tested.** 24 checks green, 0 failing, 0 running. Full hygiene pentad clear:
green CI · 0 unresolved threads · 0 non-thread review findings · `CLEAN` against base · green review
at head. 14 Go tests pass, `golangci-lint` 0 issues, `ksail workload validate` green on **both**
overlays (555 files each). **5 ablations, each RED with the intended scope** — including the one that
matters most, hand-editing the committed mirror to inject `C-0017` cluster-wide (the real #2837
defect), which the drift test catches and answers with the regeneration command.

**2. Reviewed.** **Codex green at this exact head** — `Reviewed commit: 4b92c32b99`, "Didn't find any
major issues", 0 review objects. One successful lane satisfies the gate, so CodeRabbit was not
re-requested. CodeRabbit's own outcome at this head is recorded above as `cr:no-gate` (two refusals,
neither started a review; note the near-miss scope line inside its rate-limit block).

**3. Tried and evaluated as a user.** Exercised, not reasoned about:
- Ran the generator; the mirror renders through `kubectl kustomize` with 19 policies and its
  `kubescape.io/custom-object: exceptions` selector label intact — i.e. the plugin still finds it.
- **414,845-probe differential** over the plugin's own matching rule, old form vs new: **0
  divergences**; the control (inject `C-0017`) fires and names `C-0017`, so the probe is not vacuous.
- **Live prod read:** the deployed ConfigMap is byte-identical to the pre-change committed file, and
  all 19 live control sets survive regeneration.
- The Kubescape scan input is **byte-identical** before and after, so the compliance gate that guards
  merges is provably untouched.
- Generation is idempotent; the gate is triggered by all three break-paths (CR edit, ConfigMap edit,
  generator/template edit) via the `k8s` path filter.

Promoting and enqueuing. Platform gates `main` behind a merge queue, so this enqueues rather than
merges directly; `deploy-prod` runs on `merge_group`.

Issue #2651, opened 2026-07-15
Activate and verify Headlamp user namespaces


devantler:
> 🤖 Generated by the Daily AI Engineer (Codex instance)

## Evidence

Platform issue #1807 requires each stateful workload to graduate through an explicit user-namespace pilot. Headlamp is the next pilot after Longhorn, but activation must stay separate from the default-off gate in #2650 so the rollout can happen only after the concurrent node-auth changes in #2635 have settled and live storage health is known-good.

Headlamp persists plugins on the `headlamp-plugins` PVC. Its chart renders `hostUsers: true` by default, so activation must explicitly set `spec.values.hostUsers: false` and preserve the existing volume identity throughout the experiment.

## Impact

This is the live verification step for Headlamp's two-container pod. A failed rollout could prevent the plugin sidecar from writing files, prevent the main container from reading those files, or disturb the Longhorn-backed plugin volume. Recovery must therefore revert only the user-namespace value and must never delete the PVC.

## Acceptance criteria

- [ ] #2650 is merged, #2635 has settled, every production node is `Ready`, Longhorn is healthy, and Headlamp is stable before activation.
- [ ] Record the existing Headlamp PVC, PV, and Longhorn volume identities before changing the gate.
- [ ] Enable the exact-target Headlamp values patch so the rendered Deployment has `spec.template.spec.hostUsers: false`.
- [ ] Verify the Deployment is `Available` 1/1 and the pod is 2/2 `Ready`.
- [ ] Verify both containers have non-zero subordinate UID/GID mappings and the plugin installer logs contain no `EACCES` or `EPERM` errors.
- [ ] Verify installed plugin files are visible from the main Headlamp container, and the route, OIDC login, and plugins still work.
- [ ] Verify the same PVC/PV/Longhorn volume remains bound and healthy, with no mount or warning events.
- [ ] Document the observation window and evidence on the issue.
- [ ] If recovery is needed, re-disable the gate, reconcile to healthy, and retain the existing PVC.

Part of #1807.

devantler:
> 🤖 Generated by the Agentic Engineer

## Pre-activation evidence — acceptance criteria 1 and 2

Activation is up as **#2842** (draft). Recording the before-state here, because criterion 2 is precisely the set of identities a recovery must find unchanged.

### Criterion 1 — the gate chain, re-verified live

Every part re-checked against the cluster today rather than inherited from a note:

[...]

PR #2857, opened 2026-07-29
test(headlamp): measure the user-namespace mapping in prod

devantler:

## Why

Headlamp's user-namespace pilot (#2651) is verified on eight of its nine acceptance criteria. The one left open asks that the containers hold a **non-zero subordinate UID/GID mapping** — and it has stayed unmeasured since activation, because reading that mapping from inside Headlamp needs `exec`, which nobody should hold standing access to just to check a box.

So the pilot has been running in a state we believe is correct but have not measured. This adds the means to measure it, without granting anyone new access.

## What

A disposable Job that reads its own ID maps and exits. It runs **in the headlamp namespace, as headlamp's UID/GID, on headlamp's node** — `hostUsers` is a pod-level field, so the real question is whether this kubelet honours it at all, and a pod asking for the same thing in the same place answers exactly that.

It ships **default-off**, staged behind a short-lived activation PR like the `userns-longhorn-smoke` component beside it. Self-review is what put it there: the apps Flux Kustomization runs `wait: true` with no explicit healthChecks, so a Job reporting `Failed` takes the whole apps layer not-Ready and can fail the merge-queue deploy — and a negative result is precisely what this probe is built to produce. A probe must not be able to wedge the layer it is measuring, so activation is a deliberate, watched act rather than a standing one.

That hazard applies to **any** Job added to this overlay, not just this one, so the comment now records it — the existing component's comment attributed its staging only to its Longhorn volume.

## Operational note

Merge order is: this lands inert → a short-lived PR uncomments it and the deploy runs it → its logs go on #2651 → #2858 removes the component. #2651 does not close on this PR.

Part of #2651

### DIFF
--- a/k8s/providers/hetzner/apps/kustomization.yaml
+++ b/k8s/providers/hetzner/apps/kustomization.yaml
@@ -20,14 +20,18 @@ resources:
   # capture the completed Job's logs, then remove it again so its generic
   # ephemeral PVC and Longhorn volume are garbage-collected.
   # - userns-longhorn-smoke/
-  # Disposable measurement of the subordinate UID/GID mapping a user-namespaced
-  # pod receives in the headlamp namespace (#2651, acceptance criterion 5).
-  # Unlike userns-longhorn-smoke above this one ships ACTIVE: that component is
-  # staged behind a second PR so its generic ephemeral PVC and Longhorn volume
-  # are garbage-collected promptly, and this Job has no volumes at all — it
-  # reads /proc/self and exits, capped by backoffLimit: 0 and
-  # activeDeadlineSeconds. Remove it again once its logs are recorded on #2651.
-  - userns-headlamp-mapping-probe/
+  # Default-off disposable measurement of the subordinate UID/GID mapping a
+  # user-namespaced pod receives in the headlamp namespace (#2651, acceptance
+  # criterion 5). Staged exactly like userns-longhorn-smoke above, and for a
+  # reason that applies to EVERY Job reconciled here: this Flux Kustomization
+  # runs `wait: true` with no explicit healthChecks, so kstatus evaluates every
+  # applied resource and a Job that reports Failed — whether from a real
+  # negative result, an unschedulable pod, or a pull error — takes the whole
+  # apps layer not-Ready and can fail the merge-queue deploy. A probe must never
+  # be able to wedge the layer it is measuring, so activation is a deliberate,
+  # watched act: uncomment this resource in a short-lived PR, capture the
+  # completed Job's logs onto #2651, then remove the component again (#2858).
+  # - userns-headlamp-mapping-probe/
   # Prod-only Coroot Postgres integration for the platform-owned CNPG DBs:
   # additive CiliumNetworkPolicies that let the observability cluster-agent
   # scrape each on 5432 (the Cluster patches below stamp the
--- a/k8s/providers/hetzner/apps/kustomization.yaml
+++ b/k8s/providers/hetzner/apps/kustomization.yaml
@@ -20,6 +20,14 @@ resources:
   # capture the completed Job's logs, then remove it again so its generic
   # ephemeral PVC and Longhorn volume are garbage-collected.
   # - userns-longhorn-smoke/
+  # Disposable measurement of the subordinate UID/GID mapping a user-namespaced
+  # pod receives in the headlamp namespace (#2651, acceptance criterion 5).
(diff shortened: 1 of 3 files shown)


devantler:

## Pre-flight evidence (measured, not assumed)

Recording what I checked before shipping, so a reviewer does not have to re-derive it and so the eventual activation is low-surprise.

**The probe's shape already works on the target node.** Nine user-namespaced pods run in prod today — `backstage`, `crossview` (×2), `headlamp`, `homepage` (×2), `umami` (×2), `whoami` — and one of them, `umami/umami-provision-tenants`, is a **Job** that ran on **prod-worker-3**, headlamp's own node, and reached `Succeeded`. So a `hostUsers: false` Job scheduling and completing on that node is demonstrated, not hoped for.

That also raises the prior on the outcome: the kubelet is evidently honouring `hostUsers` across five apps. The probe is still worth taking because **none of those pods' mappings have ever been read** — the criterion asks for the value, and "many pods asked for it and nothing broke" is the same inference the issue already declined to close on.

**Admission will not block it.** `verify-app-images` is an `ImageValidatingPolicy` whose `matchImageReferences` is scoped to `ghcr.io/devantler-tech/*`; third-party images admit unverified. The pinned `busybox` digest is **already running in this cluster**, so the pull is proven. `validate-pod-security` and `validate-host-restrictions` are both satisfied — the latter constrains `hostUsers: true` in opted-in namespaces, and this sets `false` explicitly.

**It will schedule.** prod-worker-3 is at 21% CPU / 26% memory allocated, with no taints. The `podAffinity` selector matches the live headlamp pod's `app.kubernetes.io/name` + `app.kubernetes.io/instance` labels.

**The assertions were exercised before shipping.** Four synthetic maps: a real mapped case passes; the identity map `0 0 4294967295` fails; a hypothetical non-zero-based full-width map fails; and uid-mapped/gid-identity fails, so it is not a partial guard. Ablating the length assertion flips the third case to passing while the second is still caught by the host-start assertion — the two guards are complementary rather than one shadowing the other.

**Static validation:** both overlays build, `ksail --config ksail.prod.yaml workload validate` passes 556 files, the naming validator is clean, and the Kubescape gate went green while the component was still wired in — so the manifest itself is proven against the posture scan even though it now ships staged.

devantler:



CI is green at this head (16 success, 0 failures) and the PR is one concern: a staged, default-off diagnostic Job plus the overlay comment recording why any Job here must be staged.

Worth a reviewer's attention specifically: the claim that a Job reporting `Failed` under this Kustomization's `wait: true` (no explicit `healthChecks`) takes the whole apps layer not-Ready — that is the reasoning the default-off staging rests on.

@coderabbitai review

devantler:

## Readiness — promoting

**Pentad at `323a47a73cb88bbc201b062eae4e20139751819a`:** CI green (16 success, 12 skipped, 0 failures) · 0 unresolved threads · 0 actionable review findings (CodeRabbit's newest body carries only `🔇 Additional comments (3)`, its non-actionable section) · `mergeStateStatus: CLEAN` · **`cr@323a47a7`** — summary comment posted 15:41:39Z, after the 15:37Z request, naming this exact head, with no rate-limit or service marker.

**Programmatically tested.** `ksail --config ksail.prod.yaml workload validate` passes 556 files; both cluster overlays build; the naming validator is clean; the Kubescape gate went green while the component was still wired in, so the manifest is proven against the posture scan.

The probe's assertions were RED/GREEN'd before shipping, against four maps: a real mapped case passes, the identity map fails, a non-zero-based full-width map fails, and uid-mapped/gid-identity fails. **Ablation:** removing the length assertion flips the third case to passing while the second is still caught by the host-start assertion — so the two guards are complementary, and neither is carrying the other.

## What I exercised, and what I did not

Stating this precisely, because the honest answer is split.

**Exercised:** the behaviour this PR actually delivers is *"the component exists, is well-formed, and is not applied."* I built it **both** ways and observed both — wired in, it renders into the hetzner apps overlay and passes the full CI gate including Kubescape; staged, it is absent from that overlay build (`grep -c` → 0) while still building standalone and satisfying the naming rules. It is correctly absent from the local overlay in both states.

I also verified the premise the staging rests on directly against the live cluster: the `apps` Flux Kustomization really does carry `wait: true` with an empty `healthChecks`.

**Not exercised, deliberately:** the Job has not run, so no mapping has been read. That is the point of shipping it staged — activation is a separate, watched PR. #2651 stays open until that log exists, and this PR does not claim otherwise.

Issue #2845, opened 2026-07-28
security(k8s): give each workload an explicit service-account-token verdict (CKV_K8S_38)


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`CKV_K8S_38` ("service account tokens are only mounted where necessary") is the second-largest
group in #2787's checkov backlog — **8 of the 42** Kubernetes-framework findings measured on `main`
2026-07-28:

| workload | file |
|---|---|
| `Deployment.kube-system.coredns` | `k8s/providers/docker/infrastructure/controllers/coredns/deployment.yaml` |
| `CronJob.longhorn-system.longhorn-stale-node-cleanup` | `k8s/providers/hetzner/infrastructure/controllers/longhorn/cron-job-stale-node-cleanup.yaml` |
| `Job.openbao.vault-snapshot-init` | `k8s/bases/infrastructure/vault-backup/job.yaml` |
| `CronJob.openbao.vault-snapshot` | `k8s/bases/infrastructure/vault-backup/cron-job.yaml` |
| `Job.openbao.vault-config` | `k8s/bases/infrastructure/vault-config/job.yaml` |
| `Deployment.opencost.opencost-usage-scraper` | `k8s/bases/infrastructure/opencost/components/usage-scraper/deployment.yaml` |
| `Deployment.cdi.cdi-operator` | `k8s/bases/infrastructure/controllers/cdi/cdi-operator.yaml` |
| `Deployment.kubevirt.virt-operator` | `k8s/bases/infrastructure/controllers/kubevirt/kubevirt-operator.yaml` |

## Why it matters

A mounted service-account token is a credential sitting in the container filesystem. Where the
workload never calls the API server it buys nothing and is simply reachable by anything that gets
code execution in that pod — which is the whole reason the check exists. Kubescape does not cover
this: its own exception CRs disable the pod-security controls cluster-wide, and the Kyverno
pod-security policies match `kinds: [Pod]` while `k8s/` contains no Pod objects.

## Why this is NOT a blanket sweep

At least three of the eight legitimately need their token — the KubeVirt and CDI operators are
controllers, and the Longhorn cleanup CronJob acts on nodes through the API. Setting
`automountServiceAccountToken: false` across all eight would break them. Each workload needs an
individual verdict, and the ones that keep the token need that recorded so the next pass does not
re-litigate it.

## What

Per workload, either set `automountServiceAccountToken: false` on the pod spec, or record in the
[...]

PR #2860, opened 2026-07-29
security(k8s): record an explicit service-account-token verdict per workload

devantler:

## Why

A mounted service-account token is a live credential sitting in the container
filesystem. Eight workloads in `k8s/` mount one, and until now none of them said
whether it was needed — so the scanner reported eight findings that nobody could
act on, and a *new* workload quietly mounting a token looked exactly the same as
the eight known ones.

## What

Six first-party workloads now carry an explicit verdict on the pod spec plus the
reason the token is required, recorded where the scanner reads it. All six were
checked against what the workload actually does, not assumed — each one reaches
the API server (cluster DNS, `kubectl`, or an OpenBao Kubernetes-auth login).

The result: the check goes from six undifferentiated findings to zero, with every
exception scoped to one workload and carrying its justification — and a new
workload that mounts a token without stating why still fails. No behaviour
changes; these workloads already mounted the token.

**Not covered here:** the two remaining workloads are KubeVirt's and CDI's
vendored upstream operator bundles, which are re-vendored wholesale from each
release, so anything written into them is silently lost on the next bump. They
need a mechanism that survives re-vendoring — filed separately rather than
guessed at here.

Part of #2845

### DIFF
--- a/k8s/bases/infrastructure/opencost/components/usage-scraper/deployment.yaml
+++ b/k8s/bases/infrastructure/opencost/components/usage-scraper/deployment.yaml
@@ -8,6 +8,8 @@ metadata:
     app.kubernetes.io/part-of: opencost
     app.kubernetes.io/component: metrics-scraper
     platform.devantler.tech/replica-floor: exempt
+  annotations:
+    checkov.io/skip1: CKV_K8S_38=Prometheus kubernetes_sd and authenticated kubelet scraping both need the token
 spec:
   # Two replicas would scrape identical node series and race them into the
   # remote-write receiver. Recreate avoids an overlapping rollout; the WAL is
--- a/k8s/bases/infrastructure/vault-backup/cron-job.yaml
+++ b/k8s/bases/infrastructure/vault-backup/cron-job.yaml
@@ -25,6 +25,8 @@ kind: CronJob
 metadata:
   name: vault-snapshot
   namespace: openbao
+  annotations:
+    checkov.io/skip1: CKV_K8S_38=the snapshot container reads its SA JWT to log in via OpenBao Kubernetes auth
 spec:
   schedule: "30 3 * * *"
   # Pin the cron evaluation to UTC instead of inheriting kube-controller-manager's
@@ -50,6 +52,11 @@ spec:
             app: vault-snapshot
         spec:
           serviceAccountName: vault-snapshot
+          # Required: the snapshot container reads
+          # /var/run/secrets/kubernetes.io/serviceaccount/token and presents it
+          # as the `jwt` for `auth/kubernetes/login`, so the token IS the
+          # credential.
+          automountServiceAccountToken: true
           restartPolicy: OnFailure
           # The full restricted-PSS context is set explicitly (not left to the
           # add-security-context Kyverno mutation) so the pod stays hardened
--- a/k8s/bases/infrastructure/vault-backup/cron-job.yaml
+++ b/k8s/bases/infrastructure/vault-backup/cron-job.yaml
@@ -52,11 +52,11 @@ spec:
             app: vault-snapshot
         spec:
           serviceAccountName: vault-snapshot
-          # Required: the snapshot container reads
+          # The token IS this pod's credential: the snapshot container reads
           # /var/run/secrets/kubernetes.io/serviceaccount/token and presents it
-          # as the `jwt` for `auth/kubernetes/login`, so the token IS the
-          # credential.
-          automountServiceAccountToken: true
+          # as the `jwt` for `auth/kubernetes/login`. Left at the mount-by-default
+          # value rather than set explicitly — an explicit `true` reads identically
+          # to Kubernetes and trips Kubescape C-0034.
           restartPolicy: OnFailure
           # The full restricted-PSS context is set explicitly (not left to the
           # add-security-context Kyverno mutation) so the pod stays hardened
--- a/k8s/bases/infrastructure/vault-backup/job.yaml
+++ b/k8s/bases/infrastructure/vault-backup/job.yaml
@@ -50,10 +50,11 @@ spec:
         app: vault-snapshot
     spec:
       serviceAccountName: vault-snapshot
-      # Required: the snapshot container reads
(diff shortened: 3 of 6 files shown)


devantler:

## What I exercised, and what the controls prove

**Behavioural, not reasoned.** Every claim below came from running the scanner, not
from reading its docs.

**1. The skip mechanism was chosen by measurement, not documentation.** Built a
three-case fixture and ran checkov against it:

| form | result |
|---|---|
| `checkov.io/skip1: CKV_K8S_38=<reason>` annotation | **SKIPPED**, reason recorded in the report |
| `# checkov:skip=CKV_K8S_38: <reason>` YAML comment | **still FAILED** |
| `automountServiceAccountToken: false` | PASSED |

The comment form is a silent no-op — it reads as applied and changes nothing. Had I
used the form that looks idiomatic, this PR would have shipped six annotations that
do nothing and a green claim to go with them.

**2. Target check.** `CKV_K8S_38` on `k8s/`: **8 FAILED → 2 FAILED, 6 SKIPPED**, each
skip carrying its own reason.

**3. Whole-tree regression control.** Kubernetes-framework failures **43 → 37**,
compared as sets of `check_id + resource` against a clean `main` worktree. The removed
set is exactly those six; the **added set is empty**.

**4. Ablation.** Dropped *only* the `vault-config` annotation: that resource returns to
FAILED (2 → 3) and the other five stay skipped. Restored by `cp` and verified by
SHA-256 — the annotation is load-bearing and the effect is per-resource, not global.

**5. No behaviour change.** `automountServiceAccountToken: true` makes the existing
effective default explicit. Checked that none of the ServiceAccounts involved
(`vault-snapshot`, `vault-config`, `opencost-usage-scraper`) sets
`automountServiceAccountToken: false` — a pod-level `true` would have *overridden* such
an opt-out and started mounting a token that was deliberately withheld. None does; all
six already mounted the token.

**6. Validation.** `ksail workload validate` and
`ksail --config ksail.prod.yaml workload validate` — **556 files each, exit 0**.

## Known residual, stated rather than papered over

Two workloads still fail: the vendored KubeVirt and CDI operator bundles. They are
re-vendored wholesale on every bump, so an annotation written into them is silently
dropped later — which is why they need a different mechanism and are filed as #2861
instead of being patched here.

devantler:

## Correction at `6cf7144a` — the explicit mount flag tripped a second scanner

The first head added `automountServiceAccountToken: true` to the three `openbao`
workloads, because #2845 prefers the pod-spec field over a comment ("so the decision
is visible where the workload is defined"). CI's `kubescape` check then reported
**3 new alerts** — C-0034 *Automatic mapping of service account*, at exactly the three
lines the flag was added on.

That is a real regression I introduced, and my own regression control missed it: I
diffed **checkov** failures before/after and never re-ran **Kubescape**. Two scanners
cover this tree and I only controlled for one.

### What I tried, and what the measurements actually said

**C-0034's own remediation text says to disable auto-mount on the ServiceAccount and
opt in per pod.** So I set `automountServiceAccountToken: false` on the `vault-snapshot`
and `vault-config` ServiceAccounts and re-scanned. **It cleared nothing** — C-0034's
failing set was byte-identical before and after (7 resources, none cleared, none added).
Worth stating plainly because it is the fix the tool itself recommends, and it does not
work here. It is also *unsafe* on its own: SA-level `false` with no pod-level `true`
would leave these pods with no token at all, which breaks the OpenBao Kubernetes-auth
login they depend on. The pair is all-or-nothing, so I reverted it.

**Limit of my local evidence, stated rather than glossed:** CI scans with an exceptions
file generated from the `ClusterSecurityException` CRs
(`ksail workload scan --exceptions …`); my local runs had none, so the two sets are not
comparable. Locally the three `openbao` workloads fail C-0034 *with and without* the
flag — which does not match CI calling them new. `openbao` is genuinely absent from the
C-0034 exception list in
`k8s/bases/infrastructure/cluster-security-exceptions/service-account-tokens.yaml`
(which does cover `kube-system`, `kubevirt`, `cdi`, `longhorn-system`, `opencost`), so
I could not settle from here whether the alerts are caused by the flag or are
pre-existing findings re-fingerprinted by the changed line numbers. **CI is the
authority on that, so this push is the test.**

### What changed

The flag is gone; the verdict stays, in the comment. No behavioural difference either
way — an explicit `true` is exactly the value these pods already got by default, so it
was documentation competing with a scanner rather than a control.

`coredns` keeps its explicit flag: it is in `kube-system`, which the C-0034 exception
covers, and it did not appear in any scan's failing set.

**Unchanged by this push:** `CKV_K8S_38` is still 6 SKIPPED / 2 FAILED — the checkov
disposition rides on the annotation, not the flag. Re-measured after the revert.

If CI still reports the alerts at this head, the cause is the line-shift fingerprinting
rather than the flag, and the honest fix is a scoped `openbao` C-0034 exception in the
file above — which I would rather do deliberately than fold into this PR.

devantler:



@coderabbitai review

devantler:

## Readiness — promoting

**Pentad at `6cf7144a`:** checks 17 success / 12 skipped / **0 failing** · unresolved threads **0** · non-thread review findings **0** · `mergeStateStatus` **CLEAN** · green review **`cr@6cf7144af3`** (summary comment naming this exact head, posted after the 16:45:10Z request, no quota marker, six files processed — "No actionable comments were generated").

**Tried and evaluated as a user.** Exercised the real scanners and read their output, rather than reasoning from docs — detail in the comment above. The two that decided the shape of this PR:

- `CKV_K8S_38` on `k8s/`: **8 FAILED → 2 FAILED + 6 SKIPPED**, whole-tree control 43 → 37 with an **empty added-failure set**, and an ablation that returns exactly one resource to FAILED.
- `kubescape` at this head: **"No new alerts in code changed by this pull request."** This is the one that corrected me — the first head's explicit `automountServiceAccountToken: true` produced 3 new C-0034 alerts, and CI is what settled that the flag caused them, since my local scan had no exceptions file and could not tell.

`ksail workload validate` passes on both overlays (556 files, exit 0).

The residual is stated, not hidden: two vendored operator bundles still fail `CKV_K8S_38`, tracked in #2861.

PR #2866, opened 2026-08-01
fix(security): restrict doggy-countdown route hostname

devantler:
### Motivation

- Prevent the remotely-sourced `doggy-countdown` tenant from claiming arbitrary hostnames on the shared platform Gateway, which combined with `external-dns` could publish attacker-controlled names in the platform zone. 
- The change enforces a hostname constraint for the tenant identity that Flux currently reconciles from a public OCI artifact so a compromised publishing path cannot create phishing endpoints under trusted platform hostnames.

### Description

- Add a Kyverno ClusterPolicy `restrict-doggy-countdown-hostnames` that enforces `HTTPRoute` objects in namespace `doggy-countdown` to declare exactly one hostname: `simba.${domain}`, and denies hostname-less, mixed, or other hostnames. 
- Register the new policy in `k8s/bases/infrastructure/cluster-policies/kustomization.yaml` so it is applied with the platform cluster policies. 
- Add offline Kyverno regression fixtures under `tests/restrict-doggy-countdown-hostnames/` including `resources.yaml` (good and exploit variants) and `kyverno-test.yaml` to assert the intended hostname is allowed while exploit variants are rejected. 

### Testing

- Ran `python3 scripts/validate-naming.py` which passed and reported naming conventions satisfied. 
- Ran `python3 scripts/validate-embedded-json.py` which parsed embedded JSON blobs cleanly. 
- Ran the Kyverno CLI test suite (`kyverno test ./tests`, Kyverno v1.18.2) which executed the new fixtures plus existing ones and reported all tests passed (17 tests). 
- `ksail workload validate` could not be executed in the environment because KSail is not installed, so KSail-backed static validation was not run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a664953c6dc832ba6e6db04dae577e3)


devantler:

Picked this up as a long-idle actionable PR (last push 2026-08-01, ~14 days; no active-work signal on
any surface) and cleared its **DIRTY** merge state.

**Conflict — one file, resolved as a union.** `cluster-policies/kustomization.yaml`: this branch adds
`restrict-doggy-countdown-hostnames.yaml`, and `main` has since added
`restrict-tenant-issuer-refs.yaml` in the adjacent line. Both entries belong, so both are kept in
alphabetical position.

Taking either side alone would have been wrong, and the incoming side is the dangerous one to drop:
it is itself a merged security fix (#2833, *"enforce the tenant ClusterIssuer boundary at
admission"*), so resolving in favour of this branch would have silently reverted an admission policy
already live on `main`.

**Verified:** both referenced policy files exist, no conflict markers remain, the policy base builds
(1848 lines), and **both** policies appear in the built output.

No behaviour on this branch changed — the merge only integrates `main`. CI is re-running at the new
head, and the review gate still applies.

devantler:

**CI now runs 33 checks instead of 5, and one fails. Root-caused — the 36 "unresolved Flux
substitution" lines are diagnostic noise, not the failure.**

Bringing the branch up to date widened the changed-path set, so `🔐 Validate EKS Authorization` runs
for the first time on this PR. It fails, and the log's 36 substitution lines are misleading: the
validator states in `main.go` that they are *"DIAGNOSTIC, not a control"*, emitted only to explain a
hash that moved. Promoting them to errors was measured on 2026-07-21 to fail the approved tree on
thirty-plus HelmReleases, because post-build substitution is the platform's normal mechanism.

**The actual error is one line:**

```
unapproved rendered authorization surface fingerprint:
e02856de56721b06897d6c151974796b08ecd73c23b461edde4c4f1efb44636b
```

**Why this PR moves it.** `restrict-doggy-countdown-hostnames.yaml` is authorization-capable and
contains `${domain}` (lines 36 and 43). A resource carrying an unresolved substitution is deliberately
forced into the aggregate authorization surface so its `${…}` literal is covered by the fingerprint —
so adding this policy legitimately changes surface membership, and the approved constant
(`expectedRenderedSurfaceSHA`, `main.go:366`) must move with it.

**Confirmed it is this PR and not `main`:** the same test passes on a tree that is `main` plus a
comment-only change, and fails here. This branch differs from `main` by only the four files it adds.

**What is needed — and why I have not just pasted the new hash.** That constant's own comment sets the
approval bar: measure the rendered membership by `apiVersion|kind|namespace|name` set difference
across the production render trees, confirm exactly which entries move, and record that in the
rationale beside the value. Approving a security fingerprint without that measurement would defeat the
control. The delta here is expected to be exactly the one new `ClusterPolicy`, but *expected* is not
*measured*.

So this is the remaining blocker, and it is a single well-scoped step: run that membership diff, and if
it shows only the new policy, update `expectedRenderedSurfaceSHA` to the value above with the
measurement recorded alongside it.

Separately: the **DIRTY** merge state is cleared (see the previous comment) and every other check is
green — 0 failures across the other 31.

devantler:

**`expectedRenderedSurfaceSHA` updated to `e02856de…` — the only blocker on this PR.**

The new Kyverno `ClusterPolicy` is selected into the fingerprinted authorization surface (Kyverno ClusterPolicies count as indirect authorization policies), so the aggregate value had to move. Approving it deliberately rather than mechanically:

**Membership** — measured against main `44f14fab`, rendering all five production roots on both sides: **524 documents on main, 525 here**, differing by set difference in both directions over the complete `apiVersion|kind|namespace|name` identity by exactly one entry:

```
kyverno.io/v1  ClusterPolicy  restrict-doggy-countdown-hostnames
```

Zero removed, zero renamed. After removing just that entry the remaining **524 documents are byte-identical** to main's render — checked with a negative control that fired on a one-line perturbation, so the identical result is not vacuous. Nothing added, removed or renamed touches a Role, ClusterRole, RoleBinding, ClusterRoleBinding, ServiceAccount or IAM document.

**Direction of the change** — the added policy is a single `validate.deny` rule matching `HTTPRoute` in the `doggy-countdown` namespace and requiring exactly the tenant's assigned hostname, with `validationFailureAction: Enforce`. It creates no identity, binding, role or policy document and grants nothing to the `aws/aws` service account; it can only refuse admission, so it moves the surface strictly in the restrictive direction.

**Corroboration** — two independent renderers agree on the value: the required CI job on the approved toolchain, and a local render on kubectl v1.36.1 with the same embedded kustomize v5.8.1. Agreement across a kubectl patch difference is what rules out a renderer-version artifact.

One trap recorded in the code comment while verifying this: `TestValidateAuthorizationAcceptsCommittedPolicy` calls `validateAuthorization` directly and so never passes through `validateRendererVersion` — only `run()` does. A passing `go test` is therefore **not** evidence that the local toolchain matches the pin.

Ablation on the approved constant: flipping one hex digit fails the test; restoring it passes.

devantler:



`🔐 Validate EKS Authorization` now passes at this head — the surface fingerprint was approved with a full membership and byte-identity measurement (see the comment above). Checks green, no unresolved threads, `mergeStateStatus: CLEAN`. No review has been requested at this head.

@coderabbitai review

devantler:

**Design collision with #2740 — flagging rather than acting, since both approaches are defensible and
only one should land.**

#2740 was opened to close the same class of gap and carried an open [P1] asking for the boundary to
cover *every* `tenant-edit` reconciler namespace, not one tenant. I resolved that P1 in `ed8dee15` by
generalising the existing `restrict-wedding-app-route-hostnames` policy into
`restrict-tenant-route-hostnames`: one policy with a per-tenant allow-list rule for `wedding-app`,
`ascoachingogvaner` **and** `doggy-countdown`, plus a catch-all that denies HTTPRoutes in any other
ksail tenant namespace so a newly onboarded tenant fails closed.

That overlaps this PR, which adds a separate `restrict-doggy-countdown-hostnames` policy for the third
tenant. If both merge, two `Enforce` policies govern `doggy-countdown` routes and a route must satisfy
both.

**One concrete incompatibility, worth checking either way.** This PR allows `simba.${domain}`, which
resolves per-cluster; #2740 allows the literal `simba.platform.devantler.tech`. The tenant's own
`deploy/httproute.yaml` declares exactly one hostname — `simba.platform.devantler.tech` — with a
comment tying it to external-dns's `devantler.tech` domain filter, and no `.lan` variant. So on a
local cluster `${domain}` resolves to `platform.lan` and this policy would deny the tenant's actual
route. That is latent today because apps are opt-in locally, but it is a real difference in behaviour
rather than a stylistic one.

**Suggestion, not a decision:** consolidate on the single generalised policy and close this PR, since
one policy covering all tenants plus a fail-closed catch-all is what keeps the *next* tenant from
being missed — which is exactly how `ascoachingogvaner` and `doggy-countdown` came to be uncovered.
The counter-argument is that this PR is further along in one respect: it updates
`scripts/validate-eks-ci-role-policy/main.go` and ships kyverno fixtures, and #2740 currently fails the
`🔐 Validate EKS Authorization` check.

I have not closed anything here — the gap this PR fixes is real until #2740 merges, and picking
between them is a judgement call worth making explicitly.

devantler:

**This PR overlaps #2740, and #2740 is the more correct implementation. Recommending this one be closed as superseded once #2740 merges.**

**Overlap.** #2740 adds a single `restrict-tenant-route-hostnames` ClusterPolicy that covers *every* tenant namespace — `wedding-app`, `ascoachingogvaner` and `doggy-countdown` — plus a fail-closed catch-all rule that denies all hostnames in any ksail tenant namespace with no explicit allow-list. Its `doggy-countdown-approved-hostnames` rule already enforces exactly what this PR enforces, so merging both would leave two `Enforce` ClusterPolicies validating the same namespace with two allow-lists to keep in sync.

**Why #2740's version is the correct one.** This PR requires the hostname to equal `simba.${domain}`. That literal is resolved by Flux post-build substitution when the policy is applied, so on any cluster whose `${domain}` is not `platform.devantler.tech` the rule requires a hostname the tenant does not declare. The tenant's own `deploy/httproute.yaml` in `devantler-tech/doggy-countdown` declares a single **literal** hostname:

```yaml
hostnames:
  - simba.platform.devantler.tech
```

There is no `${domain}` substitution and no `.lan` variant on the tenant side. So on a non-production cluster this policy would deny the tenant's real, unmodified route. #2740 hardcodes `simba.platform.devantler.tech`, which matches the declared route on every cluster.

**Suggested disposition.** Leave this open until #2740 merges so the doggy-countdown protection is never absent, then close as superseded. No finding here is lost: the hostname restriction, the hostname-less-route denial, and the Kyverno tests all exist in #2740's policy.

devantler:

**Closing as superseded — #2740 has merged and its policy is live in production.**

As set out in my earlier comment, #2740 adds `restrict-tenant-route-hostnames`, which covers `doggy-countdown` alongside the other two tenants plus a fail-closed catch-all for any unlisted tenant namespace. It merged as `64767a99` and is now deployed and enforcing:

```
clusterpolicy/restrict-tenant-route-hostnames   Enforce   4 rules   Ready=True
```

Verified against the live cluster: `doggy-countdown/doggy-countdown` serves `simba.platform.devantler.tech`, which the merged policy's `doggy-countdown-approved-hostnames` rule allows, and a hostname-less or foreign-hostname route in that namespace is denied.

**No finding from this PR is lost.** The hostname allow-list, the hostname-less-route denial, and the test fixtures all exist in the merged policy — and, as noted before, this PR's `simba.${domain}` form would have denied the tenant's real route on any cluster whose domain is not `platform.devantler.tech`, because the tenant declares a literal hostname rather than a substitution. The merged version is the correct one.

The separate concern this PR raised about the deprecated `spec.validationFailureAction` field is tracked repo-wide in #3163.

PR #2902, opened 2026-08-02
fix(ci): fingerprint pinned Helm chart versions

devantler:
### Motivation

- The EKS authorization validator normalized exact HelmRelease chart SemVer pins to a sentinel, which allowed pinned chart version changes to bypass the authorization fingerprint despite Helm charts being able to materialize RBAC outside the static Kustomize render.
- Restore fail-closed behavior so any change that can influence rendered RBAC (including exact pinned versions) moves the authorization fingerprint and requires explicit review.

### Description

- Removed the normalization that replaced exact SemVer HelmRelease version fields with a sentinel so HelmRelease documents are serialized verbatim into the authorization surface. (modified `scripts/validate-eks-ci-role-policy/main.go`).
- Simplified `authorizationSurfaceEntry` to marshal the full document text for fingerprinting instead of projecting away the pinned version.
- Updated the expected aggregate authorization fingerprint constant and the surrounding rationale to reflect the restored fail-closed control (modified `scripts/validate-eks-ci-role-policy/main.go`).
- Replaced the previous unit test that preserved version normalization with a regression test that asserts exact, ranged, wildcarded, substituted, and omitted Helm version selectors all move the authorization fingerprint (modified `scripts/validate-eks-ci-role-policy/main_test.go`).

### Testing

- Ran the targeted package tests: `PATH=/tmp:$PATH go test ./scripts/validate-eks-ci-role-policy -count=1` and the modified package tests passed.
- Ran the wider repository test run: `PATH=/tmp:$PATH go test ./... -count=1` and observed unrelated failures in other test packages that depend on external binaries, environment fixtures, or system utilities; these are not caused by the authorization logic change and were due to missing `kubectl` / `yq` or test fixture expectations in the sandbox.
- Ensured `gofmt` and basic static checks (`git diff --check`) reported no formatting or staging errors for the modified files.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a6ddebabc18832bac68df2d2e6accd0)

### DIFF
--- a/scripts/validate-eks-ci-role-policy/main.go
+++ b/scripts/validate-eks-ci-role-policy/main.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


devantler:

## Exact-head static review — superseded/incorrect at `5e3f60e298ca169bd4646d2afd3e58bac1e493c5`

No branch code was checked out or executed.

This draft reverses the reviewed boundary already merged in #2767. That fix normalizes only exact immutable SemVer pins while keeping chart/source identity, values, post-renderers, substitutions, ranges, wildcards, and missing versions byte-exact; the required manifest path separately renders and security-scans the selected exact chart artifact. The current patch removes that boundary and makes every routine pinned Helm update move the manual authorization fingerprint again—the failure that #2767 fixed across the blocked Helm-update queue.

The replacement unit test proves only that a version string changes the hash. It does not demonstrate an RBAC authorization bypass through the exact-chart render/security gate, so it cannot justify reintroducing the known CI failure mode. The branch is also currently DIRTY against `main`. This should not merge as-is; any renewed security claim needs an issue-backed end-to-end RED proof that an exact pinned chart can evade the existing rendered-manifest authorization controls.

devantler:

Closing this stale draft as superseded and incorrect. Merged PR #2767 deliberately normalizes only immutable exact SemVer pins while the required manifest path renders and security-scans the selected chart artifact; current `main` still carries that reviewed boundary. This branch removes the boundary, reintroduces manual fingerprint churn for routine pinned-chart updates, is conflicted with `main`, and supplies no end-to-end bypass proof that would justify reversing #2767. The earlier exact-head static review remains the technical record.

Issue #2674, opened 2026-07-17
refactor(validation): port naming validator to Go


devantler:
> 🤖 Generated by the Daily AI Assistant

Part of #2623

## Evidence

`scripts/validate-naming.py` is the oldest remaining first-party Python validator. CI invokes it directly and `AGENTS.md` documents that command, while Platform’s tooling constitution is Bash-or-Go. The validator has no focused regression suite today, so a mechanical rewrite without parity fixtures could silently weaken manifest naming enforcement.

## Audience and impact

Platform maintainers and contributors need the naming gate to stay deterministic and fail closed without retaining a second scripting runtime. A parity-preserving Go port removes that runtime dependency while keeping every existing diagnostic and exemption stable.

## Expected behavior

The Go command accepts the current zero-argument repository-root flow, also works from another current directory, reports the same grouped/sorted diagnostics and exit status, and preserves all Kubernetes, Flux, patch, CR-folder, vendored-bundle, and Talos naming rules.

## Acceptance criteria

- [ ] Add a focused Go package/command with red-first tests covering every current diagnostic group and exemption.
- [ ] Preserve deterministic ordering, success/failure messages, exit codes, acronym-aware Kind conversion, pluralization, `.enc.yaml`, multi-document, and alternate-CWD behavior.
- [ ] Delete `scripts/validate-naming.py`; add no Python file or Python invocation.
- [ ] Update CI path filters/job commands and `AGENTS.md` to use the Go validator and its tests.
- [ ] Pass race tests, vet/lint, actionlint, the real naming command, and both local/prod static Platform validations without starting a cluster.

## Size

Medium — one isolated validator and its direct CI/documentation consumers.

PR #2914, opened 2026-08-02
chore: stop tracking Python bytecode

devantler:

## Why

A compiled Python file is checked into the repository, and nothing was stopping the next one. It
arrived through a shared CI job that stages the whole tree when it pushes fixes back — the same route
that once committed 182 MegaLinter report files, which is why that tool already has a rule here.
Python did not.

## What

Ignores Python bytecode and drops the tracked file. The pattern is scoped so unrelated stray files
stay visible rather than being silently swallowed.

This does not change the standing position that the Python validators are a migration target — that
work is tracked separately. It stops the mess they make in the meantime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,9 @@ megalinter-reports/
 
 # MegaLinter also drops a generated GitHub config tree here.
 .megalinter_github_conf/
+
+# Python bytecode. The validators and their tests are a migration target, but while they
+# exist every run of them drops a __pycache__ tree next to the source, and the shared CI job
+# stages everything when it pushes auto-fixes back — which is how an empty .pyc reached main.
+__pycache__/
+*.py[cod]


devantler:

## Readiness evidence at `a7537659`

**Condition 3 (tried and evaluated as a user).** This one has a real runtime surface — the user is
whoever next runs the Python validators and then stages files — so I exercised it rather than
reasoning about the pattern.

In a clean clone, before the change:

```
$ python3 -c "import py_compile; py_compile.compile('_probe.py')"
$ git status --porcelain scripts/tests/
?? scripts/tests/__pycache__/_probe.cpython-314.pyc     <-- untracked, so a `git add .` takes it
?? scripts/tests/_probe.py
```

That is the mechanism, reproduced: nothing was ignoring it, which is how the tracked `.pyc` on `main`
arrived in the first place — through #2576, a generated-tests PR, staged wholesale.

After the change the `.pyc` is gone from the listing. **The control is the part that matters**: I
also dropped a `_probe.stray` next to it, and that one still shows. Without it, a pattern that
ignored the entire directory would have produced an identical-looking pass.

**Condition 1 (programmatically tested).** CI is green, 0 pending. Nothing in the repository
references the removed path — checked across YAML, shell and Markdown before deleting it. The file
was 0 bytes, so no behaviour depended on its contents.

**Scope note.** This does not touch the standing decision that the Python validators are a migration
target; platform#2623, #2765, #2674 and monorepo#2140 already cover that and I deliberately did not
open a duplicate. This only stops the artifacts they emit from reaching the repository meanwhile.

devantler:



@coderabbitai review

Requesting at head `a7537659`. CI is green with nothing pending and there are no threads. The change
is a `.gitignore` addition plus dropping one tracked artifact; the reproduction and the
over-breadth control are recorded in the readiness comment.

devantler:




@codex review

Advancing the ladder. CodeRabbit refused at this head with **55 minutes** stated — that window is
flat and long rather than shrinking, and it was consumed by a review this lane just served on
platform#2911, so waiting it out is not the cheap option it was there. Recorded
`cr:no-gate reason=rate-limit@a7537659`.

devantler:

**Correcting my own request markers on this PR.** Both name
`a75376591dcd1e5e6ba1dd0ff85e4ab8bbcbeb2c`, which is not a commit — it does not resolve. I had the
short `a7537659` from `git log --oneline` and filled in the remaining characters instead of reading
the full value.

The real head is `a75376591d9cecd3880e4f1dbc5ea52d54cb4d31`.

Nothing downstream was affected: Codex anchors its own result by the commit it reviewed, and its
clean pass names `a75376591d`, which is this head. But a request marker exists so another instance
can tell a live request from a stale-head one, and one carrying an unresolvable sha cannot do that
job — so it is corrected here rather than left to be discovered.

devantler:

## Readiness at `a75376591d` — promoting

**Review.** `codex@a75376591d`, green — *"Didn't find any major issues"*, reviewed commit matching
this head. Checked both surfaces, not just the clean pass: zero `## Review finding` issue comments,
zero review objects, zero inline comments, zero unresolved threads. Codex publishes findings and its
green on different surfaces and the two can coexist, so the clean pass alone would not have been
enough.

CodeRabbit was tried first per lane priority and refused with a stated **55 minutes** — flat and
long, not the shrinking window that made waiting worthwhile on platform#2911 an hour ago, and it was
long because that review had just consumed the quota. Recorded `cr:no-gate reason=rate-limit` and
advanced one lane. Bugbot untouched.

**Everything else.** CI 13 success / 16 skipped / 0 failures. `CLEAN` against base. Condition 3 is
recorded above, with the over-breadth control that distinguishes this from a pattern that would have
swallowed unrelated files.

PR #2952, opened 2026-08-04
fix(ci): require the Flux verify block to constrain a signer

### LINKED ISSUE #2948: Flux verify gate accepts a block that constrains no signer

## Evidence

`scripts/validate-flux-verify` (added in #2944, closing #2920) asserts that a Flux `verify` block sits at the path KSail actually reads, and that its `provider` is a value KSail treats as "on". Its enabled-predicate borrows KSail's own definition, which is deliberate — it stops the gate drifting into passing configs KSail would ignore.

That definition is `provider` is a non-blank string. It does not require `matchOIDCIdentity`, `secretRef`, or `trustedRootSecretRef`.

Flux treats a missing `secretRef` as **keyless** verification, and `.spec.verify.matchOIDCIdentity` is what constrains the certificate subject and OIDC issuer. So a block reduced to:

```yaml
verify:
  provider: cosign
```

is present, at the right path, "on" by KSail's definition, and accepts **any** keylessly-signed artifact from **any** signer on the platform's root source. The gate passes it.

That is the same failure class the gate exists to catch — configured, in effect, and constraining nothing — one level down from the misplaced-block case that caused the original three-week outage.

**Not a live exposure.** `ksail.prod.yaml` today carries three tight `matchOIDCIdentity` entries pinning issuer to GitHub's OIDC token endpoint and subject to specific workflow refs (`ci.yaml@refs/heads/gh-readonly-queue/main/*`, `cd.yaml@refs/heads/main`, and the DR rebuild path). Verified at `origin/main`.

The impact is **regression cover**. A future edit that removes or empties `matchOIDCIdentity` — a merge resolution, a cleanup, a copy from a laxer example — silently converts the root source from "only our CI may publish this" to "anyone may", and every gate in CI stays green. The gate's own promise ("verification is genuinely in effect") would then be reporting a stronger guarantee than the config delivers, which is worse than no gate on that dimension because it is trusted.

## Expected behaviour

The validator additionally requires the verify block to constrain *who* it trusts: at least one `matchOIDCIdentity` entry with a non-empty `issuer` and `subject`, **or** a `secretRef`, **or** a `trustedRootSecretRef`. A block carrying only `provider` fails with a message naming the missing constraint and the one edit that fixes it.

## Acceptance criteria

- [ ] A verify block with only a non-blank `provider` fails validation, naming the missing identity constraint.
- [ ] The real `ksail.prod.yaml` still passes unchanged (it satisfies the matcher branch today).
- [ ] A block whose `matchOIDCIdentity` is present but empty, or whose entries have a blank `issuer` or `subject`, also fails — presence is not the assertion, non-emptiness is.
- [ ] A `secretRef`-only block passes (keyed verification constrains the signer without OIDC matchers).
- [ ] An ablation flips only the new arms, leaving the existing 25 green.

## Out of scope

Whether the *matchers themselves* are tight enough is a different question, already tracked: #2818 (matchers accept any fixed workflow revision) and #2815 (image-layer subject matchers). This issue is only that the validator accepts a block with **no** matcher at all.

## Size

Small — one predicate plus its ablation arms, in a file that already has the test harness.

## Provenance

Raised by CodeRabbit on #2944 as a `Security & Privacy` observation (CWE-347) inside its `🔇 Additional comments` section, which the review-sweep convention excludes from blocking findings. Assessed on merit rather than on its section, confirmed against the live config, and filed rather than folded into #2944 so that gate could ship without re-staling its review.

devantler:

## Why

The gate that proves our Flux signature verification is switched on could pass a configuration that trusts **anyone**. It checked that the verify block sits where KSail reads it and that verification is on — but not *who* we are willing to trust. A block reduced to `provider: cosign` satisfies both and accepts any signed artifact from any signer on the platform's root source, which is the path every controller, policy and tenant arrives through.

**This is not a live exposure** — production already pins three specific GitHub workflow identities, verified. The value is regression cover: today a future edit that empties those matchers would flip the root source from "only our CI may publish this" to "anyone may", and every check would stay green. A gate that reports a stronger guarantee than the config delivers is worse than no gate on that dimension, because it is trusted.

It is the same failure this gate was written for after a three-week outage — configured, in effect, guaranteeing nothing — one level further in.

## What

The gate now also requires the verify block to name who it trusts, and rejects blocks that only *look* constrained (an empty matcher list, or matchers with a blank issuer or subject). Keyed setups remain valid on their own, so nobody is pushed into switching the gate off to ship one.

Production config passes unchanged.

Fixes #2948
Part of #2627

### DIFF
--- a/scripts/validate-flux-verify/main.go
+++ b/scripts/validate-flux-verify/main.go
@@ -64,6 +64,74 @@ func enabled(provider string) bool {
 	return strings.TrimSpace(provider) != ""
 }
 
+// constrainsSigner reports whether the verify block says WHO it trusts.
+//
+// A block that is present, at the read path, and enabled by KSail's predicate
+// can still verify nothing useful: Flux treats a missing secretRef as KEYLESS
+// verification, and matchOIDCIdentity is what pins the certificate issuer and
+// subject. So `verify: {provider: cosign}` accepts any keylessly-signed
+// artifact from any signer — configured, in effect, and guaranteeing nothing.
+// That is the same class as the misplaced-block outage this validator was
+// written for, one level further in.
+//
+// Three shapes constrain a signer, and any one of them is sufficient:
+// an OIDC matcher naming both an issuer and a subject, a secretRef pinning a
+// public key, or a trustedRootSecretRef pinning a trust root. The keyed
+// branches are accepted deliberately — a future keyed setup is a complete
+// configuration, and failing it would only teach someone to bypass the gate.
+func constrainsSigner(block map[string]any) bool {
+	return namesSecret(block["secretRef"]) ||
+		namesSecret(block["trustedRootSecretRef"]) ||
+		hasUsableOIDCMatcher(block["matchOIDCIdentity"])
+}
+
+// namesSecret reports whether a LocalObjectReference actually names something.
+// A `secretRef:` whose name is absent or blank pins no key, so presence of the
+// key is not the assertion — a non-blank name is.
+func namesSecret(value any) bool {
+	reference, ok := asMapping(value)
+	if !ok {
+		return false
+	}
+
+	name, _ := reference["name"].(string)
+
+	return strings.TrimSpace(name) != ""
+}
+
+// hasUsableOIDCMatcher reports whether at least ONE matcher entry pins both an
+// issuer and a subject.
+//
+// Requiring one usable entry rather than every entry is deliberate: a
+// half-written or commented-out sibling must not veto a block that does
+// constrain a signer, or the gate starts failing correct configs and gets
+// switched off. Requiring BOTH fields within that entry is equally deliberate —
+// a blank issuer lets the subject alone decide trust, and a blank subject
+// trusts every workflow the issuer will sign for, which on a public OIDC
+// issuer is everyone.
+func hasUsableOIDCMatcher(value any) bool {
+	entries, ok := value.([]any)
+	if !ok {
+		return false
+	}
+
+	for _, entry := range entries {
+		matcher, ok := asMapping(entry)
+		if !ok {
+			continue
+		}
+
+		issuer, _ := matcher["issuer"].(string)
+		subject, _ := matcher["subject"].(string)
+
+		if strings.TrimSpace(issuer) != "" && strings.TrimSpace(subject) != "" {
+			return true
+		}
+	}
+
+	return false
+}
+
 // asMapping normalises the two shapes yaml.v3 decodes a mapping into.
 //
 // A mapping whose keys are all strings decodes as map[string]any, but ONE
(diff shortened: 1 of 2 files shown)


devantler:

**Verification record** (detail kept out of the PR body, which stays PM-level).

**RED → GREEN.** The 5 rejection arms were written first and failed against the unmodified validator; the 3 acceptance arms passed before and after, which is what shows the change only tightens. `TestRealConfigValidates` already reads the real `ksail.prod.yaml`, and it stays green — acceptance criterion 2 needed no new test.

**Ablations — 4 arms, each still compiling.** An ablation that stops compiling proves nothing, so each is the weaker thing a careless author would plausibly write:

| # | Ablation | Arms flipped |
|---|---|---|
| 1 | drop the check entirely (`return true`) — the state before this PR | **5** (every rejection arm) |
| A | OIDC matcher checked for **presence only** (`len(entries) >= 0`) | **3** — empty list, blank issuer, blank subject |
| B | `secretRef` checked for **presence only** | **1** — blank-name secretRef |
| C | re-add the `trustedRootSecretRef` branch | **1** — the new KSail-field arm |

**A and B flip disjoint sets** (verified with `comm -12`: empty intersection), so the OIDC and keyed branches are independently load-bearing rather than one masking the other. Their union is 4 of the 5; the 5th (provider-only) carries neither key, so only removing the whole check flips it — which is why ablation 1 is not redundant with A+B.

Restore was verified rather than assumed: the arms after restore are byte-identical to the pre-ablation baseline (`cmp`), re-run with `-count=1` because a cached `ok` would prove nothing.

**🔴 Self-review caught a real fail-open before review, which is why there is a second commit.** The first commit accepted `trustedRootSecretRef` because *Flux* has that field. **KSail does not** — its `FluxVerifySpec` models exactly `provider`, `secretRef`, `matchOIDCIdentity` (checked at `v7.178.14`, the version this repo pins). So that block would have had its key dropped in silence, rendered as a bare `provider`, and **passed the gate while trusting everyone** — this validator's founding bug, reintroduced one level in. Now rejected, with the reason pinned by an arm and by ablation C.

**Tried as a user**, not just tested — ran the built binary against a bare `provider: cosign` config and read what it actually prints:

```
flux verify contract: spec.workload.flux.verify is enabled but constrains no signer, so it
accepts any keylessly-signed artifact from any signer (Flux reads a missing secretRef as
keyless): add a matchOIDCIdentity entry with a non-blank issuer and subject, or a secretRef
naming a key Secret (those are the only two KSail reads — a trustedRootSecretRef is dropped
in silence)
```

That exercise is what surfaced the last defect: the message was still advising `trustedRootSecretRef` as a fix the gate would reject. A guardrail that names an edit which does not work is worse than one that says nothing, so the message now names only edits that resolve it.

**DevEx side of the two-sided test.** Security floor: a verify block that constrains nobody can no longer reach `main`. Everyday cost: **unchanged** — the real config passes untouched, and the failure names the exact edit. The keyed branch passes on its own so a future keyed setup is never pushed into disabling the gate.

`gofmt` clean, `go vet` clean, `golangci-lint` 0 issues.

devantler:

@coderabbitai review

devantler:

**Readiness at `06af5fe555eddb9225f8df2ca625ab506ce7b997`** — promoting.

| Pentad | State |
|---|---|
| Required checks | green (`CI - Required Checks` success; every check-run completed) |
| Unresolved threads | 0 |
| Non-thread review findings | 0 — the newest CodeRabbit review carries only the excluded `🔇 Additional comments (2)` |
| Conflict with base | none (`CLEAN`) |
| Green review | `cr@06af5fe5` — current-head review completion, `📥 Commits` names this head, "No actionable comments", no quota/service marker |

Both finding surfaces (review objects **and** issue comments) were re-read at the moment of judgement rather than relied on from an earlier sweep. User evaluation is recorded above — the built binary was run against a bare `provider` config and its output read as its user, which is what surfaced the misleading remediation text.

PR #2955, opened 2026-08-04
fix(ci): give the merge-queue publication gate its own job

### LINKED ISSUE #2950: Merge-queue validator can be poisoned by an earlier step in the shared changes job

## Evidence

#2949 brings the merge-queue publication-contract gate up to the `cd.yaml` gate's rules with one deliberate exception: `gateJobRunsOnlyItsValidator` is not applied.

On the direct-push route that rule constrains what else may run in the gate job, because the runner's `$GITHUB_ENV` and `$GITHUB_PATH` bridges carry **across steps**. A step earlier in the job can therefore export a variable or prepend a PATH entry that shadows `go` for a later step, while that later step's own run block still contains exactly the two permitted command lines and passes every check aimed at it.

The merge-queue validator lives in `changes`, a shared job that also runs:

```
📑 Checkout · ⚙️ Setup Go · 🩹 Validate merge-group heal contract
🖋️ Validate DR signing contract · 🚦 Validate concurrency queue values · 🔍 Filter paths
```

`gateJobRunsOnlyItsValidator` cannot be applied as-is, because it exists to assert the job runs *nothing but* its validator — true of the dedicated `validate-publication-contract` job, false of `changes` by design.

**Not currently exploited:** no step in `changes` writes `$GITHUB_ENV` or `$GITHUB_PATH` (verified on `origin/main`). `⚙️ Setup Go` does place `GOTOOLCHAIN` into the step environment through the action, which shows the mechanism is live and benign today.

## Why it matters

#2949 closes the workflow-, job- and step-scope `env` and `shell` bypasses, and the run-block bypass. This is the one remaining member of that family on the merge-queue route, so the route is close to parity with `cd.yaml` but not at it — and a reader of #2949 could reasonably assume otherwise. It needs a workflow edit, exactly like every other bypass this contract refuses; it is not reachable by an outside contributor.

## Expected behaviour

The validator step in `changes` is protected against environment poisoning by steps that precede it in the same job, without requiring `changes` to become a dedicated gate job.

## Options (a spike may be the right first move)

1. **Move the validator into its own job.** Cleanest — it inherits `gateJobRunsOnlyItsValidator` unchanged and the `needs` rule from #2949 already requires production to depend on whichever job carries it. Costs one extra runner and a checkout/setup-go per run.
2. **Refuse `$GITHUB_ENV` / `$GITHUB_PATH` writes in the steps preceding the validator** within `changes`. No new job, but it is a denylist over shell syntax — the pattern #2949's own comments record as an unbounded guessing game that an allowlist replaced.
3. **Accept and document**, on the grounds that it needs a workflow edit. Weakest, and inconsistent with refusing the six sibling bypasses.

Option 1 is the recommendation: it reuses an existing, already-tested rule instead of writing a second weaker one, which is the principle the rest of this validator is built on.

## Acceptance criteria

- [ ] A step preceding the validator that writes `$GITHUB_ENV` or `$GITHUB_PATH` fails the contract, or the validator no longer shares a job with steps that could.
- [ ] The shipped `ci.yaml` passes unchanged.
- [ ] An ablation per new rule, each failing a disjoint arm and asserting its own error reason.
- [ ] `cd.yaml`'s existing coverage is unchanged.

## Size

Small–medium, depending on the option chosen.

## Provenance

Raised in my own local review round on #2949 (all three external lanes were quota-limited at that head) and stated in that PR rather than left implicit. Kept out of #2949 to keep it one concern.

devantler:

## Why

The merge-queue route is the normal way changes reach production, and its publication-contract gate was riding along inside a shared job. That job runs several other validators first, and anything running earlier in a job can quietly change the environment the later steps see — so an earlier step could have neutered the gate while the gate's own definition still looked perfectly correct.

The direct-push route already refuses this. The merge-queue route could not, because the rule that refuses it only makes sense for a job dedicated to the gate.

## What

Gives the merge-queue gate its own job, the same shape the direct-push route already uses, so both routes are now held to the identical rule set rather than the weaker one deciding policy. Production jobs and the required-checks aggregate wait on it.

Costs one extra CI job per run.

Fixes #2950

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,13 +39,8 @@ jobs:
           go test ./scripts/validate-merge-group-heal
           go run ./scripts/validate-merge-group-heal .github/workflows/ci.yaml
 
-      - name: 🖋️ Validate DR signing contract
-        run: |
-          go test ./scripts/validate-dr-signing
-          go run ./scripts/validate-dr-signing .github/workflows/dr-rebuild.yaml ksail.prod.yaml
-
-      # The DR contract above checks WHAT the verify block allows. This checks
-      # that the block is in effect at all: it ran at spec.cluster.verify for
+      # The DR contract runs in its own dedicated job (below). This checks that
+      # the verify block is in effect at all: it ran at spec.cluster.verify for
       # ~3 weeks, where KSail discards it, and every text-based check passed
       # throughout (#2920).
       - name: 🔐 Validate Flux verification is in effect
@@ -151,6 +146,37 @@ jobs:
               - 'talos/**'
               - 'talos-local/**'
               - '.github/workflows/ci.yaml'
+
+  validate-publication-contract:
+    name: 🖋️ Validate Publication Contract
+    # A DEDICATED job, mirroring cd.yaml's, and that is the point rather than a
+    # style choice. The runner's $GITHUB_ENV and $GITHUB_PATH bridges carry
+    # across steps, so any step sharing a job with this validator could shadow
+    # `go` for it while the validator step itself still read as correct. Keeping
+    # the job to nothing but its own validator is what the contract's
+    # gateJobRunsOnlyItsValidator rule asserts, and it can only hold here.
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # checkout repository
+    steps:
+      - name: 📑 Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+
+      - name: ⚙️ Setup Go
+        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16 # v6.5.0
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      - name: 🖋️ Validate DR signing contract
+        # Identical invocation to cd.yaml's, deliberately: the two routes to
+        # production must be gated by the same check, or they drift and the
+        # weaker one becomes the real policy.
+        run: |
+          go test ./scripts/validate-dr-signing
+          go run ./scripts/validate-dr-signing .github/workflows/dr-rebuild.yaml ksail.prod.yaml
 
   validate-eks-authorization:
     name: 🔐 Validate EKS Authorization
@@ -769,7 +795,7 @@ jobs:
 
   deploy-prod:
     name: 🚀 Deploy to Prod
-    needs: [changes, validate-eks-authorization]
+    needs: [changes, validate-eks-authorization, validate-publication-contract]
     if: github.event_name == 'merge_group' && needs.changes.outputs.k8s == 'true'
     runs-on: ubuntu-latest
     environment: prod
@@ -808,7 +834,7 @@ jobs:
 
   heal-prod-on-failure:
     name: 🩹 Heal Prod (restore main after unsuccessful merge-group deploy)
-    needs: [changes, deploy-prod]
+    needs: [changes, deploy-prod, validate-publication-contract]
     # The deploy above runs on the SPECULATIVE merge ref, before the PR lands on
     # `main`. When it fails the merge group is ejected — but the OCI `latest` tag
     # (and the prod cluster reconciling it) is left pointing at that ejected
@@ -877,6 +903,7 @@ jobs:
     needs:
       [
         changes,
+        validate-publication-contract,
         validate-eks-authorization,
         validate,
         naming,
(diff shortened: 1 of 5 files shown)


devantler:



@coderabbitai review

devantler:



CodeRabbit declined this review at `b2fea0a1` — account-level review limit, stated window ~29 minutes (reopens ~20:01Z). Recorded so a later run resumes the lane rather than re-spending the same provider. CI is green and the PR is otherwise pentad-clear.

devantler:

**Exercised, not just reasoned about.**

- **RED on the real shipped workflow, not a fixture.** Applying `gateJobRunsOnlyItsValidator` to the shared `changes` job refused the actual `ci.yaml` with exactly the poisoning reason: *"changes runs more than one command step; a step before the validator can export BASH_ENV or a fake go through $GITHUB_ENV/$GITHUB_PATH and leave the validator step itself looking correct."*
- **Two new ablation arms, proven independent.** Ablating the run-step branch flips only the second-command-step arm; ablating the unlisted-action branch flips only the unlisted-action arm; `comm -12` of the two flipped sets is empty. Both ablations compile (an ablation that stops compiling proves nothing).
- **Re-aimed one arm after it silently tested the wrong job.** Anchoring on the `⚙️ Setup Go` step put the poison in `changes` — that step name occurs in 4 jobs. Re-anchored on the validator step name, which is unique.
- **Ran in real CI:** `🖋️ Validate Publication Contract` executed at `b2fea0a1` — `conclusion=success`, 19:16:36Z→19:16:56Z. Confirmed it *ran* rather than being skipped, and that `CI - Required Checks` covers it, which matters because `deploy-prod` is skipped on `pull_request` events.
- Full `./scripts/...` suite green, including `validate-merge-group-heal`, whose pinned `needs:` line this change had to move.

devantler:



@coderabbitai review

devantler (commented):

**Reviewed commit:** `b2fea0a186f347f0a35018289298dc3b3ee90ed0`

**Per-lane evidence, checked directly on this PR at this head — not from a portfolio digest:**
- **CodeRabbit** — rate-limited. Requested twice (19:32Z, 20:16Z); both acks read `Review finished.` with the summary refreshed in place to `Review limit reached`. Latest stated window: **46 minutes** (~21:02Z), which parks this past the end of the run.
- **Codex** — hard usage limit, no retry window: *"You have reached your Codex usage limits for code reviews… add credits."* Only an admin can restore it.
- **Cursor Bugbot** — no check-run at this head; the account-level Cursor spend limit is exhausted across the portfolio (33 PRs carry `neutral` + `output.title=Error`). Admin-only to raise.

### What I checked

- **The rule actually bites on the real workflow.** Applying `gateJobRunsOnlyItsValidator` while the validator still lived in `changes` refused the shipped `ci.yaml` with exactly the poisoning reason. That is the RED, taken against the real file rather than a fixture.
- **The two new ablation arms are independent.** Ablating the run-step branch flips only the second-command-step arm; ablating the unlisted-action branch flips only the unlisted-action arm; `comm -12` empty. Both ablations still compile — an ablation that stops compiling proves nothing.
- **The gate is reachable from every production job.** `deploy-prod` and `heal-prod-on-failure` both `needs` it, each pinned by its own ablation arm.
- **A gate failure cannot pass silently on a `pull_request` event.** `deploy-prod` is skipped there, so the new job is wired into `ci-required-checks` in both `needs` and `job-results`. Without that, a failing gate would have been invisible on PRs — the one thing that would have made this change a regression.
- **Healing behaviour is unchanged.** `heal-prod-on-failure` is governed by `always()`, so the added dependency does not alter eligibility; and if the gate fails, `deploy-prod` is skipped, so `deploy-prod.result == 'skipped'` correctly does *not* trigger a heal.
- **It ran for real.** `🖋️ Validate Publication Contract` executed at this head, `conclusion=success`, 19:16:36Z→19:16:56Z — confirmed run, not skipped.

### Considered and accepted

`cdContractGateJob` and `mergeQueueContractGateJob` now hold the **same string**. I checked this cannot blur a failure: the two call sites emit distinct messages (`direct-push production path…` vs `merge-queue workflow…`), `validateProductionRoutes` only ever parses `ci.yaml`, and the rename ablation still fails for its own reason. Keeping them as separate named constants is deliberate — they describe two different files that merely happen to agree today.

### Residual risk (unchanged by this PR, stated for the record)

`actions/setup-go` necessarily writes `$GITHUB_ENV`/`$GITHUB_PATH`, so the allowlist — not step isolation — remains the control for the two permitted setup actions. That is the same posture `cd.yaml` has always had; this PR brings the merge-queue route up to it rather than beyond it.

Issue #2922, opened 2026-08-03
Cosign verify is configured but not enforced: the deploy path never applies it to the live OCIRepository


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

#2919 merged the cosign policy onto the key KSail actually reads
(`spec.workload.flux.verify` → consumed at `pkg/svc/installer/flux/resources.go:378`). The
merge-queue deploy for it succeeded (`🚀 Deploy to Prod` green) and published a new artifact.

**The live resource is still unverified afterwards.** Read from the prod cluster after that deploy:

- `ocirepository/flux-system` in `flux-system` — `spec.verify` **absent**, `Ready=True`,
  serving a new revision.
- Resource `generation: 3`, created 2026-05-23 — it is long-lived, not recreated by the deploy.

So the config half is now correct and the enforcement half is still off. The gap #2627 tracks is
**not** closed by #2919 alone.

## Why it does not take effect

Two contributing facts, both checked:

1. KSail applies verification by **patching** the OCIRepository, from
   `ensureVerifyIfConfigured` — which sits on the Flux **install / `SetupInstance`** path
   ("creates the FluxInstance CR and configures OCIRepository settings"), not on the path a
   routine `workload push` + `cluster update` deploy exercises against an existing cluster.
2. The `flux-system` OCIRepository is **generated by the flux-operator `FluxInstance`**, whose
   `spec.sync` (`kind: OCIRepository`, `ref: latest`, `url: …/platform/manifests`) carries **no**
   verify field. A resource the operator reconciles from a spec without verify is liable to have a
   direct patch reconciled back out.

This is the create-time-only class already seen in #2524: a field that validates green, is read by
the consumer, and never reaches the running system because the update path does not apply it.

## Expected behaviour

After a normal prod deploy, `ocirepository/flux-system` carries the configured
`spec.verify` and Flux enforces signatures on the infrastructure source.

## Acceptance criteria

- [ ] A routine deploy (not a bootstrap) results in `spec.verify` present on the live
      `flux-system` OCIRepository.
- [ ] The setting survives a flux-operator reconcile — i.e. it is expressed where the operator will
      keep it (most likely `FluxInstance.spec.sync`) rather than as a patch it can revert.
- [ ] Verified **live**, by reading the resource back after a deploy, not by a green validate.
      A green `ksail` validate is explicitly not sufficient evidence here.
- [ ] A negative control: an artifact signed by a non-pinned ref is rejected, so the control is
[...]

Issue #2951, opened 2026-08-04
Cilium rollout gate silently suppresses Talos machine-config sync indefinitely


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

The `🔄 Update cluster (Talos machine config)` step in `.github/actions/deploy-prod/action.yml` — the platform's **only** Talos machine-config sync — has been skipped on every prod deploy for at least four days, while each deploy reported success.

| Deploy | `__self.cluster_update` |
|---|---|
| 2026-08-04T16:09Z ([30927795575](https://github.com/devantler-tech/platform/actions/runs/30927795575)) | `outcome=skipped duration_ms=0` |
| 2026-08-04T15:33Z ([30924710728](https://github.com/devantler-tech/platform/actions/runs/30924710728)) | `outcome=skipped` |
| 2026-08-03T07:58Z ([30795581205](https://github.com/devantler-tech/platform/actions/runs/30795581205)) | `outcome=skipped` |
| 2026-08-01T11:04Z ([30696960482](https://github.com/devantler-tech/platform/actions/runs/30696960482)) | `outcome=skipped` |

The cause is the first arm of its condition:

```yaml
if: >-
  steps.cilium_rollout_gate.outputs.active != 'true' &&
  steps.wait_flux_revision.outcome == 'success'
```

In the 16:09Z run `wait_flux_revision` was `success`, and the gate step printed `Cilium homogeneous-device rollout gate active: Cluster Autoscaler is suspended.` The gate is still active: `kube-system/cluster-autoscaler-hetzner-cluster-autoscaler` reads **`0/0`** on prod.

## The problem

The gate is deliberate and its reasoning is sound — KSail owns Cluster Autoscaler and `cluster update` could reconcile it back to one replica mid-rollout, racing a scale-up before the first Cilium canary is verified. The step's own comment describes it as **temporary**, released by *"a reviewed rollout-completion or rollback artifact"*.

The defect is not the coupling, it is that the coupling is **unbounded and silent**:

- A safety gate for *one* component indefinitely disables an *unrelated* reconcile loop.
- Nothing reports it. The deploy is green, the summary says success, and the skipped step reads as normal conditional behaviour.
- The longer it holds, the larger the un-reconciled machine-config delta becomes — and the first `cluster update` after release applies all of it at once, against nodes that have drifted.

[...]

Issue #1570, opened 2026-05-25
Supply-chain: cosign-verify the infrastructure Flux OCI artifact (extend #1559 beyond apps)


devantler:
> 🤖 Generated by the Daily AI Assistant

## Problem

[#1559](https://github.com/devantler-tech/platform/pull/1559) added keyless cosign signature verification (`spec.verify`) to the **app** `OCIRepository` sources (`wedding-app`, `ascoachingogvaner`), signed by `reusable-workflows/.github/workflows/publish-app.yaml` via GitHub OIDC (Fulcio/Rekor). That closes the supply-chain gap for application artifacts.

The **infrastructure** source is still unverified. The `flux-system` `OCIRepository` — which carries the platform's *own* manifests (`variables`, `infrastructure-controllers`, `infrastructure`, `apps` Kustomizations all `sourceRef` it) — has **no `spec.verify`**. It's the single largest blast radius in the cluster (it defines every controller, policy, and gateway), yet source-controller will reconcile whatever is pushed to `oci://…/manifests` without verifying provenance. Upstream Helm OCI charts (`hcloud-csi`, `longhorn`, `external-dns`, `openbao`, …) are likewise unverified.

## Why this isn't a one-line change

The infra manifests artifact is published by `ksail --config ksail.prod.yaml workload push` in `cd.yaml` — and (unlike the app artifacts) there is **no signing step**, so `spec.verify` can't simply be switched on; the artifact must be **signed first**. This also raises a portfolio question for KSail: should `ksail workload push` grow first-class keyless signing (an `--sign` / OIDC option), so signed publishing is a platform capability rather than a bolted-on `cosign sign` step? See `applications/ksail`.

## Proposed direction (incremental, behaviour-preserving until the final switch)

1. **Sign the infra artifact on publish.** Either (a) add a keyless `cosign sign` step after `ksail … workload push` in `cd.yaml` using the same GitHub-OIDC identity pattern as `publish-app.yaml`, or (b) — preferred long-term — add signing support to `ksail workload push` and call it. Keep the single `release → cd.yaml` deploy path (no parallel workflows).
2. **Verify on the cluster.** Add `spec.verify` (keyless, matching the `cd.yaml`/KSail workflow signing identity) to the `flux-system` `OCIRepository` base. Roll out to **`local` first**, confirm `Ready=True`, then `prod`.
[...]

PR #2961, opened 2026-08-04
docs(ksail): say plainly that the infra cosign block is not enforced live

devantler:

## Why

The cosign block in `ksail.prod.yaml` reads as an active supply-chain control — its comment states
that it "makes Flux REJECT any infra artifact whose signature is missing". On the running cluster it
does not. #2922 establishes why, from live Server-Side Apply ownership: the flux-operator generates
`ocirepository/flux-system` and owns its entire spec, so a routine deploy never lands `spec.verify`
on the live resource.

A control that is documented as enforcing but is inert is worse than one that is plainly absent —
it is what lets the gap survive a reader who checks the config and moves on. CI does not catch it
either: the `validate-flux-verify` guard checks the block is in the right *place*, which is a
genuinely useful check, but no text check can see that the setting never reaches the cluster.

## What

Comment-only. States the as-is behaviour, points at #2922, gives the one command that checks the
live resource, and names the signature enforcement that *is* active today so the entry does not
read as "no verification anywhere". The configuration itself is unchanged — it is correct, it is
what the install path consumes, and it is what a DR rebuild applies.

Part of #2922


devantler:


@coderabbitai review

devantler:


@codex review

devantler:


CodeRabbit is rate-limited (55 min) and Codex is at its usage limit, so requesting Bugbot next. The bare trigger follows in the next comment — Bugbot exact-matches the whole comment body, so it cannot carry this line.

devantler:
@cursor review

devantler:


@coderabbitai review

devantler:

## State: ready except for review — parked deliberately, not stalled

**Head `eba3b7fb7bdc2e890d627f74f5f051555ef4b189`.** CI **19 pass / 12 skipped / 0 pending**.
Verified locally at this head:

- **Comment-only**, proven rather than asserted: `yq -o=json` of this file and of `origin/main`'s
  copy are byte-identical, so nothing structural changed.
- `scripts/validate-flux-verify` passes — and it is **not** passing vacuously: deleting the verify
  block from a copy makes it exit 1 with `no verify block at spec.workload.flux.verify`.

## Why it is not promoted: all three review lanes are down, with per-PR evidence at head

| Lane | Evidence on this PR | Recovers |
|---|---|---|
| CodeRabbit | summary naming `9eaad2b599`: `Review limit reached`, **45 minutes** from 22:34Z | ~23:19Z |
| Codex | `You have reached your Codex usage limits for code reviews` — no window given | needs credits (maintainer) |
| Cursor Bugbot | check-run `conclusion=neutral` + `output.title=Error` (run never started); `cursor[bot]`: `Bugbot couldn't run - usage limit reached` | needs an admin to raise the limit |

A local review round is permitted in exactly this situation, and I am **choosing not to use it here**.
This diff is prose, and prose is its entire payload — my own review passes have already had to correct
it **three times**:

1. "the operator owns its whole spec" overstated the measurement and contradicted the finding the
   change rests on (flux-operator owns five named fields, **not** `verify`);
2. the rewrite had silently deleted the standing "don't hand-write an in-repo override" warning, while
   the new text actively invited that exact mistake;
3. the stated *reason* was wrong outright — see below.

Given that record, a real external reviewer is worth one tick. CodeRabbit is free on this repo and
recovers at ~23:19Z, before the next scheduled run.

## Substantive correction folded in

The comment originally blamed an install-only KSail code path. That was inherited from #2922's body
and was wrong: the deploy's `ksail cluster update` is **skipped entirely** by the Cilium rollout gate
(#2951) — `outcome=skipped duration_ms=0` on both prod deploys tonight — so the command that would
apply this has not run at all. The text now says that, and #2951 is linked as a blocker on #2922.

## To finish

Request CodeRabbit at this head after ~23:19Z, resolve anything it raises, then promote and merge.
No code work is outstanding.

devantler:


@coderabbitai review

devantler:


@codex review

devantler (commented):

**Reviewed commit:** `eba3b7fb7bdc2e890d627f74f5f051555ef4b189`

### Per-lane evidence (direct, per-PR, at this head)

| Lane | Outcome | When |
|---|---|---|
| CodeRabbit | `⚠️ Action not completed — Review rate limited.` No retry window stated. The ack's first line read "I will review the changes" — the body is authoritative. | 2026-08-04T23:25:06Z |
| Codex | `You have reached your Codex usage limits for code reviews` — needs account credits, no window; maintainer-only to restore. | 2026-08-04T23:26:51Z |
| Cursor Bugbot | `Bugbot couldn't run - usage limit reached` on this PR — account/team-level, no window, not head-dependent; maintainer-only to restore. | 2026-08-04T22:26:41Z |

### What I checked, and how

This is a prose-accuracy change whose entire purpose is to stop a comment overclaiming, so every factual claim was verified against live state rather than read for plausibility.

1. **"the live ocirepository/flux-system carries no spec.verify"** — **TRUE.** Verified with a positive control, because an empty grep is indistinguishable from a failed command: the object reads back 55 lines of YAML with `spec` keys `interval, provider, ref, secretRef, timeout, url`, and an explicit probe returns `verify ABSENT`.
2. **"`ksail cluster update` is skipped by the Cilium rollout gate — `outcome=skipped duration_ms=0` on every prod deploy for days"** — **TRUE on both deploy paths.** The job-level `🚀 Deploy to Production` step reports `success`, so the composite's internal outcome is the only thing that answers this:
   - real path (`ci.yaml deploy-prod` on `merge_group`), run `30956176478`, 2026-08-04T22:22:58Z → `end-action id=__self.cluster_update;outcome=skipped;conclusion=skipped;duration_ms=0`
   - manual path (`cd.yaml`), 2026-08-03 and 2026-08-01 → `outcome=skipped` on both
   The quoted `duration_ms=0` is exact, not approximate.
3. **"flux-operator owns url/ref/interval/provider/secretRef … but NOT verify"** — **TRUE**, and it is what makes the SSA-survival claim sound rather than hopeful.
4. **Comment-only** — **PROVEN.** Parsed structure vs `main` (`6e3bf87d3a`) is identical at 3141 bytes while the text differs (14777 → 15764 bytes). The size figures are quoted because an empty-file comparison also reports "identical"; an earlier attempt of mine did exactly that and was caught by this control.
5. **The documented recovery command runs** and returns what the comment says it returns.
6. **No load-bearing warning was dropped in the rewrite.** The prior comment's operative caution — do not hand-write an in-repo OCIRepository override, because it loses the reconcile fight — survives and is strengthened with the blast radius. Only the origin-story framing (`#1559` history) was removed; the live gap reference `#1570` is retained. Rewriting a comment block is the standard way to silently delete a live "don't do this", so this was checked line by line rather than assumed.

### One thing I checked and withdrew

I initially believed the forward-looking sentence ("once #2951 releases that gate this may simply start working") was incomplete, on the grounds that the `CD` workflow has run only via `workflow_dispatch` since 2026-06-19. That was **my** error, not the PR's: `cd.yaml`'s own header states prod deploys happen automatically in the CI merge queue and `CD` is only the direct-push fallback. I was reading the wrong workflow. The sentence is accurate — merge-queue deploys are automatic, so releasing the gate is genuinely sufficient. Recording it because the inference was wrong in the direction that would have produced a bogus finding.

### Verdict

Comment-only, every factual claim verified against live cluster and workflow state, no operative warning lost. This replaces a comment that asserted the block "makes Flux REJECT any unsigned infra artifact" — which is false today — with an accurate statement plus the check a reader can run themselves.

devantler:

## Promoting — all three readiness conditions hold at `eba3b7fb7b`

1. **Programmatically tested** — 31 checks, 0 failing. The `validate-flux-verify` guard passes and was previously ablated (deleting the block → exit 1), so it is known non-vacuous. Comment-only proven structurally: parsed YAML identical to `main` at 3141 bytes while the text differs (14777 → 15764).
2. **Reviewed** — local review round at this exact head, verdict **no P0/P1**, posted after direct per-PR evidence that all three lanes are limited (CodeRabbit rate-limited 23:25:06Z with no window, Codex out of credits 23:26:51Z, Bugbot usage-limited). Pentad re-checked at merge time: 0 unresolved threads queried directly, 0 findings, CLEAN.
3. **Tried and evaluated as a user** — I ran the command this comment tells a reader to run, against the live cluster, and confirmed it returns what the comment claims (`verify ABSENT`, with a positive control proving the object was actually read). I also verified the causal claim on the real deploy path — `end-action id=__self.cluster_update;outcome=skipped;duration_ms=0` on merge_group run `30956176478` — rather than inheriting it.

The change makes a false statement true: the block previously claimed it "makes Flux REJECT any unsigned infra artifact", which it does not do today. #2922 tracks closing the gap for real; this stops the comment from asserting a control that is not in effect.

PR #2968, opened 2026-08-05
fix(ci): prove every publication step's failure stops the run

### LINKED ISSUE #2967: security(ci): the publication contract cannot tell a step that ran from one whose failure was discarded

## Problem

The publication contract proves each step of the production publication chain **exists and sits in the right place**. That is a positional property. Nothing proved that a step's **failure actually stops the run**.

Measured on `main` (`2c2d4c09`), adding one line to the composite action:

```yaml
      id: verify_evidence
      continue-on-error: true
```

leaves every guard green:

| Guard | Result with the gate suppressed |
|---|---|
| `go test ./scripts/validate-dr-signing` | ✅ ok |
| `go run ./scripts/validate-dr-signing …` | ✅ "DR publication contract passed." |
| `bash scripts/tests/test-verify-published-evidence.sh` | ✅ passed=30 failed=0 |

The gate stays present, correctly ordered between the provenance attestation and the promotion, invoked with the resolved staging digest, and reading `ENFORCE=true` — so the ordering chain, the wiring assertions and the enforcement ratchet are all still satisfied. Its verdict is simply discarded, and `promote_latest` then moves the mutable `latest` tag onto bytes carrying no usable evidence.

## Impact

`verify_evidence` is the last check between "attestations were requested" and "production starts following these bytes within the minute". Attestations here are published with `create-storage-record: false`, so *created* and *retrievable from the registry* are different facts, and that gate is what distinguishes them. A one-line change neutralises it while CI reports the publication contract as enforced — the control reads as present and is not.

This is the same class the validator already documents at four sites (`requireEnforcedStep`: "a step was matched and its metadata discarded, so `if: false` or `continue-on-error: true` left it named-but-not-run"), one level further down — inside the composite action's own steps, where none of the existing checks look.

## Expected behaviour

No step of the publication action may suppress its own failure or carry a condition that could skip it. A step whose failure does not stop the run has no place in a chain whose whole purpose is producing bytes production will trust.

## Acceptance criteria

- [ ] `validate-dr-signing` rejects `continue-on-error` on any publication-action step, naming the offending step.
- [ ] It rejects an `if:` condition on any publication-action step, on the same refuse-rather-than-interpret basis the job-level check already uses.
- [ ] It fails closed when the action cannot be parsed or declares no steps.
- [ ] The shipped action still passes unchanged (no false positive).
- [ ] Coverage is by construction rather than an allowlist of step ids, so a newly added step is covered without editing the check.

Rough size: small.

Part of #2627 (its criteria 2 and 3 depend on this property holding).

devantler:

## Why

Production trusts whatever arrives through the root OCI artifact, and one step — `verify_evidence` — is the last check between "attestations were requested" and "production starts following these bytes". That gate could be switched off by a single line while CI still reported the publication contract as fully enforced, because every existing check proved only that steps exist in the right *order*, never that a step's failure actually stops the run.

## What

Every step in the publication action must now prove its failure is fatal, reusing the check this repository already applies to workflow-level steps. The shipped action passes unchanged; the suppressed one is refused by name, so the failure points at the line to change.

Fixes #2967
Part of #2627

### DIFF
--- a/scripts/validate-dr-signing/main.go
+++ b/scripts/validate-dr-signing/main.go
@@ -180,7 +180,62 @@ func validatePublicationAction(action string) error {
 		return errors.New("publication action must verify latest resolves to the staged digest")
 	}
 
+	return publicationStepsAreEnforced(action)
+}
+
+// publicationStepsAreEnforced proves every publication step's failure actually
+// stops the run.
+//
+// 🔴 THE FIFTH SITE OF THE requireEnforcedStep MISTAKE, and the one the checks
+// above cannot reach. Everything before this point is positional: it proves a
+// step EXISTS and sits in the right place in the chain. Position says nothing
+// about whether the step's verdict is honoured. `continue-on-error: true` on
+// `verify_evidence` keeps the gate present, correctly ordered, invoked with the
+// resolved digest and reading ENFORCE=true — so the ordering chain here, the
+// wiring assertions in test-verify-published-evidence.sh, and the enforcement
+// ratchet all still pass — while the gate's failure is discarded and
+// `promote_latest` moves the mutable tag onto bytes with no usable evidence.
+// That is precisely the state root verification exists to make impossible.
+//
+// Every step is covered rather than a named subset: this action's whole purpose
+// is producing bytes production will trust, so a step whose failure does not
+// stop it has no place in the chain, and an allowlist would need extending each
+// time a step is added — the omission this file has already made four times.
+func publicationStepsAreEnforced(action string) error {
+	document, err := decodeWorkflow(action)
+	if err != nil {
+		return fmt.Errorf("publication action is unreadable, so its steps cannot be proven enforced: %w", err)
+	}
+	runs, ok := document["runs"].(map[string]any)
+	if !ok {
+		return errors.New("publication action has no runs mapping, so its steps cannot be proven enforced")
+	}
+	steps, ok := runs["steps"].([]any)
+	if !ok || len(steps) == 0 {
+		return errors.New("publication action declares no steps, so nothing can be proven enforced")
+	}
+	for index, raw := range steps {
+		step, ok := raw.(map[string]any)
+		if !ok {
+			return fmt.Errorf("publication action step %d is not a mapping, so it cannot be proven enforced", index)
+		}
+		if err := enforcesFailure(step, describePublicationStep(step, index)); err != nil {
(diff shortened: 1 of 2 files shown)


devantler:

## Verification record

**RED, with a control on the ablation.** The five new cases fail on the unmodified validator, every one with `got <nil>` — the check simply accepted each suppression. `go vet` exits 0 on that same tree, so the failure is the assertion rather than a build break, and each ablation is guarded by `if ablated == publisher { t.Fatalf(...) }`, so a case that stopped matching a real line would fail loudly instead of passing vacuously.

**GREEN.** All five pass, and the full package is unchanged otherwise (`ok github.com/devantler-tech/platform/scripts/validate-dr-signing`).

**Exercised as a user — the real CLI, both directions, on the real files.** This is what CI runs, so it is the path that matters:

| Tree | Command | Result |
|---|---|---|
| shipped action, untouched | `go run ./scripts/validate-dr-signing …` | `DR publication contract passed.` (rc=0) — no false positive |
| one line added: `continue-on-error: true` under `id: verify_evidence` | same command | rc=1 — `publication action step "verify_evidence" sets continue-on-error: true, so it can fail the publication contract and still report success` |

The second row is the decisive one: **that exact ablation passed every guard before this change** — `go test` ok, `DR publication contract passed.`, and `test-verify-published-evidence.sh` at `passed=30 failed=0`. The action was restored from a byte-for-byte backup after each ablation and verified identical with `diff -q`.

**The mechanism was verified, not assumed.** I did not take it on trust that `continue-on-error` even applies inside a composite action — if GitHub ignored the key there, the whole finding would be wrong. GitHub's metadata-syntax reference documents it on `runs.steps[*]` as: "Prevents the action from failing when a step fails. Set to `true` to allow the action to pass when this step fails." A suppressed `verify_evidence` therefore does not stop the action, and the later unconditional `promote_latest` runs.

**Fail-closed paths checked:** unparseable action, missing `runs` mapping, absent or non-list `steps`, and a step that is not a mapping each return an error rather than passing. `continue-on-error: false` is correctly accepted as the no-op it is; any non-boolean value (including an expression) is refused rather than interpreted, matching the existing job-level precedent in this file.

**Scope:** additive only — 55 lines in `main.go`, 74 in `main_test.go`, nothing removed, no production workflow or action file touched.

devantler:



@coderabbitai review

Issue #2823, opened 2026-07-27
Six posture exceptions in controller-rbac.yaml are structurally inert, and the migrated four have not gone to zero


devantler:
> 🤖 Generated by the Agentic Engineer

> ## ⚠️ STATUS 2026-08-15 — read this before doing anything below
>
> **Step 1 is ANSWERED. Do not re-run it.** The measurement is
> [in this comment](https://github.com/devantler-tech/platform/issues/2823#issuecomment-5299398585):
> on the CI scan — the only surface that applies these exceptions — the `controller-rbac` CR
> **suppresses 17 findings** (C-0053 ×13, C-0037 ×3, C-0031 ×1), and C-0053's residual 10 are
> **100% out-of-scope** (tenant namespaces absent from the selector). By this issue's own decision
> rule, **the mechanism works**.
>
> **The sections below are preserved as dated evidence, and two of their conclusions are refuted.**
> The title's "structurally inert" is wrong for three of the six controls, and every count in
> *Problem* and *What is NOT yet established* was taken from the kubescape-operator's stored scan
> resources — which apply **no exceptions at all** (`appliedIgnoreRules` is empty, recorded in
> `ci.yaml` since #2969). A non-zero count there is not evidence about an exception, so those tables
> cannot answer the question this issue asks.
>
> This banner exists because the original *Direction* sent **15 separate runs over 19 days** to
> re-measure the same refuted premise, twice ending in public self-correction. Measure exception
> effect **only** by diffing two `ksail workload scan --exceptions` runs.

## Problem

`k8s/bases/infrastructure/cluster-security-exceptions/controller-rbac.yaml` suppresses six controls through a `match.namespaceSelector`. That file already documents why such a selector cannot suppress an RBAC-anchored finding, and says to migrate the six "if they surface failing objects".

**They do — all six.** Measured 2026-07-27 against `admin@prod`:

| Control | Name | Failing objects | Anchored on an RBAC object | Cluster-scoped (empty namespace) |
|---|---|---|---|---|
| C-0053 | Access container service account | 155 | 100% | 103 |
| C-0037 | CoreDNS poisoning | 41 | 100% | 30 |
| C-0188 | CIS-5.1.4 minimize access to create pods | 20 | 100% | 13 |
| C-0031 | Delete Kubernetes events | 18 | 100% | 14 |
| C-0063 | Portforwarding privileges | 9 | 100% | 5 |
[...]

Issue #2933, opened 2026-08-03
fix(security): a Kubescape framework-load failure still reports zero findings instead of failing


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

The in-cluster Kubescape configuration scan is **aborting on every run**. In the current pod's ~45h
log window there are **4 scan attempts and 0 successes**, including both scheduled runs:

```
2026-08-02T09:12:30Z  scanning failed: open /home/nonroot/.kubescape/security.json: no such file or directory
2026-08-03T09:12:19Z  scanning failed: open /home/nonroot/.kubescape/security.json: no such file or directory
```

each preceded by a `failed to list frameworks` error. The newest stored scan object is from
**2026-08-01**, so posture data has been frozen for ~2 days. The CI posture gate is unaffected and
stays green, which is exactly the divergence that lets this go unnoticed.

## Root cause

Read-only diagnosis, from the deployed manifest:

- the policy-artifact cache (`/home/nonroot/.kubescape`) is backed by an **`emptyDir`**, so it is
  wiped whenever the pod restarts
- **`KS_OFFLINE=true`** then prevents re-downloading the framework artifacts
- **`KS_DOWNLOAD_ARTIFACTS=true`** is set at the same time, which contradicts the offline flag
- nothing seeds the cache — there is no init container, ConfigMap, or persistent volume behind it

The pod restarted ~45h ago. From that moment the cache was empty, no framework could be loaded, and
every scheduled scan has evaluated **zero controls**.

## Why this is Security rather than a Chore

**A broken scanner and a compliant cluster read identically.** With no framework loaded, every
control reports zero findings — so the posture surface presents as clean while nothing is actually
being evaluated. Any consumer that trusts those numbers inherits a false all-clear, and a real
regression introduced in the last two days would be invisible.

This already has a concrete downstream victim: #2855 asks for "zeroed" controls to be graduated to
Kyverno `Enforce`. Under the current state **every** control looks zeroed, so a candidate list built
now would graduate controls that were never evaluated. That issue cannot safely proceed until this
is fixed.

## Expected behaviour

The scheduled scan loads its framework and evaluates controls on every run, and a run that cannot
load its framework **fails loudly** rather than reporting zero findings.

## Acceptance criteria

- [ ] The framework artifacts survive a pod restart — seeded deliberately rather than left to an
      `emptyDir` that offline mode cannot refill
[...]

PR #2969, opened 2026-08-05
docs(ci): correct where Kubescape exceptions are actually applied

devantler:

## Why

A comment in the CI workflow told readers that the security exceptions are applied by the
in-cluster scanner as well as by CI. They are not. Anyone checking whether an exception is
working — which is exactly what two open security issues are about — would look at the
in-cluster numbers, see the finding still listed, and conclude the exception is broken when
it is fine.

That misread already cost a run: the in-cluster numbers quoted on #2823 are pre-exception
figures, and a fix verified against them could never show an effect either way.

## What

Corrects the comment to say where exceptions actually take effect (the CI scan and the
Headlamp view), and warns off using the in-cluster scanner's stored results to judge an
exception. Comment-only — no behaviour change.

Part of #2823

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -500,12 +500,18 @@ jobs:
         #
         # The scan consumes the platform's justified exceptions: the
         # ClusterSecurityException CRs (k8s/bases/infrastructure/
-        # cluster-security-exceptions/ — the single source of truth, also
-        # consumed in-cluster by the kubescape-operator) are converted at scan
-        # time into Kubescape's native exceptions format by
+        # cluster-security-exceptions/ — the single source of truth) are
+        # converted at scan time into Kubescape's native exceptions format by
         # scripts/generate-kubescape-exceptions (#2264). With runtime-
         # enforced and except-only findings suppressed, the threshold gates the
         # residual REAL posture instead of sitting ~15 points under it.
+        #
+        # Two surfaces apply these exceptions: this scan, and the Headlamp view
+        # via the generated headlamp-exceptions ConfigMap. The kubescape-operator's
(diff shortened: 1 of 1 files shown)


devantler:

## Readiness evidence at `659f8104`

**1. Programmatically tested.** Comment-only change to `.github/workflows/ci.yaml`; MegaLinter passed
on this head and the workflow parses. No behaviour to regress.

**2. Reviewed.** Pending — see the lane note below.

**3. Tried and evaluated as a user.** This change has **no exercisable runtime surface** (a YAML
comment), so per the contract I verified the *claim it makes* instead of the code path, live against
prod rather than inheriting it from the issue that prompted it:

- sampled the operator scan CRDs per-object: **21 objects / 243 control entries, `appliedIgnoreRules`
  empty throughout** — confirming the comment is right that the operator's stored scans do not apply
  the exceptions
- ⚠️ and a correction worth having in the record: the aggregated storage API **strips `spec.controls`
  on LIST calls** and only populates it on a single-object GET. A list-based read reports
  `spec.controls: null` for all 1542 objects, which looks exactly like "the scanner recorded nothing".
  Anyone re-checking this comment with `get workloadconfigurationscans -A` will get a reading that
  cannot distinguish an empty result from a stripped field. I have added the same caveat to #2933.

**Lane state.** Codex and Cursor Bugbot are both hard usage/billing limited today (no retry window,
admin action required). The only CodeRabbit status on this head reads
`success — Review skipped: automatic reviews are disabled`, which is **not** a review — note that
CodeRabbit publishes `state=success` for skipped and rate-limited runs alike, so the status
`description` is the discriminator, not `state`. A real review is still owed here.

devantler:


@coderabbitai review

PR #2977, opened 2026-08-05
fix(ci): retry the cordon claim when an unrelated Node write breaks its CAS test

### LINKED ISSUE #2976: Prod deploy evicted from the merge queue when an unrelated Node write breaks the cordon-claim CAS test

## Evidence

The prod deploy fails intermittently in `🔑 Stage Flux and consumer GHCR pull credential`, evicting the PR from the merge queue:

```
##[error]Could not atomically claim and cordon Talos node autoscale-cx33-12ac9ae5b0b9effe; refusing to drain it.
cordon-claim: The request is invalid: the server rejected our request due to an error in our request
```

Measured on 2026-08-05 across `ci.yaml` runs: **4 `merge_group` failures interleaved with 8 successes** on unchanged code —

| run | PR | time |
|---|---|---|
| 30997196586 | pr-2970 | 10:23Z |
| 31003661088 | pr-2971 | 11:58Z |
| 31006527422 | pr-2970 | 12:38Z |
| 31011548385 | pr-2971 | 13:43Z |

Intermittency on unchanged code rules out a malformed patch, which would fail every time.

## Diagnosis

`claim_node_cordon_ownership` (`scripts/refresh-flux-ghcr-auth.sh`) builds an RFC 6902 patch whose first op is

```json
{"op": "test", "path": "/metadata/resourceVersion", "value": "<version read earlier>"}
```

A failed JSON Patch `test` op is exactly what the API server rejects as 422, which `kubectl` renders as *"The request is invalid: the server rejected our request due to an error in our request"*.

That test asserts **nothing whatsoever changed on the Node object**, but a Node has several independent writers. Read live from the failing node's `managedFields`:

| manager | operation | last write |
|---|---|---|
| `kubelet` | Update `status` | 15:21:32Z (≈3 min before sampling) |
| `kube-controller-manager` | Update | 14:01:21Z |
| `Go-http-client` | Update | 14:01:40Z |
| `hcloud-cloud-controller-manager` | Update `status` | 03:17:32Z |
| `cilium-operator-generic` | Update `status` | 03:17:32Z |

So a routine kubelet status heartbeat — which has no bearing on whether the node is safe to cordon and drain — invalidates the claim and fails the whole deploy. There is **no retry**: `claim_node_cordon_ownership` is called once per node and returns non-zero on the first conflict.

The guard is simultaneously **too strict** (any unrelated status write aborts it) and only *indirectly* protective: `resourceVersion` is a proxy for "the invariants I validated still hold", not the invariants themselves.

## Why it matters

This is the delivery path. Every eviction costs a full merge-queue cycle, and because the trigger is an unrelated background write, re-queuing without a fix is a coin flip — the exact "re-queue without root-causing" trap the contract warns about. It is not specific to the two dependency PRs that surfaced it; any PR can hit it.

## Expected behaviour

The claim survives writes that do not affect drain safety, and still refuses when a genuinely conflicting change occurred.

## Acceptance criteria

- [ ] A CAS conflict on the claim patch is **retried** (bounded, with the node re-read and the caller's captured invariants re-validated on each attempt) rather than failing the deploy on first conflict.
- [ ] The claim no longer aborts on writes irrelevant to drain safety. Either narrow the `test` ops to the actual invariants — `metadata.uid`, `spec.unschedulable` == captured `was_cordoned`, `spec.taints` == captured `initial_taints`, and the owner annotation being unclaimed/ours — or keep `resourceVersion` but re-read and re-validate on retry.
- [ ] Exhausting the retry budget still **fails closed** with the existing refusal message; the safety property is unchanged.
- [ ] RED/GREEN proof in `scripts/tests/refresh-flux-ghcr-auth/` using the existing fake `kubectl`: a first-attempt 422 followed by success must fail before the fix and pass after, and a persistent conflict must still refuse to drain.

## Size

Small–medium. Contained to `claim_node_cordon_ownership` and its two call sites, plus tests.

⚠️ This touches the path that drains prod nodes. The retry must re-validate rather than blindly re-patch — a retry that skips revalidation would convert a safe refusal into an unsafe drain, which is strictly worse than the current flake.

devantler:

## Why

The production deploy has been failing intermittently and evicting PRs from the merge queue — 4 evictions today alone, interleaved with 8 clean runs on unchanged code. The cause is a safety check that is stricter than it needs to be: before draining a node, the deploy claims it with a patch that asserts *nothing at all* has changed on that node since it was read. But a Kubernetes node is written by several background controllers, so a routine kubelet heartbeat — nothing to do with whether the node is safe to drain — was enough to abort the whole deploy.

Because the trigger is unrelated background activity, re-queuing an evicted PR is a coin flip rather than a fix.

## What

The claim now retries against a fresh read instead of giving up on the first conflict — but only while every fact it originally relied on still holds. A genuine conflict (someone else cordoning the node, taking ownership, changing its taints, or the node being replaced) still refuses to drain, exactly as before, and running out of retries still fails closed with the same refusal.

Fixes #2976

⚠️ This touches the path that drains production nodes, so the safety behaviour is pinned by tests in both directions: a benign conflict must now succeed, and a persistent or genuine conflict must still refuse.

### DIFF
--- a/scripts/refresh-flux-ghcr-auth-safety.sh
+++ b/scripts/refresh-flux-ghcr-auth-safety.sh
@@ -63,6 +63,46 @@ select_talos_node_targets() {
   rm -f "${unsorted_targets}"
 }
 
+# Validate that an unclaimed node is still exactly what the capturing read saw,
+# so a rejected atomic claim can be retried against a fresh resourceVersion.
+# A Node object has several independent writers (kubelet status heartbeats, the
+# cloud controller manager, the CNI operator), and any of their writes rejects
+# the claim's resourceVersion test without changing anything relevant to drain
+# safety. Retrying is only sound while every fact the caller captured still
+# holds: same node identity, not being deleted, still unowned by this bridge,
+# and unchanged scheduling intent. A concurrent cordon, ownership grab, taint
+# edit, or node replacement therefore still refuses instead of retrying.
+node_claim_preconditions_still_hold() {
+  local state_file="$1"
+  local initial_node_uid="$2"
+  local was_cordoned="$3"
+  local initial_node_taints="$4"
+
+  jq -e \
+    --arg owner_annotation \
+    "platform.devantler.tech/ghcr-auth-drain-owner" \
+    --arg recovery_annotation \
+    "platform.devantler.tech/ghcr-auth-drain-recovery" \
+    --arg uid "${initial_node_uid}" \
+    --argjson was_cordoned "${was_cordoned}" \
+    --argjson initial_taints "${initial_node_taints}" '
+    .metadata.uid == $uid
+    and .metadata.deletionTimestamp == null
+    and ((.metadata.annotations[$owner_annotation] // "") == "")
+    and ((.metadata.annotations[$recovery_annotation] // "") == "")
+    and ((if (.spec.unschedulable // false) then 1 else 0 end)
+      == $was_cordoned)
+    and (((.spec.taints // [])
+      | map(select((
+          .key == "node.kubernetes.io/unschedulable"
+          and .effect == "NoSchedule"
+          and (.value // "") == ""
+        ) | not))
+      | sort_by([.key, .effect, (.value // ""), (.timeAdded // "")]))
+      == $initial_taints)
+  ' "${state_file}" >/dev/null
+}
+
 # Validate the scheduling state captured immediately before a Talos reboot.
 # The bridge must still own its serialization annotation while the node stays
 # cordoned; neither a bridge-created nor a pre-existing cordon may proceed after
--- a/scripts/refresh-flux-ghcr-auth.sh
+++ b/scripts/refresh-flux-ghcr-auth.sh
@@ -921,6 +921,53 @@ verify_bootstrap_quarantine_covers_unproved_destinations() {
 claim_node_cordon_ownership() {
   local node_name="$1" owner_token="$2" state_file="$3" result_file="$4"
   local recovery_record="${5:-}"
+  local was_cordoned="${6:-}" initial_taints="${7:-}"
+  local resource_version node_uid initial_node_uid
+  local attempt=1
+  local max_attempts="${CORDON_CLAIM_MAX_ATTEMPTS:-5}"
+
+  initial_node_uid="$(jq -er '.metadata.uid' "${state_file}")"
+
+  while :; do
+    build_and_apply_cordon_claim \
+      "${node_name}" "${owner_token}" "${state_file}" "${result_file}" \
+      "${recovery_record}" && return 0
+
+    # Only a conflict against a still-conforming node is retryable, and only
+    # when the caller supplied the scheduling facts needed to re-verify that.
+    if [[ -z "${was_cordoned}" || -z "${initial_taints}" ]] ||
+      ((attempt >= max_attempts)); then
+      break
+    fi
+
+    if ! kubectl \
+      --context "${KUBE_CONTEXT}" \
+      get node "${node_name}" \
+      --output json \
+      >"${state_file}" 2>"${result_file}"; then
+      break
+    fi
+
+    if ! node_claim_preconditions_still_hold \
+      "${state_file}" "${initial_node_uid}" \
+      "${was_cordoned}" "${initial_taints}"; then
+      break
+    fi
+
+    attempt=$((attempt + 1))
+  done
+
+  echo "::error::Could not atomically claim and cordon Talos node ${node_name}; refusing to drain it."
+  emit_safe_operation_output "cordon-claim" "${result_file}"
+  return 1
+}
+
+# Render the claim patch from the current captured state and apply it. Split out
+# so a rejected claim can be rebuilt against a re-read resourceVersion without
+# duplicating the patch shape.
+build_and_apply_cordon_claim() {
+  local node_name="$1" owner_token="$2" state_file="$3" result_file="$4"
+  local recovery_record="${5:-}"
   local resource_version node_uid
   resource_version="$(jq -er '.metadata.resourceVersion' "${state_file}")"
   node_uid="$(jq -er '.metadata.uid' "${state_file}")"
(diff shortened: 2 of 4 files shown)


devantler:

## Validation evidence at `4568409b`

**RED/GREEN, proven in both directions.** The new tests were run against the *unfixed* scripts first by restoring both from `HEAD` and confirming a zero diff before the run:

- `TestUnrelatedNodeWriteBeforeAtomicClaimIsRetried` — **RED** on unfixed:
  `command exit = 1, want 0` / `::error::Could not atomically claim and cordon Talos node prod-worker-1` / `cordon-claim: resourceVersion test failed`. **GREEN** after the fix.
- `TestRetriedClaimStillRefusesWhenTheRetryBudgetIsExhausted` — passes **both** before and after, deliberately. It is the fail-closed negative control, not a regression test; a change that made it newly pass would mean the safety property had been weakened.

Full suite: **GREEN, 116.3s**, including the pre-existing `TestConcurrentCordonBeforeAtomicClaimStopsTheRoll`, which is the second negative control — a genuine concurrent cordon must still refuse. It does, and it now refuses *faster*: the re-read sees `spec.unschedulable=true ≠ was_cordoned`, so the precondition check bails immediately rather than burning the retry budget.

**Diagnosis was measured, not assumed.** Read live from the failing node's `managedFields` (`kubectl --context admin@prod`, read-only): five independent writers, with `kubelet` having updated `status` ~3 minutes before sampling. That is what makes a whole-object `resourceVersion` test unstable in a way unrelated to drain safety.

**Lint** — `shfmt -i 2 -ci -d` and `shellcheck -x` clean on both changed shell files; `gofmt -l` clean. An untouched repo script was run as a control to confirm the linters were actually exercising repo settings rather than silently no-opping.

### Reviewed my own diff for the seams that matter

- `((attempt >= max_attempts))` returns exit 1 when false, which under `set -e` would abort the script — it is inside an `if` condition, where `set -e` is suspended, so it is safe. The green suite corroborates this.
- `CORDON_CLAIM_MAX_ATTEMPTS` only bounds the retry count. It cannot disable the refusal; setting it to `1` reproduces exactly the old behaviour.
- `build_and_apply_cordon_claim` is referenced before its definition in file order, which is fine because both are defined at source time, well before any invocation.
- The retry re-reads into the same `cordon_state_file` the caller already re-reads via `revalidate_node_scheduling_guard`, so no caller observes a staler view than before.

### Not verified here

The fix is proven against the fake `kubectl` harness, not against a live 422 from a real API server. The harness models the CAS by comparing the patch's `test` value to the stored resourceVersion, which is the same mechanism the real server enforces — but the end-to-end proof will be this PR's own `merge_group` deploy actually completing.

devantler:


@coderabbitai review

devantler:

## Readiness re-verified independently at `4568409b`

I re-ran the proof myself in a clean clone rather than relying on the earlier evidence note.

**1. Programmatically tested.** Full package green at head: `ok … 114.440s`, whole suite.

**2. RED/GREEN by ablation** — reverted *only* the two shell scripts to base and kept the new tests, so the mechanism moved rather than being deleted:

- `TestUnrelatedNodeWriteBeforeAtomicClaimIsRetried` → **FAILS**, with exactly the production symptom:
  `::error::Could not atomically claim and cordon Talos node prod-worker-1; refusing to drain it.` / `cordon-claim: resourceVersion test failed`
- `TestRetriedClaimStillRefusesWhenTheRetryBudgetIsExhausted` → passes in **both** arms, which is correct: base always refuses, so only the benign-conflict arm should move. The refusal direction is genuinely still pinned.
- Control restored → both green again, tree clean.

**3. Tried as a user.** The exercised behaviour is the one the deploy actually hits: a benign concurrent Node write during a cordon claim. It now proceeds, while a genuine conflict (another cordoner, ownership/taint change, node replacement) and an exhausted retry budget still fail closed. That is the merge-queue eviction cause this repo has been losing deploys to.

**Review gate:** `cr@4568409b51` — CodeRabbit completed at head (status `Review completed` 16:25:13Z; summary comment updated 16:25:12Z after the 16:23:13Z request, *No actionable comments were generated*), 0 unresolved threads, 0 finding sections.

Promoting and merging.

devantler:

CodeRabbit completed at this head without producing a gate-satisfying artifact — zero review objects on the PR, and its auto-generated summary does not name the head. Per the review-lane ladder that is a completed no-gate outcome, not a finding to fix, so this advances to the next lane rather than re-spending CodeRabbit.




@codex review

devantler:

Codex reported a usage limit for code reviews at 19:22:41Z with no retry window, so that lane will not deliver at this head. Advancing to the third lane.

devantler:
@cursor review

devantler:

**Reviewed commit:** `4568409b511269eec209abf8c3580d0819fbf8d6`

Per-lane evidence, each checked directly on this PR at this exact head:

| Lane | Outcome | Evidence |
|---|---|---|
| CodeRabbit | completed, no gate-satisfying artifact | request marker 16:23:13Z; head status `success :: Review completed` but **zero review objects** on the PR, and the auto-generated summary does not name this head. Status proves a run completed, not that a review was produced. |
| Codex | usage limit | `chatgpt-codex-connector[bot]` 19:22:41Z — usage limits reached for code reviews, no retry window |
| Cursor Bugbot | usage limit | check-run `Cursor Bugbot` `conclusion=neutral` / `output.title=Error` (run never started), plus `cursor[bot]` 19:26:10Z "usage limit reached — an admin must raise the Cursor cap" |

### Findings

**P2 (Minor) — a refused claim can report no cause.** The retry re-reads the Node to decide whether a rejected claim is still safe to retry, and points that read's stderr at the same `result_file` holding the conflict output. A successful read writes nothing, truncating it, and `emit_safe_operation_output` returns early on an empty file — so the "preconditions no longer hold" break emits a bare refusal with no reason, in the concurrent-actor case that most needs one.

Verified, not inferred: reproduced by assertion, with the mechanism moved rather than the test deleted — restoring the shared sink fails `expected document to contain "cordon-claim: "`, and the fix passes. The exhausted-budget and failed-re-read paths are unaffected and still report correctly.

**Not fixed in place:** this PR is at merge-queue position 3, so its branch rejects pushes (`protected branch hook declined`). Dequeuing a CLEAN queued PR to improve an error message is disproportionate, so the fix is #2983, gated to land after this merges. Issue: #2982.

### What I checked and found sound

- The retry budget is correct: the apply runs on attempts 1–5, and exhaustion still refuses without leaving an owner marker.
- `node_claim_preconditions_still_hold` is a genuinely conservative precondition set — identity, deletion, both ownership annotations, scheduling intent, and taints excluding the unschedulable taint. A concurrent cordon, ownership grab, taint edit or node replacement all still refuse rather than retry.
- The fakes exercise both a clearing conflict and a permanent one, and assert the claim precedes the drain.

**Verdict: no P0/P1 findings** (1 × P2, addressed in #2983).

PR #2992, opened 2026-08-06
ci: let the main-validation gate finish instead of cancelling it

devantler:

## Why

`validate-main.yaml` exists to catch one specific failure: a direct push to `main` fires neither `ci.yaml`'s nor `cd.yaml`'s authorization gate, so `main` could carry a broken authorization surface while all of its own checks show green. Its header records that happening once already, invisible for two days.

The workflow currently cancels its own in-progress runs. Since it only ever runs on push to `main`, every run shares one concurrency key — so a second push evicts the first push's gate before it finishes, and a cancelled check is not a failed check. That reopens the very hole the workflow was written to close, from the other side.

## What

Runs of the gate are allowed to complete, so each push to `main` gets its own verification record. This matches the "non-preempting production lock" the repo already applies to its other production-critical workflow.

No behaviour changes for pull requests — this workflow does not run on them.

Part of devantler-tech/monorepo#2690

### DIFF
--- a/.github/workflows/validate-main.yaml
+++ b/.github/workflows/validate-main.yaml
@@ -22,12 +22,15 @@ on:
 permissions: {}
 
 concurrency:
-  group: validate-main-${{ github.ref }}
-  # This workflow only ever runs on push to main, so every run shares one group
-  # key. Cancelling would let a second push evict the first push's gate result
-  # and leave main green on a check that never finished — reopening the exact
-  # hole described above, from the other side. Each push is its own subject and
-  # its result is its own record, so runs here are allowed to complete.
+  # Every push to main is its own subject and its gate result is its own record,
+  # so each run gets its own group and none of them can suppress another.
+  #
+  # Disabling cancellation is not enough on a shared key: a concurrency group
+  # holds one running plus one pending run, and a third push cancels the pending
+  # one. With a shared `github.ref` key the second push's gate would still be
+  # evicted while queued, leaving main green on a check that never finished —
+  # reopening the exact hole described above, from the other side.
+  group: validate-main-${{ github.run_id }}
   cancel-in-progress: false
 
 jobs:
--- a/.github/workflows/validate-main.yaml
+++ b/.github/workflows/validate-main.yaml
@@ -23,7 +23,12 @@ permissions: {}
 
 concurrency:
   group: validate-main-${{ github.ref }}
-  cancel-in-progress: true
+  # This workflow only ever runs on push to main, so every run shares one group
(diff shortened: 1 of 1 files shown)


devantler:



Requesting review at `1553eb6d71ff`. CI is fully settled: **31 checks, 0 failing**, including `🧹 Lint - mega-linter` (which runs `actionlint` over the changed workflow) and `🧪 Validate Manifests`.

Note for the reviewer: this already incorporates the pending-run correction found on the sibling PR devantler-tech/ksail#6506 — a concurrency group holds one running *plus one pending* run and a third arrival cancels the pending one, so `cancel-in-progress: false` on a shared `github.ref` key is not sufficient. This workflow only triggers on push to `main`, so the group is keyed on `github.run_id` outright rather than with a ref conditional.

@coderabbitai review

devantler:

## Readiness

**Programmatically tested.** CI settled at `1553eb6d71ff`: 19 passing, 0 failing, 12 skipped. `mergeStateStatus: CLEAN`, 0 unresolved threads, 0 non-thread review findings, no conflict with base. All commits `verified=true reason=valid` (this repo enforces `required_signatures`). `actionlint` clean locally, and the repo's existing `scripts/validate-merge-group-heal` tests still pass — that validator asserts the same "non-preempting production lock" doctrine this change applies.

**Reviewed.** Green CodeRabbit at the current head — summary comment reports "No actionable comments were generated in the recent review" and names the reviewed range ending at `1553eb6d71ff7833d53e68c9c369b72b228772c1`; 0 inline comments; commit status `success :: Review completed`. It independently confirmed the mechanism: *"The `github.run_id` concurrency key gives each push a separate group. GitHub Actions cannot cancel an earlier validation run as a result of a later push."*

**Tried and evaluated as a user — with one limit stated plainly.** This workflow triggers only on push to `main`, so its runtime effect is not observable from a pull request at any head; there is no pre-merge surface to exercise. What I did verify:

- The expression form was exercised live on the sibling PR devantler-tech/ksail#6506, where two pushes 73 seconds apart produced `cancelled` for the superseded run and a live run for the newer one — proving GitHub parses and evaluates a ref/run-scoped concurrency expression as intended rather than failing the run.
- `actionlint` and CI's `🧹 Lint - mega-linter` both accept the changed file.

**The remaining half is post-merge**, and I am not claiming it now: acceptance criterion 3 of devantler-tech/monorepo#2690 — two closely-spaced pushes to `main` both retaining a completed gate result — can only be observed after this lands. The tracking item stays at 📊 Verifying rather than Done until that is seen.

Promoting on that basis.

Issue #2942, opened 2026-08-04
Publication contract checks the signed publisher exists, but not that it is the ONLY writer of the mutable tag


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`validateCDWiring` establishes that the shared deploy action still invokes the nested publisher as an
enforced step:

```go
requireEnforcedStep(
    map[string]any{"steps": runs["steps"]}, usesAction(nestedPublisherAction), …)
```

That is an **existence** check. It proves a checked publisher is present and enforced; it does not
prove it is the **only** step that writes the mutable production tag. Appending a step that runs
`workload push …:latest` directly, after the nested publisher, leaves the validator passing.

Reported by Codex on #2939 (P2, `main.go:255`), which states it reproduced the append. Confirmed by
reading the check: there is no enumeration of other writers anywhere in the function.

## Why it matters

The contract exists so the digest reconciliation consumes is the signed and attested one. A second
writer to the same mutable tag can replace that digest immediately before reconciliation — the signed
artifact is published, and then something else overwrites the tag it is published under. The ordering
guarantee #2939 establishes holds, and is then bypassed at a different point.

Not reachable by an outside contributor — it needs a workflow or composite-action edit — so this is
defence-in-depth of a gate that already exists, not a live exposure.

## Expected behaviour

Publication of the mutable production tag is constrained to the checked action, or any other step
that writes it is explicitly rejected.

## Acceptance criteria

1. A step other than the checked publisher that writes the mutable production tag fails the check.
2. The rule covers the shared composite action and the workflow that calls it.
3. The shipped configuration passes unchanged.
4. One ablation per rule, each failing a disjoint test, each build-checked.

## Size

Small–medium. Reuses the existing step-walking helpers; the new part is enumerating writers rather
than finding one.

Deliberately **not** folded into #2939: that PR is at review round 11 on a single concern (ordering on
the direct-push route), and this is a distinct property (exclusivity). Third spin-off from that
review, after #2940 and #2941.

Part of #2627

Issue #2941, opened 2026-08-04
Parse the DR rebuild publisher step instead of scanning workflow text


devantler:
> 🤖 Generated by the Agentic Engineer

**Evidence.** `validateDRWorkflow` establishes the DR rebuild's publication wiring by scanning workflow **text**, not by parsing it. `containsLine(job, "uses: ./.github/actions/deploy-prod/publish-platform-manifests")` is satisfied by that string appearing anywhere in the job — including inside a comment or a block scalar. Reported by CodeRabbit as a P1 during review of #2939, and verified here: the symbols carrying this (`validateDRWorkflow`, `extractJob`, `containsExactLine`, `containsLine`) are untouched by that PR, so this is pre-existing rather than a regression it introduced.

**Impact.** The disaster-recovery route can be pointed at a different publisher while the validator still finds the expected text and reports success — so DR delivery could publish through logic this contract never checks, which is the exact drift the shared-action requirement exists to prevent. The same reasoning applies to the credential assertions (`ghcr-token`, `hcloud-token`), which are matched the same way and so are not proven to sit on the publishing step at all.

**Why this is its own issue.** #2939 converted the *direct-push* and *merge-queue* routes from text scanning to parsing and enforced-step checks. This is the same class at the third route, outside that PR's diff, and #2939 is already at review round 10 on a single concern.

**Expected behaviour.** Parse `dr-rebuild.yaml`, locate `jobs.rebuild.steps`, and require an *enforced* step (neither skipped nor failure-suppressed) whose `uses` value **equals** `nestedPublisherAction` — reusing `requireEnforcedStep`/`usesAction`, which #2939 already introduced for exactly this. Assert `with.ghcr-token` and `with.hcloud-token` on that same step rather than anywhere in the job.

**Acceptance criteria.**
- [ ] `validateDRWorkflow` resolves the rebuild job and its publisher step through the YAML parser, not string containment.
- [ ] The publisher step's `uses` is matched by equality, and the two credentials are asserted on that step.
- [ ] The step is proven enforced (no `if:` skip, no `continue-on-error`).
- [ ] Ablations covering: the real `uses` replaced while the expected text survives **in a comment**; the same **in a block scalar**; credentials moved to a different step; the publisher step skipped/failure-suppressed. Each must be refused.
- [ ] The permission assertions (`id-token`/`packages`/`attestations: write`) move to parsed lookups too, or the issue records why they stay textual.

[...]

PR #2997, opened 2026-08-06
security(ci): assert the DR identity against parsed allow-list entries

### LINKED ISSUE #2945: DR identity allow-list is satisfied by a comment, so the gate can certify an allow-list that is not in effect

## Evidence

`validate-dr-signing` decides whether the DR rebuild identity is allowed by scanning the cluster config **as raw text** (`containsLine`), so any line containing the subject regex satisfies it — including a comment.

Measured 2026-08-04 against `ksail.prod.yaml`, changing only where the identity appears:

| config | validator |
|---|---|
| DR matcher present in `matchOIDCIdentity` (real) | `contract passed`, exit 0 |
| DR matcher **removed** | `does not allow the DR rebuild identity`, exit 1 |
| DR matcher **removed**, identical regex added as a `#` comment | **`contract passed`, exit 0** |

The middle row is the positive control: the check does fire when the identity is genuinely absent. The third row is the bypass — a comment has no effect on the cluster, and the gate accepts it.

A second, related shape: the same text scan passes when the whole `verify` block sits at a path KSail discards. `validate-dr-signing` reported `contract passed` on a config reproducing the #2627 outage exactly.

## Problem

The DR contract exists so a disaster-recovery rebuild cannot publish an artifact production will refuse — the one failure that is unrecoverable by hand, discovered mid-incident. A check satisfied by inert text certifies an allow-list that may not be in effect.

The second shape is largely covered once #2920's read-path check runs in the same job, but the comment bypass is not: that check confirms the block resolves, not that this identity is inside *it*.

## Expected

The DR identity is asserted against the parsed `matchOIDCIdentity` entries at the path KSail reads, not against the file's text.

## Direction

Parse the config (the repo already depends on `gopkg.in/yaml.v3`) and compare `spec.workload.flux.verify.matchOIDCIdentity[].subject` values exactly, rather than substring-matching lines. This is the same "parse, don't scan" correction `validateCDWiring` already documents for workflows — its four enumerated scan defeats are the same class as this one, so the reasoning is already settled in-repo and only the config half was left scanning.

Worth auditing the other `containsLine`/`lineIndexContaining` call sites in that file while there, since they share the primitive.

## Acceptance criteria

- [ ] The identity in a comment fails the check
- [ ] The identity at a path KSail does not read fails the check
- [ ] The check still passes on the real `ksail.prod.yaml`
- [ ] Ablation recorded: each new arm is shown to fail on the broken input, with a control that stays green

Part of #2627. Size: S.

devantler:

## Why

The check that guarantees a disaster-recovery rebuild can publish something production will accept
could be satisfied by text that has no effect on the cluster — including a comment. That matters
because a rejected artifact during a DR rebuild is the one failure nobody can fix by hand, and it
would only be discovered mid-incident. The gate was reporting "passed" on four different
configurations that would all leave a DR rebuild unverifiable.

## What

The gate now reads the allow-list as configuration instead of as text, and requires the DR rebuild's
signer identity to be present exactly — both halves of it. A commented-out identity, an allow-list
placed where KSail ignores it, a wider pattern that merely contains the right one, and the right
identity issued by someone else are all now failures rather than passes. No behaviour outside this
one check changes, and the real production config passes unmodified.

Fixes #2945

### DIFF
--- a/scripts/validate-dr-signing/main.go
+++ b/scripts/validate-dr-signing/main.go
@@ -248,6 +248,11 @@ func describePublicationStep(step map[string]any, index int) string {
 // path KSail renders onto the root OCIRepository. Decoding the exact path is
 // what makes a block sitting anywhere else a failure rather than a pass, because
 // such a block yields zero entries here.
+//
+// Decoding reads the FIRST YAML document only, which is what a KSail cluster
+// config is. A verify block hidden in a later document would therefore yield no
+// entries and FAIL this gate rather than satisfy it — the safe direction, and the
+// same direction every other shape below fails in.
 type clusterVerifyConfig struct {
 	Spec struct {
 		Workload struct {
--- a/scripts/validate-dr-signing/main.go
+++ b/scripts/validate-dr-signing/main.go
@@ -37,6 +37,11 @@ import (
 // DR-published artifact.
 const drIdentitySubject = `^https://github\.com/devantler-tech/platform/\.github/workflows/dr-rebuild\.yaml@refs/heads/main$`
 
+// drIdentityIssuer is the OIDC issuer half of that identity. cosign matches an
+// identity on the issuer/subject PAIR, so a correct subject under a different
+// issuer does not admit the DR signature and must not satisfy this gate.
+const drIdentityIssuer = `^https://token\.actions\.githubusercontent\.com$`
+
 // cdContractGateJob is the cd.yaml job that runs this validator on the
 // direct-push production path. Named once so the wiring check and the workflow
 // cannot drift apart silently.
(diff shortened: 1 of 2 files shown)


devantler:



@coderabbitai review

devantler:

## Readiness — self-promoting

**Programmatically tested.** RED first: all four bypasses reproduced against the scanning version,
each failing with a *distinct* message (so this is not one assertion firing four times) —
comment-carried identity, block at `spec.cluster.verify`, a widened regex merely containing the
identity, and the right subject under the wrong issuer. GREEN after the change, full package suite
clean. Two controls guard the arms: a **positive control** (the identity at the path KSail reads must
be accepted, so a rejection cannot be my synthetic YAML being malformed) and the pre-existing
`TestRealClusterConfigAllowsTheDRIdentity`, which proves the parse reaches the real path in the real
`ksail.prod.yaml`. The new message assertion was itself **ablated** — pointed at the other error
branch it fails, restored it passes — so it is live rather than decorative.

**CI green at `5263a58a`:** 21 success / 0 failing / 8 filtered-skip. The jobs that actually consume
this change were confirmed **SUCCESS by name** rather than inferred from "0 failing" (a skipped job is
invisible in a rollup): `🧪 Test`, `🏗️ Build`, `🧹 Lint - golangci-lint`, `🔍 Dead Code Analysis`,
`📊 Code Coverage`, `🧹 Lint - mega-linter`.

**Reviewed:** `cr@5263a58a` — CodeRabbit review completed at this head with no actionable finding.
Zero review threads, zero inline comments, and the only collapsed section in its summary is
`🔇 Additional comments (4)`, the non-actionable one. Status description reads `Review completed`
(not the `Review skipped` default, and no quota marker), and the summary comment was created
07:44:13Z, after the 07:42:44Z request.

**Tried and evaluated as a user** — ran the real CLI, not just `go test`:

```
DR publication contract passed.                                             # exit 0

$ # same, with the DR identity commented out (the bypass this fixes)
dr publication contract: cosign matchOIDCIdentity at spec.workload.flux.verify has 3 entries but
none is exactly the DR rebuild identity, so a DR-published artifact stays unverifiable; add an entry
with issuer <issuer> and subject <subject> (both compared exactly — a wider regex containing this
one is not accepted)                                                        # exit 1
```

The failure names the exact entry to add, so a developer who trips it can fix it from the log without
reading the validator.

**Scope note.** Only the config-side check changed. The `containsLine` primitive is still used by the
workflow checks in this file and is untouched — parsing those is #2941 and #2942, deliberately not
mixed in here.

PR #2998, opened 2026-08-06
security(ci): parse the DR publisher step instead of scanning workflow text

### LINKED ISSUE #2941: Parse the DR rebuild publisher step instead of scanning workflow text

devantler:

## Why

The disaster-recovery rebuild is the one route to production that runs when the cluster is gone —
the moment nobody can afford to debug a supply-chain policy by hand. Its publication wiring was the
last of the three routes still established by **scanning the workflow's text** rather than reading
it, and a scan cannot tell the difference between a step that runs and a string that merely appears.

Five ways the gate said "wired" about a DR route that was not — each one reproduced, and each one
**accepted by the old check**: the publisher named only in a comment, or inside an `echo`, with its
credentials sitting on an unrelated step, or with the step switched off entirely so its failure is
skipped or discarded. In every case the workflow reads correct and the contract reports success.

Not reachable by an outside contributor — it needs a workflow edit — so this hardens a gate that
already exists rather than closing a live exposure.

## What

The DR route is now parsed like the other two: the publisher is located as a real step whose action
matches exactly, proven to be neither skipped nor failure-suppressed, and its two credentials are
asserted on **that step** instead of anywhere in the job.

The three permission grants move to parsed lookups as well. The old check compared each against a
whole line *including its trailing comment*, so rewording a comment broke the gate while the grant
itself was untouched — a contract that fails on a documentation edit is one people learn to weaken.

Shipped configuration passes unchanged; no workflow or cluster config is modified.

Fixes #2941
Part of #2627

### DIFF
--- a/scripts/validate-dr-signing/main.go
+++ b/scripts/validate-dr-signing/main.go
@@ -97,37 +97,164 @@ const (
 	validatorCommand      = "go run ./scripts/validate-dr-signing .github/workflows/dr-rebuild.yaml ksail.prod.yaml"
 )
 
+// drRebuildJob is the dr-rebuild.yaml job that reaches production.
+const drRebuildJob = "rebuild"
+
+// drRequiredPermission is a job-level grant the rebuild job must carry, together
+// with what stops working without it.
+type drRequiredPermission struct {
+	name string
+	// consequence completes "so …", naming what the DR rebuild cannot do.
+	consequence string
+}
+
+// drRequiredPermissions are the three write grants the publication transaction
+// needs. Matched on the PARSED key and value, so the trailing comment on each
+// line is documentation rather than part of the contract.
+var drRequiredPermissions = []drRequiredPermission{
+	{"id-token", "keyless cosign signing cannot mint a Fulcio certificate"},
+	{"packages", "the publication transaction cannot update GHCR"},
+	{"attestations", "the SBOM and provenance cannot be recorded"},
+}
+
+// drPublisherCredential is an input the publisher step must receive, and the
+// exact expression it must receive it from.
+type drPublisherCredential struct {
+	input      string
+	expression string
+}
+
+// drPublisherCredentials are the two secrets the shared publication action needs.
+// Asserted on the publisher STEP's `with` mapping, not anywhere in the job.
+var drPublisherCredentials = []drPublisherCredential{
+	{"ghcr-token", "${{ secrets.GHCR_TOKEN }}"},
+	{"hcloud-token", "${{ secrets.HCLOUD_TOKEN }}"},
+}
+
+// validateDRWorkflow checks that the disaster-recovery route publishes through
+// the same checked action as every other route, with the grants and credentials
+// that action needs.
+//
+// 🔴 THE WORKFLOW IS PARSED, NOT SCANNED — the third route, and the last one to
+// be converted. The two production routes were parsed in #2939; this one still
+// asked `containsLine`, which is satisfied by the expected text appearing
+// ANYWHERE in the job's text. Five ways that said "wired" about a DR route that
+// was not, each reproduced as an ablation and each returning nil against the
+// scanning version:
+//
+//  1. The `uses:` line survives as a COMMENT while the step invokes a different
+//     action. A comment runs nothing, so the gate certified a publisher that is
+//     not called.
+//  2. The same text inside a BLOCK SCALAR — an `echo` in a run block — with no
+//     `uses:` step anywhere.
+//  3. The credentials named on a DIFFERENT step. A scan over the job text cannot
+//     tell which step receives them, so the publisher could be invoked with none.
+//  4. The publisher step carries `if: false`, so it never runs while the workflow
+//     reads as correct.
+//  5. The publisher step carries `continue-on-error: true`, so it runs, fails,
+//     and the rebuild continues to promote the mutable tag anyway.
+//
+// (4) and (5) are the requireEnforcedStep class this file has now fixed at six
+// sites; matching and enforcing are inseparable there for exactly this reason.
+// A DR publish that silently skipped its evidence is unrecoverable by hand at
+// the one moment nobody can afford to debug it.
 func validateDRWorkflow(workflow string) error {
-	job, ok := extractJob(workflow, "  rebuild:")
+	document, err := decodeWorkflow(workflow)
+	if err != nil {
+		return fmt.Errorf(
+			"disaster-recovery workflow does not parse, so its publication wiring cannot be established: %w", err,
+		)
+	}
+	jobs, ok := document["jobs"].(map[string]any)
+	if !ok {
+		return errors.New("disaster-recovery workflow has no jobs mapping")
+	}
+	job, ok := jobs[drRebuildJob].(map[string]any)
 	if !ok {
 		return errors.New("missing rebuild job")
 	}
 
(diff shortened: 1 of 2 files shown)


devantler:



@coderabbitai review

devantler:



@codex review

devantler:

devantler:
@cursor review

devantler (commented):

**Reviewed commit:** `0bd41dbf7ec394bcb8336a029cfee2e451953482`

All three external lanes were tried at this exact head, in contract priority order, one at a time, and each returned direct per-PR evidence of being unable to deliver:

| Lane | Evidence at this head | When |
|---|---|---|
| CodeRabbit | acked ("I will review the changes"), then a `Review limit reached` shell 4s later; commit-status description `Review rate limited`; **0** review objects | 08:24:41Z → 08:24:45Z |
| Codex | comment: "You have reached your Codex usage limits for code reviews"; **0** review objects; no reaction on the trigger | 08:28:27Z |
| Cursor Bugbot | check-run `Cursor Bugbot` `conclusion: neutral` + `output.title: Error`, summary `Bugbot run failed`; `cursor[bot]`: "Bugbot couldn't run - usage limit reached"; **0** inline comments | 08:34:11Z → 08:34:13Z |

Bugbot's limit is a Cursor **spend** limit with no retry window — only an admin can raise it — so waiting would not have produced a review this run.

### What I checked

- **The RED half is real, not asserted.** Of the 11 ablations, **five return `nil` against the pre-change scanning implementation**: publisher named only in a comment, only inside a block scalar, credentials on a different step, step `if: false`, step `continue-on-error: true`. Each ablation *moves* the mechanism (the expected text survives somewhere inert) rather than deleting it, so it separates a scan from a parse instead of merely proving something is required.
- **No direction is weakened.** `usesAction` compares the whole `uses` value for equality, where the old `containsLine` was a substring match — so `…/publish-platform-manifests-evil` used to pass and is now refused. Credentials moved from "anywhere in the job text" to "on the publisher step, exact expression". The one deliberate loosening is that a permission's trailing **comment** is no longer part of the contract; the grant itself is still required to be exactly `write`.
- **Fail-closed shapes** are covered: unparseable workflow, no `jobs` mapping, renamed job, absent `permissions`, and a blanket `write-all` scalar (refused — it grants enough today but names none of the three, so a later narrowing could drop one silently).
- **The consuming job actually ran**, rather than counting "0 failing": `🖋️ Validate Publication Contract` = **pass**, alongside `🔍 Dead Code Analysis` (confirming `extractJob`/`containsExactLine` removal left nothing dangling) and `golangci-lint`. 30 checks, 0 failing.
- **End-to-end, both directions**, through the real CLI rather than the unit tests alone: the shipped config passes; a comment-ablated copy placed inside the repo tree (so derived publisher/cd/ci paths resolve) is refused with the correct message. My first attempt at this control was invalid — it failed on a path error rather than the ablation — and was redone.

### Considered and dismissed (recorded so a later round does not re-raise them)

- **"The publisher step is enforced but not proven *reached*."** Not a hole here: GitHub runs each step as its own process, so an early `exit 0` in a preceding `run` step ends that step successfully and later steps still run. The gate-job allowlist used for `validate-publication-contract` would be wrong for a long operational job like `rebuild`.
- **The rebuild job's own `if:` is deliberately not refused** — it is the supersession gate, a legitimate and necessary condition, unlike cd.yaml's `deploy-prod`. The step-level enforcement is what matters and is checked.

devantler:

## Readiness — all three conditions, at `0bd41dbf7ec394bcb8336a029cfee2e451953482`

**1. Programmatically tested.** RED/GREEN, not asserted: 11 ablations, of which **5 return `nil`
against the pre-change implementation** (publisher in a comment; in a block scalar; credentials on a
different step; `if: false`; `continue-on-error: true`). Full package suite green, `go build ./...`
green, `go vet` and `gofmt` clean. CI **30 checks / 0 failing** — and the job that actually consumes
this change, `🖋️ Validate Publication Contract`, is **pass** rather than skipped, alongside
`🔍 Dead Code Analysis` (proving the `extractJob`/`containsExactLine` removal left nothing dangling).

**2. Reviewed.** Local review round (id 4872669913, `COMMENTED`) clean at exactly this head, after
CodeRabbit, Codex and Cursor Bugbot each returned direct per-PR evidence of being unable to deliver —
the per-lane table is in that review.

**3. Tried and evaluated as a user.** This change has an exercisable runtime surface, so I ran the
real command rather than tracing it:

```
DR publication contract passed.                                    # exit 0
```

and the negative direction through the same CLI, with the publisher `uses:` demoted to a comment and
a different action put in its place:

```
dr publication contract: rebuild job must use the shared publication action so
normal and disaster-recovery delivery cannot drift                 # exit 1
```

The ablated copy was placed **inside the repo tree** so the validator's *derived* publisher/`cd.yaml`/
`ci.yaml` paths still resolved, and removed afterwards. Recording that my first attempt at this control
was **invalid** — run from `/tmp`, it exited 1 on a missing-file error rather than on the ablation, which
would have looked like a pass for the wrong reason. It was caught by asserting the ablation had actually
applied (`grep -c` = 0) and redone.

Promoting on that basis.

PR #3004, opened 2026-08-06
fix(rbac): enumerate the github-config tenant Role's managed-resource kinds

### LINKED ISSUE #3003: The github-config tenant Role grants a resources wildcard over the org's repository authority

## Evidence

`checkov` on `k8s/` currently reports **37** failures (1745 passed, 6 skipped — measured
2026-08-06; the `.mega-linter.yml` comment's "73" is stale). Grouped by check id, exactly one is
`CKV_K8S_49` "Minimize wildcard use in Roles and ClusterRoles", and it is the only RBAC wildcard
finding in a **first-party** manifest:

```
CKV_K8S_49  k8s/bases/apps/github-config/role.yaml  Role.github-config.github-config-managed-resources
```

The Role grants `resources: ["*"]` across five `*.github.m.upbound.io` API groups.

This finding is cleanly separable from the rest of #2787's backlog:

- The 11 `CKV_K8S_40` findings are already dispositioned by #2898 / #2904.
- The `CKV_K8S_155` and `CKV_K8S_38` findings sit in **vendored upstream** CDI/KubeVirt manifests,
  where #2861 owns the re-vendor-durable verdict.
- Unlike the pod-security families, an RBAC rule is **not** subject to #2787's open "checkov reports
  on a pre-mutation view" question — Kyverno injects no RBAC, so there is no ambiguity to resolve
  first.

## Audience and impact

The `github-config` tenant ServiceAccount holds authority over the org's **real repositories**. The
wildcard grants it every current and future resource in those five groups, including kinds nobody
has reviewed, the moment a new managed-resource definition is activated.

## Expected behaviour

The Role names the managed-resource kinds it needs, and that set stays in lockstep with the
`ManagedResourceActivationPolicy` that decides which kinds exist at all.

## Acceptance criteria

- [ ] `k8s/bases/apps/github-config/role.yaml` enumerates its managed-resource kinds; `CKV_K8S_49`
      passes on the file.
- [ ] Every kind the tenant's Flux Kustomization currently has in its inventory is still granted
      (verified against live prod, so the narrowing cannot break the running app).
- [ ] A CI check fails, naming the offending kind, when the Role and the activation policy diverge
      in **either** direction, so the added maintenance step cannot be silently forgotten.
- [ ] The parity check fails closed if either of its queries stops matching, rather than passing
      vacuously on two empty sets.

Rough size: small. Part of #2787.

devantler:
## Why

The `github-config` tenant holds authority over the org's **real repositories**, and its Role
granted a blanket wildcard over five GitHub API groups — so it automatically gained every new
resource kind the moment one was activated, reviewed by nobody.

## What

The Role now names the kinds it needs, matching the activation policy that decides which kinds
exist at all. A CI check fails — naming the offending kind — if the two ever drift apart, so the
extra maintenance step can't be silently forgotten.

Verified against live prod: every kind the tenant currently has deployed is still granted, so this
cannot break the running app.

Fixes #3003
Part of #2787

> [!IMPORTANT]
> **Merge-order gate:** devantler-tech/platform#2985 and devantler-tech/platform#3004 both re-approve the same `expectedRenderedSurfaceSHA` from the same base value. Whichever merges second will conflict on that line and must **re-measure and re-approve** against the new `main` — the second PR's hash is not valid once the first lands.

### DIFF
--- a/scripts/tests/test-github-config-role-activation-parity.sh
+++ b/scripts/tests/test-github-config-role-activation-parity.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# The github-config tenant Role must grant exactly the managed-resource kinds
+# the ManagedResourceActivationPolicy turns into CRDs.
+#
+# Activating a kind without granting it leaves the tenant's Flux Kustomization
+# failing at apply time, inside a separately published OCI artifact, with an
+# RBAC error that names the ServiceAccount rather than this file. Granting a
+# kind that is not activated re-widens the tenant's authority past the
+# namespaced surface the Role exists to bound. Both directions are checked.
+#
+# The bare `github.m.upbound.io` group is deliberately out of scope: it carries
+# the ProviderConfig, which is not a managed resource and never appears in the
+# activation policy. The `*.github.m.upbound.io` suffix match excludes it.
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+repo_root="$(cd "${script_dir}/../.." && pwd)"
+role="${repo_root}/k8s/bases/apps/github-config/role.yaml"
+policy="${repo_root}/k8s/providers/hetzner/infrastructure/crossplane/managed-resource-activation-policy.yaml"
+
+for f in "${role}" "${policy}"; do
+  if [[ ! -f "${f}" ]]; then
+    echo "::error::missing input for the github-config RBAC parity check: ${f}"
+    exit 1
+  fi
+done
+
+# `<resource>.<group>` for every activated GitHub managed resource.
+activated="$(
+  yq --no-doc -r \
+    '.spec.activate[] | select(. == "*.github.m.upbound.io")' \
+    "${policy}" | sort -u
+)"
+
+# `<resource>.<group>` for every managed-resource kind the Role grants.
+# shellcheck disable=SC2016  # $g is a yq variable; the shell must not expand it.
+granted="$(
+  yq --no-doc -r \
+    '.rules[]
+       | select(.apiGroups[] == "*.github.m.upbound.io")
+       | .apiGroups[] as $g
+       | .resources[] + "." + $g' \
+    "${role}" | sort -u
+)"
+
+# Fail closed on a vacuous read. A path typo, a renamed field, or a yq upgrade
+# that changes the expression's meaning would otherwise make BOTH sides empty
+# and the comparison below pass while testing nothing.
+activated_count="$(printf '%s\n' "${activated}" | grep -c . || true)"
+granted_count="$(printf '%s\n' "${granted}" | grep -c . || true)"
+
+if ((activated_count < 5)); then
+  echo "::error::read ${activated_count} activated GitHub managed resources from ${policy}; expected at least 5 — the query matched nothing meaningful"
+  exit 1
+fi
+
+if ((granted_count < 5)); then
+  echo "::error::read ${granted_count} granted GitHub managed resources from ${role}; expected at least 5 — the query matched nothing meaningful"
+  exit 1
+fi
+
+# A wildcard would satisfy every parity comparison below while granting the
+# whole group, so reject it explicitly rather than relying on the set diff.
+if printf '%s\n' "${granted}" | grep -q '^\*\.'; then
+  echo "::error::${role} grants a resources wildcard on a GitHub managed-resource group; enumerate the activated kinds instead"
+  exit 1
+fi
+
+missing="$(comm -23 <(printf '%s\n' "${activated}") <(printf '%s\n' "${granted}"))"
+extra="$(comm -13 <(printf '%s\n' "${activated}") <(printf '%s\n' "${granted}"))"
+
+status=0
+
+if [[ -n "${missing}" ]]; then
+  status=1
+  while IFS= read -r kind; do
+    [[ -n "${kind}" ]] || continue
+    echo "::error::${kind} is activated but the github-config Role does not grant it; add its resource name to the matching apiGroup rule in k8s/bases/apps/github-config/role.yaml"
+  done <<<"${missing}"
+fi
+
+if [[ -n "${extra}" ]]; then
+  status=1
+  while IFS= read -r kind; do
+    [[ -n "${kind}" ]] || continue
+    echo "::error::the github-config Role grants ${kind}, which no ManagedResourceActivationPolicy activates; remove it or activate the kind in k8s/providers/hetzner/infrastructure/crossplane/managed-resource-activation-policy.yaml"
+  done <<<"${extra}"
+fi
+
+if ((status != 0)); then
+  exit "${status}"
+fi
+
+echo "github-config Role grants exactly the ${activated_count} activated GitHub managed-resource kinds."
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,6 +80,7 @@ jobs:
               - 'scripts/validate-naming.py'
               - 'scripts/validate-embedded-json.py'
               - 'scripts/tests/test-openbao-oidc-role.sh'
+              - 'scripts/tests/test-github-config-role-activation-parity.sh'
               - 'scripts/tests/test-restrict-tenant-secret-stores.sh'
               - 'scripts/tests/test-kyverno-admission-vpa.sh'
               - 'scripts/tests/kyverno-admission-vpa-rules.yaml'
@@ -302,6 +303,16 @@ jobs:
         # bound_claims is a map. Passing JSON-looking text as one CLI argument
         # serializes it as a string and leaves the role unenforced.
         run: bash scripts/tests/test-openbao-oidc-role.sh
+
+      - name: 🔑 Validate github-config Role covers the activated MR kinds
+        if: needs.changes.outputs.k8s == 'true'
+        # The tenant Role enumerates managed-resource kinds instead of `*`, so
+        # activating a kind without granting it would fail at apply time inside
+        # a separately published OCI artifact. Catch the divergence here. Bash,
+        # yq and the runner's kubectl; no cluster or secrets required.
+        run: |
+          shellcheck scripts/tests/test-github-config-role-activation-parity.sh
+          bash scripts/tests/test-github-config-role-activation-parity.sh
 
       - name: ⚖️ Install Kyverno CLI
         if: needs.changes.outputs.k8s == 'true'
--- a/k8s/bases/apps/github-config/role.yaml
+++ b/k8s/bases/apps/github-config/role.yaml
@@ -9,14 +9,24 @@
 # verbs over the namespaced provider-upjet-github managed-resource API groups,
 # and only within the github-config namespace. Because the managed resources are
 # namespaced (Crossplane v2 `*.github.m.upbound.io`), a plain Role + RoleBinding
-# suffices — no ClusterRole/ClusterRoleBinding, no cluster RBAC. Widen apiGroups
-# here only as new GitHub MR groups are activated in the
-# ManagedResourceActivationPolicy.
+# suffices — no ClusterRole/ClusterRoleBinding, no cluster RBAC.
+#
+# Each managed-resource rule names its resources explicitly rather than `*`, so
+# the Role grants exactly the kinds the ManagedResourceActivationPolicy turns
+# into CRDs and nothing else. That parity is the invariant: extend BOTH this
+# file and
+# k8s/providers/hetzner/infrastructure/crossplane/managed-resource-activation-policy.yaml
+# when a new GitHub MR kind comes under management.
+# scripts/tests/test-github-config-role-activation-parity.sh fails the build
+# naming the missing kind if the two ever disagree, so the omission surfaces in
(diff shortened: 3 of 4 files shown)


devantler:

## Verification record

### RED → GREEN

`checkov -f k8s/bases/apps/github-config/role.yaml` — **3 passed / 1 failed** (`CKV_K8S_49`) before,
**4 passed / 0 failed** after.

### The guard was ablated, not just run green

Each arm changed exactly one thing and fired its own named assertion; the restored tree returned
green:

| Ablation | Assertion that fired |
|---|---|
| drop `issuelabels` from the Role | `issuelabels… is activated but the Role does not grant it` |
| grant an unactivated `teamsettings` | `the Role grants teamsettings…, which no MRAP activates` |
| restore `resources: ["*"]` | `grants a resources wildcard…; enumerate the activated kinds` |
| typo the policy query | `read 0 activated…; expected at least 5` |
| typo the Role query | `read 0 granted…; expected at least 5` |
| typo **both** (the true vacuous case) | fails closed — does **not** pass on two empty sets |

The glob is end-anchored and verified per-case: `repo.github.m.upbound.io` matches;
`github.m.upbound.io` (the ProviderConfig group, deliberately out of scope),
`notgithub.m.upbound.io` and `repo.github.m.upbound.io.evil` do not.

### Exercised as a user, against live prod

The tenant's Flux Kustomization inventory holds **10 distinct group/kind combinations** — every one
is still granted by the enumerated rules, so the narrowing cannot break the running app.
`branchprotections` and `repositorypermissions` are activated but not yet deployed; both are granted,
which is what the parity invariant asserts.

### Authorization pin re-approval

The surface fingerprint moved, as designed. Re-measured to the standard the constant's own note
demands — **515 documents on both sides, membership set difference zero in BOTH directions** over
`apiVersion|kind|namespace|name` (not counts, which cannot see a rename), **133 grant-bearing
documents unchanged**, and every `aws`-bearing line byte-identical. Exactly one entry's content
moved: this Role. Recomputed with the pinned renderer (kubectl `v1.36.2` / kustomize `v5.8.1`,
checksum-verified), so the value is not a local-toolchain artifact.

The consuming CI job ran rather than skipped: `🧪 Validate Manifests` → step
`🔑 Validate github-config Role covers the activated MR kinds` → **success**.

### DevEx

The Role and the activation policy already had to change together — the policy's comments assert the
coupling in prose. This makes the omission a CI failure that **names the missing kind and the file to
edit**, instead of a runtime apply error inside a separately published OCI artifact.

devantler:

**CI red here is a GitHub Actions incident, not this change.**

All four failing jobs at head `6bec724f09` — `🧪 Validate Manifests`, `🖋️ Validate Publication
Contract`, `🔀 Enable Auto-Merge`, `🔍 Scan for Workflow Vulnerabilities` — failed with the identical
signature, and none of them reached a step that reads this diff:

```
Failed to resolve action download info. Error: Service Unavailable
##[error]Service Unavailable
```

githubstatus.com confirms an **active partial outage of GitHub Actions** (opened 15:22Z; at 15:45Z
"some workflow runs are failing to start or failing partway through, and some requests to the Actions
REST API are returning errors"; still unresolved at 15:53Z), including unexpected workflow rate
limiting.

I re-ran the failed jobs once, then **stopped** — re-running into an active outage is futile and the
incident notes rate limiting, so repeated attempts make things worse. No re-queue, no review request
until Actions recovers; a review spent on a head whose CI cannot complete would only have to be
re-secured afterwards.

Next tick: confirm the incident is resolved, re-run whatever is still red, and only then request a
review (CodeRabbit first).

Self-review of the diff is done and found no defect. One thing I checked explicitly, because it
looked wrong: `select(. == "*.github.m.upbound.io")` in the new parity script reads like exact string
equality, which would have made the check vacuous. It is not — `yq`'s `==` glob-matches on strings
(unlike `jq`'s). Verified on v4.53.3 against a fixture with a negative control: it matches
`repositories.repo.github.m.upbound.io` and correctly excludes both the bare `github.m.upbound.io`
ProviderConfig group and an unrelated `*.aws.upbound.io`. The script and its comment are accurate.

devantler:

## The 9 red checks at `6bec724f09` are a GitHub Actions outage, not this change

Every one of them failed in the **`Set up job`** step — before any repository code ran. The
annotation is:

> Failed to resolve action download info. Service Unavailable

Corroborated independently by GitHub's own status API, which reports the **Actions** component in
`partial_outage` at the time of writing.

Affected: `Analyze (actions)`, `🖋️ Validate Publication Contract`, `📦 Tidy`,
`🧹 Lint - golangci-lint`, `🧪 Validate Manifests`, `🔍 Dead Code Analysis`, `🏗️ Build`, `🧪 Test`,
`📊 Code Coverage`.

**What did complete, before the outage, is the evidence that matters:**

- `🔐 Validate EKS Authorization` → **SUCCESS** at this head, so the re-approved surface fingerprint
  is correct in CI and not a local-toolchain artifact.
- On the previous head `156a38d3f`, `🧪 Validate Manifests` → **success**, with the step
  `🔑 Validate github-config Role covers the activated MR kinds` → **success**. The only change since
  is the `main.go` constant plus its comment.

`gh run rerun --failed` is currently refused (`cannot be retried` / `already running`), which is
consistent with the outage. No re-queue attempt beyond that: re-running into a degraded Actions
service just burns minutes. This needs a clean wave once the outage clears — no code change is
implied, and none has been made.

devantler:



Self-reviewed at this head, and the central invariant is verified **behaviourally against the live
cluster**, not just read off the diff. Every enumerated resource matches the kinds actually activated
as CRDs on prod, group for group:

| apiGroup | live CRDs | enumerated in the Role |
|---|---|---|
| `repo.github.m.upbound.io` | branchprotections, defaultbranches, issuelabels, repositories, repositoryrulesets | same 5 |
| `team.github.m.upbound.io` | teammemberships, teamrepositories, teams | same 3 |
| `actions.github.m.upbound.io` | repositorypermissions | same 1 |
| `enterprise.github.m.upbound.io` | organizationrulesets | same 1 |
| `github.m.upbound.io` | providerconfigs, providerconfigusages, clusterproviderconfigs | providerconfigs only |

The `github.m.upbound.io` row is the deliberate one and it checks out: `ProviderConfigUsage` is
written by the provider controller rather than this ServiceAccount, and `ClusterProviderConfig` is
cluster-scoped so a namespaced Role could not grant it regardless. Excluding both is correct.

No P0/P1 from my own pass. Note that every check on this head is `CANCELLED` from the active GitHub
Actions outage rather than from a defect, so CI has not validated anything yet — I am requesting
review in parallel because CodeRabbit does not depend on Actions and the head will not move meanwhile.
This PR stays a draft until CI actually runs green.

@coderabbitai review

devantler:

## Readiness record

**Tried and evaluated against the live cluster.** Re-ran the parity claim directly against
`admin@prod` rather than trusting the CI test alone:

```
kubectl get managedresourceactivationpolicy -o json | jq -r '.items[].spec.activate[]' | grep -i github
```

returns **10** activated GitHub managed-resource kinds. The Role in this PR enumerates **10**, and
they match one-for-one across all four API groups:

| API group | Activated live | Enumerated in Role |
|---|---|---|
| `repo.github.m.upbound.io` | repositories, defaultbranches, branchprotections, repositoryrulesets, issuelabels | same 5 |
| `team.github.m.upbound.io` | teams, teammemberships, teamrepositories | same 3 |
| `actions.github.m.upbound.io` | repositorypermissions | same 1 |
| `enterprise.github.m.upbound.io` | organizationrulesets | same 1 |

No activated kind is ungranted, so the wildcard removal cannot cost the tenant an apply.

**The one red check is infrastructure, not this change.** `Analyze (actions)` failed at 15:47Z with
`Failed to resolve action download info. Error: Service Unavailable` after two retries — the GitHub
Actions incident, before any action of this workflow ran. The job refuses re-run
(`This workflow run cannot be retried`). The repository's only required status check,
`CI - Required Checks`, is green at this head, as is the added parity test.

**Review:** CodeRabbit completed at this head with no actionable findings and 5/5 pre-merge checks.

devantler:



@coderabbitai review

devantler:




@codex review

devantler:




CodeRabbit and Codex are both quota-limited at this head. Requesting Cursor Bugbot next, per lane priority. The bare trigger follows in the next comment (Bugbot exact-matches the whole comment body).

devantler:
@cursor review

scripts/tests/test-github-config-role-activation-parity.sh
```diff
@@ -0,0 +1,95 @@
...
+    "${policy}" | sort -u
+)"
+
+# `<resource>.<group>` for every managed-resource kind the Role grants.
+# shellcheck disable=SC2016  # $g is a yq variable; the shell must not expand it.
+granted="$(
+  yq --no-doc -r \
+    '.rules[]
+       | select(.apiGroups[] == "*.github.m.upbound.io")
+       | .apiGroups[] as $g
```
devantler, line 42: nit — `.apiGroups[] as $g | .resources[] + "." + $g` forms the cross product of a rule's groups and resources. Every rule here carries exactly one `apiGroups` entry, so the output is correct today and this is not a defect. It would silently produce wrong pairs if a future rule ever listed two groups — at which point the parity comparison would report kinds that were never granted. Worth a comment pinning the one-group assumption, or `select(.apiGroups | length == 1)`, if this file is touched again.

devantler (commented):

**Reviewed commit:** `769f4021e35480216e132fd9009640c430c91e09`

All three external lanes were requested at this head, in priority order, and none delivered a review.
Each is a provider quota state, not a verdict on this change, and none states a usable retry window:

- **CodeRabbit** — `Review rate limited`; its comment reads "you've reached your PR review limit, so we couldn't start this review" (00:38Z).
- **Codex** — "You have reached your Codex usage limits for code reviews" (00:44Z).
- **Cursor Bugbot** — check-run `neutral` with `output.title: Error`, comment "Bugbot couldn't run - usage limit reached" (00:45Z). The `Error` title means the run never started, as distinct from a review that found nothing.

### What I checked, behaviourally

**Parity against the live cluster.** `kubectl --context admin@prod` reports **10** activated GitHub
managed-resource kinds; the Role enumerates **10**, matching one-for-one across all four API groups.
No activated kind is left ungranted, so removing the wildcard cannot cost the tenant an apply.

**The new parity test actually enforces, in both directions.** Run at this head it reports
`grants exactly the 10 activated GitHub managed-resource kinds`. Three ablations, each naming the
assertion expected to fire:

| Ablation | Expected | Result |
|---|---|---|
| Drop a granted kind (`issuelabels`) | "activated but the Role does not grant it" | fired, naming the kind |
| Grant a kind nothing activates | "which no ManagedResourceActivationPolicy activates" | fired, naming the kind |
| Restore `resources: ['*']` | the explicit wildcard guard | fired |

The file restores clean and the test passes again afterwards, so the ablations were reversible and
the pass is not stale. The script's fail-closed floors (`< 5` on either side) mean a vacuous read
from a `yq` change or a path typo errors rather than comparing two empty sets.

**Direction of the change.** Strictly a tightening: every resource the new rules admit, the wildcard
already admitted. `providerconfigs` is added under the bare `github.m.upbound.io` group for the
ProviderConfig the tenant applies; that group is deliberately outside the parity check because it
carries no managed resource, and `ProviderConfigUsage` is correctly left out as controller-written.

**CI at this head:** 0 failing, 0 pending, `mergeStateStatus: CLEAN`, 0 unresolved threads. The
earlier reds on this branch were inherited from `main` (#3009) and cleared when #3008 merged.

Verdict: no P0/P1 findings

scripts/tests/test-github-config-role-activation-parity.sh
devantler: > 🤖 Generated by the Agentic Engineer

Refuted as non-blocking, with the reasoning recorded rather than waved through.

Checked rather than assumed: every rule in `role.yaml` carries exactly one `apiGroups` entry today, so the cross product is one-to-one and the output is correct — the test reads the same 10 kinds the live cluster reports.

More importantly it **fails safe**, which is what decides this. If a future rule ever listed two groups, the cross product would produce pairs that are not granted, and the parity comparison would then report them as `extra` and **fail the build**. The wrong direction — a silent pass over a real mismatch — is not reachable this way, and the `< 5` floors already refuse a vacuous read.

So this is a robustness note for whoever next edits the file, not a defect in it. Deliberately not pushing a change for it: the three external review lanes are quota-exhausted at this head, so a new commit would discard the current-head review evidence and leave a production-fix-carrying PR unreviewable for the rest of the window. Captured for a follow-up edit of this file instead.

devantler:

**Parked — do not re-queue.** This PR is pentad-clear (CLEAN, green checks, 0 unresolved threads,
clean current-head review) and has still been **evicted from the merge queue twice**, at
2026-08-07T00:49Z and 02:12Z. Nothing is wrong with this change.

The blocker is repository-wide, not PR-specific: the root `flux-system` OCIRepository fails cosign
verification, so every `merge_group` deploy times out waiting for its artifact to be served and the
PR is evicted. Diagnosis and the structural fix are in #3013.

Re-queuing before that clears just repeats the eviction and burns a prod deploy plus a heal deploy
each time. Wait for the source to report `SourceVerified=True`.

Issue #3005, opened 2026-08-06
A failed merge-group deploy leaves prod on never-merged code, because the heal shares the deploy path and fails with it


devantler:
> 🤖 Generated by the Agentic Engineer

## 🔴 Production delivery is still halted — one cluster write unblocks it

**Status at 2026-08-07 05:2xZ: ~14 hours halted.** Every GitOps delivery to prod is refused. This
needs **one manual cluster write from @devantler**; nothing in CI or Git can reach it, for the reason
in *Why nothing automatic can fix this* below.

**Verified live, this tick:**

| | Value |
|---|---|
| `ocirepository/flux-system` (`flux-system`) | `Ready=False` — `failed to verify the signature using provider 'cosign keyless': no matching attestations: unsupported: multiple identities are not supported at this time` |
| Live `matchOIDCIdentity` entries | **3** (separate list entries) |
| `main` declares | **1** entry, with the three signers OR'd *inside* the subject regex |
| Source of the 3 entries | `fluxinstance/flux` → `spec.kustomize.patches[6]`, applied by a never-merged candidate |

**The fix restores the control, it does not remove it.** `main` already carries the correct
single-entry form (`k8s/providers/hetzner/infrastructure/controllers/flux-instance/flux-instance.yaml`,
patch index 6, same as live). The write below simply makes the live object match `main`.

```bash
kubectl --context admin@prod edit fluxinstance flux -n flux-system
```

In `spec.kustomize.patches[6].patch`, replace the **three** `matchOIDCIdentity` entries with this
single entry — copied verbatim from `main`:

```yaml
matchOIDCIdentity:
  - issuer: '^https://token\.actions\.githubusercontent\.com$'
    subject: '^https://github\.com/devantler-tech/platform/\.github/workflows/(ci\.yaml@refs/heads/gh-readonly-queue/main/.+|(cd|dr-rebuild)\.yaml@refs/heads/main)$'
```

**Then it self-heals, with no drift.** flux-operator re-renders the OCIRepository with one matcher →
cosign verifies → the root source goes `Ready` → `infrastructure-controllers` reconciles and applies
`main`'s FluxInstance, which carries that identical value.

### Why nothing automatic can fix this

The trust configuration is delivered **through the channel it gates**. The OCIRepository's matchers
are rendered by `fluxinstance/flux`, which is owned by the `infrastructure-controllers` Kustomization
(`kustomize.toolkit.fluxcd.io/name` on the live object) — and that Kustomization's source is the very
[...]

PR #3008, opened 2026-08-07
fix(ci): accept the shared cosign matcher in the DR publication contract

### LINKED ISSUE #3009: main fails its own DR publication contract: the two cosign validators are mutually unsatisfiable

## What is wrong

`main` fails its own production publication contract, and every pull request inherits the failure.

Reproduced on unmodified `main` (`0a00b2cd`), no local changes:

```
dr publication contract: cosign matchOIDCIdentity at spec.workload.flux.verify has 1 entries
but none is exactly the DR rebuild identity ...
exit status 1
```

## Why it happens

Two guards became mutually unsatisfiable:

- `validate-flux-verify` requires **exactly one** `matchOIDCIdentity` entry, because cosign rejects
  a multi-entry list outright and would then verify nothing.
- `validate-dr-signing` required an entry **string-equal** to the disaster-recovery identity.

Three signers must be trusted, so the single permitted entry has to express them as an alternation
inside the subject — which can never be string-equal to any one of them. Satisfying either guard
breaks the other.

## Why it went unnoticed

The change that collapsed the allow-list merged while GitHub Actions was degraded. Only the
GitHub-managed run executed on the merge commit; `Validate Main` never ran, so no run has exercised
this contract on `main` since.

## Impact

The contract runs on the merge queue, so **no pull request can reach production while it fails** —
including a pull request that would deploy the correction. Combined with #3005 this is what has kept
delivery halted.

## Acceptance criteria

- [ ] `validate-dr-signing` passes on an unmodified checkout of `main`.
- [ ] A matcher widened with an untrusted signer is still refused (the lengthening direction).
- [ ] A mistyped permitted form is caught rather than compared equal to itself.
- [ ] Both validators pass on the same shipped tree.

Size: small.

devantler:

## Why

`main` currently fails its own production publication contract, and nothing has said so — the
`Validate Main` workflow never ran on the commit that introduced it, so it landed silently.

Two guards now demand incompatible things. One requires the cluster's cosign allow-list to hold
**exactly one** signer entry, because cosign refuses a multi-entry list outright and would verify
nothing. The other required an entry **exactly equal** to the disaster-recovery signer. Three
signers have to be trusted, so the single entry must list them as alternatives — and an entry
listing three alternatives can never be equal to one of them. The two rules cannot both be
satisfied.

The practical cost is larger than one red check: this contract runs on the merge queue, so **no
pull request can reach production while it fails**, including the ones that would deploy the fix.

## What

Compares the allow-list against the **permitted forms** of the matcher rather than against a single
identity, so trusting several signers is expressible again.

The exact comparison is deliberately kept. A widened matcher still admits the DR signer, so a check
that merely asked "does this admit disaster recovery?" would wave through an extra signer — the one
direction this contract exists to refuse. Adding or removing a trusted signer stays a deliberate,
reviewed edit.

A new guard checks each permitted form against the real DR certificate and a battery of untrusted
ones, so a typo in the definition is caught rather than agreeing with itself forever.

Fixes #3009

---

> [!IMPORTANT]
> **Merge order — this PR first, then #3004.**
>
> This one changes only `scripts/validate-dr-signing/`, which is **not** in the `k8s` change filter,
> so its merge-queue run will **not** deploy. It unblocks the queue; it does not touch the cluster.
>
> Production delivery is separately halted (#3005): the live root source declares three cosign
> matchers against `main`'s one, so it fails verification and `infrastructure` /
> `infrastructure-controllers` are stalled. Workloads are serving throughout — a stalled pipeline,
> not an outage.
>
> #3004 *does* change `k8s/`, so merging it **after** this lands runs `ksail cluster update` in the
> merge queue and applies the corrected matcher directly, outside the broken channel — which is what
> heals production. Nothing here should be merged in the other order.

### DIFF
--- a/scripts/validate-dr-signing/main.go
+++ b/scripts/validate-dr-signing/main.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


devantler:



@coderabbitai review

devantler:

## Verification record

### RED, on unmodified `main`

Reproduced before writing anything — a clean worktree at `0a00b2cd`, no local changes:

```
dr publication contract: ... has 1 entries but none is exactly the DR rebuild identity
exit status 1
```

So the defect is in `main` itself, not inherited from a branch. `Validate Main` never ran on that
commit, which is why nothing reported it.

### GREEN

Both validators now pass on the same shipped tree — the point of the change is that they can be
satisfied *together*:

```
$ go run ./scripts/validate-dr-signing ...   # DR publication contract passed.
$ go run ./scripts/validate-flux-verify ...  # exit 0
```

`go test ./scripts/...` passes, `golangci-lint` reports 0 issues, and CI agrees:
`🖋️ Validate Publication Contract` and `CI - Required Checks` are green at this head.

### Ablations — each guard was made to fail on purpose

| Ablation | Assertion expected to fire | Result |
|---|---|---|
| Widen the shipped three-signer matcher with an extra signer | exact-match refusal | fired |
| Drop the leading `^` from a permitted form | over-breadth control naming the boundary escape | fired |
| Permitted entry placed alongside a second entry | multi-entry refusal | fired |
| Issuer that matches nobody / everybody | issuer guard, both directions | fired |
| Remove the permitted matcher from the shipped config | the ablation's own "changed nothing" guard | fired |

The last row matters: the two pre-existing ablations had silently become no-ops, because they
substituted a form the config no longer carries. Their own guard caught that rather than the
assertion, so they are now anchored to whichever permitted form is actually shipped.

### Self-review found one defect, which is fixed here

Returning as soon as a permitted entry was found would have accepted a **multi-entry** allow-list.
cosign fails closed on those, so such a configuration verifies *nothing* — the check would have
reported "contract passed" about a cluster where a DR artifact is unusable. That is now refused,
with a positive control proving the same entry alone still passes.

### Scope

No production manifest changes. `ksail.prod.yaml` and the FluxInstance are untouched; only the
validator and its tests change, so this cannot alter what the cluster trusts — it changes what CI
will *accept* as a correct declaration.

Issue #2990, opened 2026-08-06
Triage trivy's 29 HIGH/CRITICAL findings, starting with the secret-carrying configuration class


devantler:
> 🤖 Generated by the Agentic Engineer

**Evidence:** with the KSV-0039 artifact suppressed (#2988), trivy's residual on this tree is **164
findings**, and the severity split makes the priority obvious: **CRITICAL 15 · HIGH 14 · MEDIUM 104
· LOW 31**. Those **29 HIGH/CRITICAL** were previously indistinguishable inside an undifferentiated
count of 614, which is why they have never been triaged as a group.

By control class:

| severity | class | n |
|---|---|---|
| CRITICAL | broad resource management in cluster RBAC | 10 |
| CRITICAL | secret-management permissions in RBAC | 4 |
| CRITICAL | admission-webhook management permissions | 5 |
| HIGH | cluster networking management permissions | 8 |
| HIGH | container root filesystem not read-only | 3 |
| HIGH | configuration objects flagged as carrying secret material | 2 |
| HIGH | pod exec permissions | 1 |

> This repository is public, so the per-file inventory is deliberately **not** reproduced here — it
> is a weakness map. The class-and-count posture above is what is needed to review the fix; the
> site-level detail is held in the private operator notes and can be re-derived at any time with
> `scripts/megalinter-scan-counts.sh` plus a severity filter.

**Audience/problem:** these are the findings that would actually matter if one were real, and the
scanner cannot distinguish "an operator legitimately needs this permission" from "we granted more
than we meant to". Only a per-class verdict can.

**Shape of the work — measured, so it is smaller than 29 suggests:**

1. **A clear majority sit in vendored upstream operator manifests** — third-party RBAC we do not
   author. #2899 already established the precedent for dispositioning vendored operator findings as
   a class for checkov, and the same reasoning should be applied once here rather than per finding.
2. **A small group is deliberate first-party tenant RBAC**, where the permission is the intended
   design and wants a scoped, justified skip naming that reason.
3. **The remainder are scattered singles** needing individual verdicts — fix at the manifest where
   the grant is genuinely wider than intended.

**Do the configuration-objects-carrying-secrets class first.** It is only 2 findings and it is the
one class here that could indicate real secret material rather than broad-but-intended RBAC, so it
is both the cheapest and the highest-consequence thing to settle. Everything else is a permissions
judgement; that one is a fact to check.

**Acceptance criteria:**
[...]

PR #3016, opened 2026-08-08
fix(trivy): scope ConfigMap secret false positives

devantler:

## Why

Trivy reports two HIGH KSV-0109 findings as ConfigMaps containing secrets. Inspection shows that neither contains credential material: one embeds public JavaScript that reads values from an ESO-mounted Secret, and the other contains a public `external_secrets_replicas` replica count. The scanner is matching identifier text.

## What

- migrate the existing Trivy disposition to the path-capable YAML ignore format
- add separate KSV-0109 entries scoped to only the two reviewed ConfigMaps, each with its checked reason
- pass the ignore file explicitly through both MegaLinter and the local CI-equivalent scan helper
- add a test-first command-boundary guard so the helper cannot silently stop loading the scoped file

## Verification

- RED: the new helper contract test failed because `--ignorefile .trivyignore.yaml` was absent
- GREEN: `bash scripts/tests/test-megalinter-scan-counts-ignorefile.sh`
- `shellcheck` and `bash -n` on both scan scripts
- `actionlint .github/workflows/ci.yaml`
- YAML structural validation with `yq`
- real Trivy reproduction: 165 → 163 findings; KSV-0109 2 → 0
- ablation with no ignore file: exactly 2 KSV-0109 findings return
- 26 other local HIGH/CRITICAL findings remain reported

Part of #2990.
Part of #2787.


devantler:


@coderabbitai review

devantler:

## Readiness evaluation

**Reviewed head:** `d1338d89174db0d42ec81d87b33ecd841e1411b7`

I exercised the behavior as the scan consumer, not only the configuration text:

- the repository helper invoked real Trivy with the scoped YAML ignore file and reported 163 residual findings instead of 165
- KSV-0109 reported 0 at the two reviewed paths
- the no-ignore ablation restored exactly 2 KSV-0109 findings
- unrelated HIGH/CRITICAL findings remained visible
- all 20 GitHub checks passed, the branch is mergeable, review threads are empty, and CodeRabbit completed a current-head review with no actionable comments

Judgment: the two false positives are removed narrowly without weakening KSV-0109 elsewhere, and the scan remains useful to the next #2990 slice.

PR #3027, opened 2026-08-09
security(k8s): record the CNPG degraded-alert service-account-token verdict

### LINKED ISSUE #2845: security(k8s): give each workload an explicit service-account-token verdict (CKV_K8S_38)

devantler:

## Why

`CKV_K8S_38` flags workloads that mount a service-account token they may not need — a credential
sitting in the container filesystem. One workload was still failing that check: the CronJob that
alerts on a degraded CloudNativePG database.

It genuinely needs its token — it calls the Kubernetes API to list CNPG clusters — and someone had
already written that reasoning into the manifest as a comment. But a scanner cannot read a comment,
so the finding stayed open and kept the checkov gate off.

## What

Records the same verdict in the form checkov actually reads — the resource-scoped
`checkov.io/skip1` annotation that the other seven `CKV_K8S_38` workloads in this repo already use.
No behaviour change: the annotation is inert to Kubernetes, and the token itself is untouched.

`CKV_K8S_38` now reports **zero** across the repository, taking the tracked checkov total from
**12 to 11**.

## Worth knowing at review time

**This issue's evidence was stale and the scope turned out much smaller than written.** #2845 listed
**8** failing workloads measured on 2026-07-28; re-measuring on current `main` found **1**. Seven were
resolved in the intervening two weeks. The one remaining was not even on the original list — it is a
newer workload that arrived after the issue was filed.

**One acceptance criterion was not met as written, and could not be.** Criterion 4 says no linter
suppression may be added. Every one of the eight workloads legitimately needs its token, so there is
nothing to fix at the manifest — an annotation is the only way to record the verdict. That is the
route the parent #2787 explicitly permits ("anything genuinely irreducible gets a scoped, justified
skip with its reason, never a blanket disable"), and it is what the previous seven dispositions
already did. Flagging it rather than quietly reinterpreting the criterion.

Fixes #2845
Part of #2787

### DIFF
--- a/k8s/bases/infrastructure/controllers/coroot/cron-job-cnpg-degraded-alert.yaml
+++ b/k8s/bases/infrastructure/controllers/coroot/cron-job-cnpg-degraded-alert.yaml
@@ -58,6 +58,8 @@ kind: CronJob
 metadata:
   name: cnpg-degraded-alert
   namespace: observability
+  annotations:
+    checkov.io/skip1: CKV_K8S_38=the check container reads its SA token to list CNPG Clusters through the API server
 spec:
   schedule: "*/30 * * * *"
   concurrencyPolicy: Forbid


devantler:

## The red CI here is not this change

`🧪 Validate Manifests` fails on the step `📣 Validate the rollout gate cannot suppress sync silently`,
which this PR does not touch — it only adds a `checkov.io/skip1` annotation to one CronJob.

The cause is repository-wide and started at 00:00Z today: a test fixture in
`scripts/tests/test-cilium-rollout-gate-suppression-signal.sh` hard-codes `2026-07-26` as a
suppression-marker date and asserts the reporter exits 0. That date reached the reporter's 14-day
hard-fail bound today, so the reporter correctly fails and the assertion fails with it. Every platform
PR opened from now on hits this — the last green PR run was 2026-08-08T21:36Z.

Fix is in **#3029**. This PR is blocked behind it and will be rebased once that merges; nothing to fix
here in the meantime.

Separately, the *real* activation marker carries the same date and the gate is active, so prod deploys
are affected too — that is **#3028**, and it needs a maintainer decision.

devantler:

## Parked on a named blocker — the red CI here is not this PR's fault

Diagnosed rather than re-run. `Validate Manifests` fails on
`FAIL: a single marker read from PLATFORM_ROOT must succeed`, and `CI - Required Checks` is only the
aggregator reporting it downstream. That assertion is a control in
`scripts/tests/test-cilium-rollout-gate-suppression-signal.sh` that still carried an absolute date
while the rest of the file uses relative ones; it aged past its bound during 2026-08-09 and now fails
on every pull request in this repository. Nothing in this PR's own diff is involved.

The fix is **#3029**, which is green and ready but cannot merge yet — merging here runs the deploy
path, and that path is blocked by **#3028**. So this PR stays parked until #3028 is decided and #3029
lands; re-running its checks before then would just reproduce the same failure.

Order: #3028 → #3029 → this.

devantler:


@coderabbitai review

devantler:


CodeRabbit returned `Review rate limited` at this head, so there is no code finding to answer — advancing to the next lane rather than re-spending the refused one.

Worth recording, because it contradicts the tempting inference: the refusal is **not** account-wide. In the same minute, `platform-tenant-template#132` — same account, same credentials — returned `Review in progress`. So a refusal observed here says nothing about the lane elsewhere, and must not be latched across PRs.


@codex review

devantler:

**Status: recovered to green, and deliberately not enqueued.**

The red here was never a finding in this PR. It was #3029's expired control fixture — a run from `00:22Z`, when the assertion still carried an absolute `2026-07-26` that aged past its 14-day bound. #3029 merged at `14:41Z`, so `update-branch` alone took this to **CLEAN, 0 failing checks** at `a5c888a89f`. Nothing in the diff changed.

**Merge is parked on a named, live-verified blocker — not on judgement.** Reading `cd.yaml`'s own header: *"Production deploys happen automatically in the CI merge queue (ci.yaml deploy-prod on merge_group)."* So enqueuing this PR **is** a production deploy, and prod currently refuses one: the Cilium rollout gate has reached its 14-day bound and hard-fails `deploy-prod` (4 dispatch failures today, most recently run 31321097931).

That is verified against the queue rather than inferred — `pr-2985`'s `merge_group` CI **failed at 16:28Z today**, so the queue is evicting right now. Enqueuing this would take the same eviction and add noise to an in-flight remediation.

The unblocker is #3035, which releases the gate. This PR waits behind it; the ordering is #3028 → #3035 → this. I am not touching either — both are being driven interactively.

Review lane at this head: CodeRabbit returned `Review rate limited`, Codex requested.

devantler:

**Readiness record at `a5c888a89f` — and a caveat CI would not have shown me.**

*The green CI does not evidence this change.* `🛡️ Vulnerability Scan` is **SKIPPED** on this PR, and that is the only job that consumes a checkov annotation. So "33 checks, 0 failing" says nothing about whether the skip actually works — a skipped job counts as fine in that tally.

*So I ran the real tool, both states.* checkov 3.3.0 against the manifest at this head versus at `main`:

| | CKV_K8S_38 |
|---|---|
| `main` (no annotation) | **Failed 1**, Skipped 0 |
| this head | Failed 0, **Skipped 1** |

The annotation is consumed as intended, verified against the tool rather than inferred from the syntax.

*One thing the run surfaced that is out of scope here.* The same CronJob also fails **CKV_K8S_35** ("prefer secrets as files over secrets as environment variables"), on `main` and at this head alike — pre-existing, untouched by this PR, and not the verdict this PR records. Flagging rather than folding it in, since mixing a second finding into a one-line verdict record is how these stop being reviewable. It needs its own fix-or-except decision.

*Review:* Codex clean pass at `a5c888a89f` (only Codex artifact on the PR — no `## Review finding` sibling). CodeRabbit was rate-limited at this head.

All three readiness conditions hold, so promoting. **Merge stays parked** on the blocker in my previous comment — #3035 must land first. Enqueuing this before then just takes the same `merge_group` eviction `pr-2985` took at 16:28Z.

Issue #3028, opened 2026-08-09
The Cilium rollout gate has reached its 14-day hard-fail bound and is blocking prod deploys


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

The Cilium rollout gate reached its **14-day hard-fail bound today** (2026-08-09).

- The activation marker in `k8s/providers/hetzner/infrastructure/controllers/kustomization.yaml:91`
  reads `2026-07-26`, and the component at line 92 is active.
- `fail_after_days=14` in `scripts/report-cilium-rollout-gate-suppression.sh:41`. 2026-07-26 → today
  is exactly 14 days, so the reporter now emits `::error::` and exits non-zero.
- Live cluster state confirms the rollout is genuinely stalled, not merely un-acknowledged:

  ```
  kubectl -n kube-system get ds cilium
  updateStrategy: OnDelete
  updatedNumberScheduled: 0 / 9      # no agent is on the current revision
  numberReady:            9 / 9
  ```

  `OnDelete` means the agents will never step on their own, so this cannot resolve without action.

## Impact

Talos machine-config sync (`ksail cluster update`) has been suppressed for two weeks — machine config
in Git is diverging from the nodes — and the deploy will now fail rather than skip it silently. That
is the gate working as designed; it has simply run out of patience.

## What this issue is NOT

CI breakage for every pull request had the **same root date** but a **separate cause**: a test fixture
hard-coded `2026-07-26` and asserted success. That is fixed separately in #3029 and is not a
resolution of this issue.

## Resolution — a maintainer decision, deliberately not attempted here

Per `AGENTS.md`, the only two resolutions are:

1. **Step the remaining Cilium agents onto the current DaemonSet revision** — a fleet-wide,
   one-node-at-a-time operation on prod, or
2. **Roll the `homogeneous-devices` component back.**

**Raising either bound is explicitly not a resolution**, so I have not touched them.

⚠️ **Option 2 is no longer the cheap option it once was.** The running agents and the desired
manifest have diverged during the freeze, so a rollback would now perform a fleet-wide CNI change
rather than a no-op. Both paths are real production CNI operations with genuine blast radius, which
is why this is being surfaced rather than actioned autonomously — my cluster access is read-only by
design, and this is not a change to make without a human deciding which path to take.

## Acceptance criteria

- [ ] The Cilium agents are on the current DaemonSet revision, or the component is rolled back.
- [ ] `updatedNumberScheduled` equals `desiredNumberScheduled`.
[...]

PR #3029, opened 2026-08-09
fix(ci): date the gate-suppression control fixture relative to today

devantler:

## Why

**Every pull request in this repository started failing CI a few hours ago**, and nothing was
committed to cause it — the failure arrives purely with the calendar.

A test fixture for the Cilium rollout-gate reporter hard-coded `2026-07-26` as the date of a
suppression marker, then asserted the reporter exits cleanly for it. Today is exactly 14 days later,
which is the reporter's hard-fail bound. So the reporter did the right thing and failed, the
assertion that expected success failed with it, and `🧪 Validate Manifests` went red for everyone.

The last green PR run was 21:36Z yesterday; the first failure was the first run after midnight UTC.

## What

Dates that fixture relative to today, which is what every other date assertion in the same file
already does. The duplicate-marker case above it keeps its absolute dates on purpose — it asserts
*failure*, and fails closed on the duplication before age is ever evaluated.

Verified the control still has teeth rather than passing vacuously: feeding it an out-of-bound date
again makes it fail at exactly that assertion. Full suite 30 passed / 0 failed.

## What this does NOT fix — please read

This unblocks CI. It does **not** resolve the underlying gate, and deliberately does not try to.

**The real activation marker carries the same `2026-07-26` date and the gate is active**, so the live
Cilium rollout gate has also reached its 14-day bound. Talos machine-config sync has now been
suppressed for two weeks, and prod deploys will fail on it. That needs a decision only you can make —
stepping the remaining Cilium agents onto the current DaemonSet revision, or rolling the component
back — and it is not something I can safely do from here. Raising the bound is explicitly not a
resolution, so I have not touched it. Filed separately as #3028.

Part of #3028 (this is the CI half; the gate itself is that issue).

### DIFF
--- a/scripts/tests/test-cilium-rollout-gate-suppression-signal.sh
+++ b/scripts/tests/test-cilium-rollout-gate-suppression-signal.sh
@@ -272,12 +272,18 @@ ok
 
 # Control for the fixture: with exactly one marker the same path succeeds, so the
 # assertion above is about duplication and not about the fixture being unreadable.
-printf '  # %s 2026-07-26\n' "${activation_marker}" >"${fixture_kustomization}"
+#
+# The date is RELATIVE, like every other date in this file. An absolute one was a
+# time bomb: this control asserts exit 0, so once the hard-coded date aged past
+# fail_after_days the reporter correctly failed and took every pull request in the
+# repository with it. The duplicate case above can keep its absolute dates because
+# it asserts FAILURE — it fails closed on the duplication before age matters.
+printf '  # %s %s\n' "${activation_marker}" "${inside}" >"${fixture_kustomization}"
 : >"${tmp}/summary"
 CILIUM_ROLLOUT_GATE_ACTIVE=true PLATFORM_ROOT="${fixture_root}" \
   GITHUB_STEP_SUMMARY="${tmp}/summary" "${report_script}" >"${tmp}/out" 2>&1 ||
   fail 'a single marker read from PLATFORM_ROOT must succeed'
-grep -Fq '2026-07-26' "${tmp}/summary" ||
+grep -Fq "${inside}" "${tmp}/summary" ||
   fail 'the fixture summary must carry the declared date'
 ok
 


devantler:

## Scope check — this was the only instance of the class

Swept `scripts/tests/` for hard-coded `YYYY-MM-DD` values to see whether other tests carry the same
time bomb. Two files matched; only one was a bomb:

| file | verdict |
|---|---|
| `test-cilium-rollout-gate-suppression-signal.sh` | **the bug** — asserted success against a date compared to *now* |
| `test-dr-rebuild-supersession-guard.sh` | **safe** — its dates are compared only to *each other* (run ordering), never to the clock. The guard contains no `date`/`now`/`%s` in any code path, and its 23 checks pass today |

The remaining two absolute dates in the cilium test are the duplicate-marker case, which asserts
**failure** and fails closed on the duplication before age is evaluated — those are correct as-is.

**The distinguishing property, for future reference:** a hard-coded date is only a bomb when the
assertion expects **success** and the code under test compares that date against **now**. Either half
alone is harmless.

## Verification

- Full suite: **30 passed / 0 failed**.
- Ablation: substituting an out-of-bound date back in makes it fail at exactly
  `a single marker read from PLATFORM_ROOT must succeed` — so the control still has teeth and is not
  passing vacuously.

devantler:

@coderabbitai review

devantler:

## Readiness record

**Programmatically tested.** Local suite 30 passed / 0 failed. Ablation (out-of-bound date substituted
back in) fails at exactly `a single marker read from PLATFORM_ROOT must succeed`, so the control still
has teeth rather than passing vacuously.

**Tried and evaluated as a user — the real behaviour, not inference.** The step that is red on every
other platform PR right now is green here, in CI, on this head:

```
run 31286408197 → job 93176000737 (🧪 Validate Manifests)
📣 Validate the rollout gate cannot suppress sync silently => conclusion=success
```

For contrast, the same step on #3027 (`run 31286015707`):
`##[error]Process completed with exit code 1` — `FAIL: a single marker read from PLATFORM_ROOT must succeed`.

**Checks:** 18 SUCCESS / 12 SKIPPED / 0 failing, `mergeStateStatus: CLEAN`.

I also name-set-diffed this run's checks against #3027's to rule out a required check going missing
(a green rollup only aggregates what ran). The only difference is `kubescape`, which is present and
SUCCESS **here** and absent on #3027 — its early failure cancelled the downstream job. This run is the
superset, so nothing is silently missing.

Awaiting a review before promoting.

devantler:

## Promoted — and deliberately NOT enqueued

Readiness is proven and the pentad is clear: checks 18 SUCCESS / 12 SKIPPED / 0 failing, 0 unresolved
threads, 0 actionable review findings, `mergeStateStatus: CLEAN`, and a green
**`cr@843b841d`** (CodeRabbit summary refreshed at 00:41:30Z, after the 00:40:23Z request, naming the
full head, no service marker; its only section is `🔇 Additional comments (1)`, which is informational).

**I am not putting it in the merge queue, and that is the considered call rather than an omission.**

`merge_group` runs `deploy-prod`, which runs the gate reporter under
`always() && cilium_rollout_gate.outputs.active == "true"` — deliberately written so an earlier failure
cannot skip it. The gate is active and 14 days old, so that step fails, the deploy fails, and the PR is
evicted. Enqueuing would not merge this; it would attempt a **real production deploy** against a cluster
whose machine config has been diverging for two weeks — the exact thing the gate exists to prevent — and
would leave a speculative artifact for the heal job on the way out.

The contract is explicit that a known root cause is landed first rather than looped through the queue,
and the root cause is **#3028**. So this sits ready, and merges the moment the rollout is resolved.

There is a one-line override that would make the queue pass. I am not using it: the gate is reporting a
real divergence, and silencing it because it is blocking us is the failure mode it was built after.

**Net position:** the pull-request half of the outage is fixed and waiting; the cluster half is #3028
and needs a decision only you can make.

devantler:

## Correction: this PR has no head-attributable green review right now

I promoted this on a CodeRabbit green recorded as "the auto-generated summary comment named the full
head". Re-checking at `843b841d97`, that is **not** what the surface shows:

- **0** CodeRabbit review objects on the PR at any head.
- The auto-generated summary comment (updated 00:43:41Z, after the request) carries a real walkthrough
  of this exact change and no service marker — but it names **no** commit sha, full or short, and has
  no `Review details`/`Commits` section. So it cannot be attributed to a head.
- The head's `CodeRabbit` status did read `success — Review completed` at 00:41:32Z, which evidences
  that a run happened. Per the contract that status is a *corroborator, never a satisfier* — it needs
  the real artifact alongside it, and there isn't one.

So the honest state is `green_review=none@843b841d97`. The review very likely did run cleanly; what
is missing is the head attribution the gate requires, and I am not going to treat "probably fine" as
a green on a leak-adjacent gate.

**No action is being taken on it this run, deliberately.** The repository cannot merge anything while
platform#3028 is open — `merge_group` runs `deploy-prod`, the Cilium rollout gate fails there, and the
PR is evicted — so spending a CodeRabbit request now would only re-stale before it could ever be used.
The review gets re-secured when #3028 clears and this becomes mergeable.

This is also a second live instance of #2729 (the summary comment is mutable and can lose its head
attribution), recorded there.

devantler:

@codex review




CodeRabbit ran at this head (`Review completed`, 00:41:32Z, after the 00:40:23Z request) but produced **no gate-satisfying artifact**: zero review objects, and its auto-generated summary comment carries no commit sha at all, so nothing is attributable to `843b841d97`. That is a lane completion without a green, not a rate limit — advancing to Codex per lane order rather than re-spending CodeRabbit.

devantler:

## Green review secured at head — still deliberately not enqueued

Codex returned `Didn't find any major issues` at `843b841d97`, which is this PR's current head, so the review gate is now satisfied. Verified alongside it: zero connector `## Review finding` comments (Codex publishes those in comment form, so a green does not by itself mean no finding), zero unresolved threads, checks green, `mergeStateStatus: CLEAN`.

Lane record for this head: `cr:no-gate@843b841d97` (CodeRabbit ran — status `Review completed` — but produced no artifact attributable to this head: no review object, and its summary comment carries no commit sha at all), then Codex green.

**The remaining blocker is unchanged and is not a review problem.** This repo merges through a queue, so merging enqueues and runs `deploy-prod` against production — and the expired Cilium rollout gate hard-fails there on config that has been diverging for two weeks. That is #3028, which needs a real production CNI decision from the maintainer. Ordering stays **#3028 → #3029 → #3027**.

So this PR is mergeable on its own merits and is held only by that sequencing. No further review requests are needed for it; the next run should re-check #3028 rather than re-reviewing here.

PR #3173, opened 2026-08-16
feat(homepage): declare Infrastructure as Code and enforce declared groups

devantler:

### Why

On the dashboard, Crossview and Simba Countdown render *below* the bookmark rail instead of with the
other services. The cause is the same for both: Homepage takes a group's position and icon from the
`layout` block in the homepage config, but the group **name** comes from an annotation on a route —
frequently in a different repository. Nothing bound the two, and the failure is silent: an undeclared
group still renders, it just sinks below every declared group with a generic icon.

Two groups were in exactly that state — `Infrastructure as Code` (a platform-owned route, in this
repo) and `Fun` (a tenant route). Neither was declared.

### What

Declares **Infrastructure as Code**, positioned ahead of Cloud and Network — Crossplane declares the
Hetzner cloud and the UniFi network, so it reads before the provider consoles it manages. That alone
moves Crossview up with the other services.

Adds a Kyverno policy, **restrict-homepage-service-groups**, so this cannot silently recur: a route
that opts into dashboard discovery may only name a group the layout declares. Homepage has no native
way to filter discovery by group, so admission is the only place the two surfaces can be bound. It
applies to the platform's own routes as well as tenants' — the Crossview route is how this was found
— and does not accept bookmark group names, which would paint a bookmark heading twice.

### Merge order

**Land `devantler-tech/doggy-countdown#15` and let it deploy before merging this.** That tenant's
route is still annotated with the undeclared group in-cluster, because its repo has no release
pipeline and a fix from 2026-08-04 has never been published. Flux re-applies on every reconcile, so
enforcing here first would deny that route at admission and break the tenant's reconciliation. The
repo is already correct; it just needs a release.

### Trade-off worth flagging

The policy's allow-list duplicates the layout's service groups rather than reading the ConfigMap.
A Kyverno context lookup would bind them perfectly but makes every annotated route in the cluster
un-appliable if that ConfigMap ever became unreadable — trading a silent cosmetic bug for a
reconciliation outage. The two files now cross-reference each other so the pair moves in one PR.

### DIFF
--- a/k8s/bases/infrastructure/cluster-policies/best-practices/restrict-homepage-service-groups.yaml
+++ b/k8s/bases/infrastructure/cluster-policies/best-practices/restrict-homepage-service-groups.yaml
@@ -45,14 +45,19 @@ metadata:
     policies.kyverno.io/category: Best Practices
     policies.kyverno.io/severity: low
     policies.kyverno.io/subject: HTTPRoute, Ingress
-    policies.kyverno.io/minversion: 1.9.0
+    policies.kyverno.io/minversion: 1.13.0
     policies.kyverno.io/description: >-
       Requires every route that opts into Homepage discovery
       (gethomepage.dev/enabled: "true") to name a service group declared in the
       dashboard's layout, so a discovered service cannot invent a group and sink
       to the bottom of the page with a generic icon.
 spec:
-  validationFailureAction: Enforce
+  # Rule-level `validate.failureAction` below, not the spec-level
+  # `validationFailureAction`: Kyverno's own v1 CRD describes that one as
+  # "Deprecated, use validationFailureAction under the validate rule instead."
+  # The eleven older policies here still use the deprecated form — it is honoured
+  # on the deployed v1.18.2 — and migrating them is tracked separately; a new
+  # policy should not add to that debt.
   rules:
     - name: declared-service-group
       match:
@@ -72,6 +77,7 @@ spec:
             operator: Equals
             value: "true"
       validate:
+        failureAction: Enforce
         message: >-
           A route with gethomepage.dev/enabled="true" must set
           gethomepage.dev/group to a service group declared in the Homepage
--- a/k8s/bases/infrastructure/cluster-policies/best-practices/restrict-homepage-service-groups.yaml
+++ b/k8s/bases/infrastructure/cluster-policies/best-practices/restrict-homepage-service-groups.yaml
@@ -1,0 +1,104 @@
+# Keeps every Homepage-discovered service inside a group the dashboard actually
+# declares.
+#
+# Why: Homepage's group ORDER, icon and position come from the `layout` block in
+# k8s/bases/apps/homepage/config-map.yaml, but the group NAME is chosen by an
+# annotation on a route — often in another repository entirely. Nothing bound the
+# two surfaces, so a route could invent a group, and Homepage would silently
+# accept it: an undeclared group still renders, but falls to the BOTTOM of the
+# page (after every declared group, bookmarks included) with a generic icon.
+# The failure is cosmetic-looking and completely silent, which is why it went
+# unnoticed in two places at once — `Fun` and `Infrastructure as Code` both
+# rendered below the bookmark rail because neither was declared.
+#
+# Homepage has no native way to restrict discovery to declared groups (its
+# Kubernetes config offers only `gethomepage.dev/instance` filtering), so the
+# binding has to happen at admission. This is that binding: a route may only
+# name a group the dashboard declares.
+#
+# Scope: every namespace, and deliberately NOT limited to tenant-applied
+# resources — the platform's own routes are just as able to invent a group, and
+# `Infrastructure as Code` (a platform route, in this repo) is exactly how this
+# was found. There is no userInfo matching here, so background scanning stays on
+# and existing drift surfaces in the PolicyReports rather than only at admission.
+#
+# SERVICE groups only. The bookmark groups in the same layout block (GitHub,
+# Media, Banking, …) are intentionally absent: Homepage renders services and
+# bookmarks as separate sections, so the config-map keeps the two name sets
+# distinct, and a service claiming a bookmark group's name would paint that
+# heading twice with different contents under each.
+#
+# KEEP IN SYNC with the `layout` block in k8s/bases/apps/homepage/config-map.yaml
+# — adding a group there means adding it here, in the same PR. That is a
+# reviewable, fail-loud edit: a route naming an undeclared group is rejected at
+# admission with the list below, instead of silently sinking to the bottom of the
+# dashboard. The list is duplicated rather than read from the ConfigMap on
+# purpose: a Kyverno context lookup would make every annotated route in the
+# cluster un-appliable if that ConfigMap ever became unreadable, trading a silent
+# cosmetic bug for a reconciliation outage.
+apiVersion: kyverno.io/v1
+kind: ClusterPolicy
+metadata:
+  name: restrict-homepage-service-groups
+  annotations:
+    policies.kyverno.io/title: Restrict Homepage Service Groups
+    policies.kyverno.io/category: Best Practices
+    policies.kyverno.io/severity: low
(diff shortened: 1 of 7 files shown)


k8s/bases/infrastructure/cluster-policies/best-practices/restrict-homepage-service-groups.yaml
```diff
@@ -0,0 +1,104 @@
...
+    policies.kyverno.io/severity: low
+    policies.kyverno.io/subject: HTTPRoute, Ingress
+    policies.kyverno.io/minversion: 1.9.0
+    policies.kyverno.io/description: >-
+      Requires every route that opts into Homepage discovery
+      (gethomepage.dev/enabled: "true") to name a service group declared in the
+      dashboard's layout, so a discovered service cannot invent a group and sink
+      to the bottom of the page with a generic icon.
+spec:
+  validationFailureAction: Enforce
```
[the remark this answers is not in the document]
devantler: > 🤖 Generated by the Agentic Engineer

**Adopted for this policy, and the guard you asked for is added. The other eleven are a separate,
now-filed migration.**

**Verified before changing anything** — this is deprecation debt, not a functional break. Kyverno's
own v1 CRD on the deployed **v1.18.2** says of the spec-level field: *"Deprecated, use
validationFailureAction under the validate rule instead."* The live
`clusterpolicy/restrict-tenant-route-hostnames` still carries `spec.validationFailureAction: Enforce`
and reports `Ready: True` with 4 validate rules, so the old form is honoured. I confirmed
`validate.failureAction` exists in the same CRD with enum `["Audit","Enforce"]` before switching.

**Changed here:** `validate.failureAction: Enforce`, spec-level field removed, `minversion` → `1.13.0`.
A new file should not add to the debt even while its neighbours carry it.

**Not changed here:** the other **11** policies (0 currently use the modern form). Migrating them is
mechanical but touches every policy at once, so it is **#3176** rather than a rider on a feature PR.

### The static guard

`scripts/tests/test-restrict-homepage-service-groups.sh`, wired into the `⚖️ Validate cluster
policies` job. It goes further than the effective action, because the allow-list duplication was the
real risk:

1. the rule sets `validate.failureAction: Enforce`, and the deprecated spec field is **absent**;
2. **the allow-list exactly equals the layout's service groups** — derived *structurally* as
   `layout keys − bookmark group names`, not by reading the section comment, so rewording that
   comment cannot quietly change what the test believes;
3. no bookmark group name is accepted as a service group.

It fails closed: extracting zero groups on any side is an error, not a vacuous pass.

**Ablations, each verified to actually land before running:**

| Ablation | Result |
|---|---|
| drop `Analytics` from the allow-list | fails, diff shows `< Analytics` |
| `failureAction: Enforce` → `Audit` | fails on the action assertion |
| add bookmark group `Banking` to the allow-list | fails, diff shows `> Banking` |

`shellcheck` clean; `kyverno test ./tests` **49/49**; `ksail workload validate` green on both overlays.

This turns the "keep in sync" comment into a tested invariant — which is a better answer to the
duplication trade-off than the comment was.

k8s/bases/infrastructure/cluster-policies/best-practices/restrict-homepage-service-groups.yaml
```diff
@@ -0,0 +1,110 @@
...
+    - name: declared-service-group
+      match:
+        any:
+          - resources:
+              kinds:
+                # HTTPRoute is what this platform actually uses. Ingress is
+                # Homepage's other discovery source and is included so a future
+                # Ingress cannot bypass the guard.
+                - HTTPRoute
+                - Ingress
```
devantler, line 71: nit: the match block sets no `operations:`, so the rule is evaluated for DELETE as well as CREATE/UPDATE. That is safe as written — on DELETE `request.object` is null, so the precondition's `|| 'false'` default makes the rule skip — but the safety is implicit, resting on a JMESPath default rather than on the match narrowing the operations it cares about.

Adding `operations: [CREATE, UPDATE]` would state the intent directly. Not doing it in this PR: it would restale a review gate that all three external lanes currently cannot regenerate (CodeRabbit rate limited, Codex and Bugbot at account usage limits), which is a poor trade for a clarity change with no behavioural difference. Worth folding into the next touch of this file.

devantler (commented):

**Reviewed commit:** `9d3e72cbdb34bb14d3e9c06e8b7b963b6de6b2c4`

Per-lane evidence, stated precisely rather than rounded up:

- **CodeRabbit** — `Review rate limited` on this head's commit status. Two attempts: the standard
  trigger was answered with the "initiate chat on the files" tip (a known misparse), and
  `@coderabbitai full review` — the documented escape from that wedge — returned the same rate limit.
  `cr:no-gate@9d3e72cbdb`
- **Codex** — `You have reached your Codex usage limits for code reviews`, observed at 19:22:03Z on
  the immediately preceding head. `codex:no-gate@…`
- **Cursor Bugbot** — check-run `neutral` + `output.title: Error`, with the `cursor[bot]` comment
  giving the cause: `Bugbot couldn't run - usage limit reached` (spend limit, no retry window,
  maintainer-liftable), 19:22:28Z on the preceding head.

⚠️ Codex's and Bugbot's evidence is from the preceding head, not this one. Both are **account-level
usage/spend limits with no stated retry window**, which a new commit twelve minutes later cannot
reset, so I did not re-spend requests to re-observe them. Only CodeRabbit's refusal is measured at
this exact head.

### What I checked

**The policy does what the PR claims, in the built output — not just in the file.** Rendered
`k8s/providers/hetzner/infrastructure` (10027 lines) and asserted the policy is present with
`failureAction: Enforce` and 13 allowed groups, so the kustomization registration is real rather than
assumed.

**Operational safety of a cluster-wide Enforce policy.** It carries no `request.userInfo` reference,
so background scanning (default on) works and existing drift surfaces in PolicyReports instead of
only at admission. On DELETE, `request.object` is null, so the precondition's `|| 'false'` default
makes the rule skip rather than evaluate against a missing object.

**That it cannot deny anything currently deployed.** Every live route's group is now declared —
verified against the cluster after `doggy-countdown` v1.0.2 landed. This was the one real
merge-ordering hazard and it is resolved, not assumed.

**Drift, which is the whole point of the change.** The group list exists in three places: the layout,
the policy allow-list, and the deny message. `scripts/tests/test-restrict-homepage-service-groups.sh`
pins all three, deriving the expected set *structurally* (layout keys − bookmark groups) rather than
by parsing a section comment. Four ablations, each verified to land before running: drop a group
(fails), `Enforce`→`Audit` (fails), leak bookmark group `Banking` (fails), stale the message (fails).
It also fails closed if any side extracts zero groups.

**Validation.** CI 20 success / 11 skipped / 0 failures; `kyverno test ./tests` 49/49;
`ksail workload validate` 576 files exit 0 on both overlays; `shellcheck` clean; naming validator clean.

### What I could NOT verify

Real admission behaviour in-cluster. `kubectl apply --dry-run=server` is blocked in this environment,
so the Kyverno CLI evaluation plus the rendered-output assertion are the proxies. The behaviour to
confirm after rollout is that a route naming an undeclared group is actually rejected.

**Verdict: no P0/P1 findings** — one non-blocking nit inline.

k8s/bases/infrastructure/cluster-policies/best-practices/restrict-homepage-service-groups.yaml
devantler: > 🤖 Generated by the Agentic Engineer

Deferring, with the reason recorded rather than dropped.

This is my own non-blocking nit and it changes no behaviour: on DELETE the precondition already
short-circuits via its `|| 'false'` default, so narrowing `operations:` would state the intent more
directly without altering what the policy does.

Acting on it now would move the head and stale a clean current-head review that none of the three
external lanes can currently regenerate — CodeRabbit is rate limited at this head, and Codex and
Bugbot are both at account-level usage limits with no retry window. Trading a satisfied review gate
for a clarity edit is the wrong way round.

Folded into the next substantive touch of this file, which will need a review round anyway.

Issue #3184, opened 2026-08-17
Prod deploy evicted from the merge queue when the autoscaler changes a node mid-transaction


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

The prod deploy's node-safety transaction races the cluster autoscaler and refuses to proceed when
an autoscale node changes underneath it. Three merge-group deploys have now died this way in under
two hours, each evicting its PR from the merge queue:

| When (UTC) | Run | Failure |
|---|---|---|
| 2026-08-16 22:39 | #3171 | node claimed mid-transaction was replaced — `identity changed … refusing image verification` |
| 2026-08-16 23:02 | #3171 | evicted again; orphaned `Lease flux-system/ghcr-auth-refresh` held past expiry |
| 2026-08-17 00:16 | #3171 (run `31980786395`) | `Talos node autoscale-cx43-2834820e83860317 identity changed, cordon ownership changed, or scheduling safety state changed before revision marker; refusing the mutation.` |

**The guard itself is correct** — refusing to mutate a node whose safety state moved is the right
call. The defect is that nothing reconciles the transaction with the autoscaler, so an ordinary
scale event is indistinguishable from a genuine safety violation.

**Residue outlives the run.** `autoscale-cx43-2834820e83860317` is still `cordoned=true` with **no**
`ghcr-auth-drain-owner` annotation — a cordon with no owner, left behind after the refusal. A
cordoned node with no owner is exactly the state the next deploy will trip over again.

**`🩹 Heal Prod` also failed** in this run, which is the free discriminator for whose fault it is:
a base-state fault, not anything in the PR's diff.

## Why it matters

**Every platform merge is blocked while this recurs.** The queue evicts on the failed `merge_group`,
so the PR bounces back to OPEN and the next re-queue hits the same race. Re-queuing without a fix is
futile by the contract's own merge-queue rule.

Prod itself was never at risk: all three deploys died before `publish_platform_manifest`, and prod is
currently **11/11 Kustomizations Ready** — so this is a delivery-lane outage, not a production
incident.

## Expected behaviour

A deploy is not evicted by a routine autoscaler scale event. Either the transaction tolerates an
autoscale node changing state while it holds no claim on that node, or it excludes autoscaler-managed
nodes from the safety marker, or it retries once against the refreshed node set.

Separately, a refused mutation leaves **no cordoned node without a drain owner** — the refusal path
should release its own cordon, the way the node fence self-releases elsewhere.

## Acceptance criteria

[...]

PR #3185, opened 2026-08-17
fix(deploy): ignore the autoscaler's advisory removal-candidate taint

devantler:

### Why

Every platform merge has been blocked: three prod deploys were evicted from the merge queue in
under two hours, each refusing with `scheduling safety state changed`. The deploy's node-safety
transaction was treating an ordinary autoscaler scale event as a safety violation.

The cluster autoscaler continuously marks under-used nodes as *removal candidates* and then
un-marks them, roughly every ten seconds. The deploy holds a node for minutes, so it was almost
guaranteed to see that marker come or go and refuse — even though nothing about the node's actual
safety had changed.

### What

The deploy now ignores that advisory "could be removed" marker when deciding whether a node is
still safe to touch, so routine autoscaling no longer evicts a deploy.

The separate marker that means a node **is** being deleted is deliberately left in place and still
stops the deploy, so nothing is loosened where it matters. A test covering that case is included
alongside the fix and is what proves the distinction holds.

Verified against the real failure: the new test reproduces the exact production error, fails
without the fix, and passes with it.

Refs #3184

### DIFF
--- a/scripts/refresh-flux-ghcr-auth-safety.sh
+++ b/scripts/refresh-flux-ghcr-auth-safety.sh
@@ -139,22 +139,23 @@ node_claim_preconditions_still_hold() {
     --arg uid "${initial_node_uid}" \
     --argjson was_cordoned "${was_cordoned}" \
     --argjson initial_taints "${initial_node_taints}" '
+    def scheduling_taints:
+      map(select((
+        (.key == "node.kubernetes.io/unschedulable"
+          and .effect == "NoSchedule"
+          and (.value // "") == "")
+        or (.key == "DeletionCandidateOfClusterAutoscaler"
+          and .effect == "PreferNoSchedule")
+      ) | not))
+      | sort_by([.key, .effect, (.value // ""), (.timeAdded // "")]);
     .metadata.uid == $uid
     and .metadata.deletionTimestamp == null
     and ((.metadata.annotations[$owner_annotation] // "") == "")
     and ((.metadata.annotations[$recovery_annotation] // "") == "")
     and ((if (.spec.unschedulable // false) then 1 else 0 end)
       == $was_cordoned)
-    and (((.spec.taints // [])
-      | map(select((
-          (.key == "node.kubernetes.io/unschedulable"
-            and .effect == "NoSchedule"
-            and (.value // "") == "")
-          or (.key == "DeletionCandidateOfClusterAutoscaler"
-            and .effect == "PreferNoSchedule")
-        ) | not))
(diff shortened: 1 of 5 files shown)


devantler (commented):

**Reviewed commit:** `3d0e49117933fb3531e6233bd351f0debc00a3d8`

### Per-lane evidence at this exact head

| Lane | Result | Evidence |
|---|---|---|
| CodeRabbit | rate limited | head commit status `success — Review rate limited`, after this round's request marker |
| Codex | usage limit | `chatgpt-codex-connector[bot]` comment 2026-08-17T03:28:50Z — "reached your Codex usage limits for code reviews" |
| Cursor Bugbot | never ran | check-run `Cursor Bugbot` `conclusion=neutral` + `output.title=Error` (started 03:34:03Z), with a `cursor[bot]` comment stating a usage/spend limit was hit |

All three are quota/billing refusals rather than slowness, which is the case the local round exists for.

### Review

The change excludes `DeletionCandidateOfClusterAutoscaler` (`PreferNoSchedule`) from the scheduling-taint comparison while leaving `ToBeDeletedByClusterAutoscaler` (`NoSchedule`) in it. That distinction is the crux and it is correct: the first says a node *could* be removed and churns on the autoscaler's ~10s re-decision loop (its value is a regenerated timestamp, so it differs even key-for-key between two adds); the second says a node *is* being removed and must still fail closed.

The normalization is applied to **both** sides (`$initial_taints | scheduling_taints`), not just the current reading. This matters more than it looks: a recovery journal written before the exclusion existed still carries the taint in its captured `initialTaints`, so a one-sided fix would have repaired the fresh-deploy path and left the recovery path — the one a killed deploy leaves behind — failing with the identical symptom.

All six normalization sites across the two scripts carry the same predicate, and the third function's lifecycle-taint list is updated consistently.

### Verification performed

- `scripts/tests/test-refresh-flux-ghcr-auth-safety.sh` — all pass, including both new arms.
- `go test -run TestAutoscaler ./scripts/tests/refresh-flux-ghcr-auth/` — both pass.
- **Ablation (neutralized the excluded key to a never-matching name, syntax preserved):** `TestAutoscalerDeletionCandidateTaintDoesNotBlockDeploy` **FAILED** and the shell arm went `not ok`, so neither test is vacuous.
- **Specificity control:** under that same ablation, `TestAutoscalerTaintIsNeverUncordoned` and "a released autoscaler deletion taint blocks a reboot" both stayed green — the change is an exclusion of the advisory taint only, not a loosening of the hard one.
- Ablation reverted; worktree clean.

### Verdict

PR #3189, opened 2026-08-17
feat(ci): recover a provably-dead GHCR deploy fence automatically

### LINKED ISSUE #3183: feat(ci): recover a provably-dead GHCR deploy fence automatically

## Evidence

The GHCR sync fence wedged the deploy lane again on 2026-08-16. `merge_group` run
[31976321946](https://github.com/devantler-tech/platform/actions/runs/31976321946) died at 22:42:11Z
holding `flux-system/ghcr-auth-refresh`, and every subsequent attempt failed closed — **including
the heal path**, which is what turns one dead run into a full stop:

```
##[error]Another GHCR synchronization transaction holds the synchronization lease; automatic
expiry takeover is disabled because Talos writes cannot be fenced.
```

Both `🚀 Deploy to Prod` and `🩹 Heal Prod` in run
[31978017291](https://github.com/devantler-tech/platform/actions/runs/31978017291) died on that exact
error. No PR could merge for ~50 minutes until a human ran one `kubectl patch`. #3071 covers *why*
the lease is lost mid-run; #3034 made the state diagnosable and manually recoverable. Neither closes
the gap that **recovery requires a human with cluster write access**.

## The trap this must not fall into

Auto-releasing on **expiry** is exactly what the design forbids, and the script says why:

> an expired shell process could resume after an automatic timeout takeover and write stale
> credentials even if every Kubernetes write uses CAS. Therefore expiry is diagnostic only.

A wall-clock timeout is not proof of death. Any "self-healing" built on `renewTime` reintroduces the
hazard the fence exists to prevent — two processes writing Talos config, which cannot be fenced.

## Why safe automation is nonetheless possible

#3034 already added the missing ingredient: the holder identity embeds the Actions run reference, and
`fence_holder_run_reference()` parses it. **GitHub's API is authoritative about terminality** — a run
reported `completed` cannot resume — so this is a positive death proof, not a timeout inference. The
`--fences` report already instructs a human to make exactly this check; the proposal is to mechanise
that check, not to weaken the invariant.

## Proposal

A pre-flight recovery step in `.github/actions/deploy-prod`, which both the deploy and heal jobs
already share, so one insertion point covers both failing paths. It releases the Lease **only** when
all of these hold, and fails closed on each:

1. The holder identity parses to a run reference. A holder from a local
   `run-ksail-prod-with-pull-auth.sh` invocation has no oracle — never auto-released.
2. That run **and attempt** is `status: completed`. The `--attempt` pin is required: a rerun reuses
   the run id, so an unpinned query reports the latest attempt and misreads a dead attempt-1 orphan
   as live.
3. **No other fence is held.** `--fences` already states the ordering — policy fences, then nodes,
   then the Lease, because "the Lease is the global exclusion fence: clearing it first lets a deploy
   start against a half-recovered cluster."
4. The existing CAS on `resourceVersion` + `holderIdentity` still guards the write.

Ships **default-off behind an opt-in workflow input** per the feature-flag-first convention, tested
in both states, and flipped on only after a real recovery is observed.

## Acceptance criteria

- [ ] Releases the fence when the holder's run+attempt is terminal and no other fence is held
- [ ] Refuses when the run is non-terminal, when the identity carries no run reference, when another
      fence is held, and when the CAS test fails — each with a distinct message
- [ ] RED/GREEN coverage for all five paths in `scripts/tests/refresh-flux-ghcr-auth/`, using the
      existing `fake_kubectl_test.go` harness
- [ ] Both flag states covered
- [ ] Recovery is reported in the job log with the liveness evidence it relied on

Rough size: M — the primitives exist; this is wiring plus tests on a safety-critical path.

## Sequencing note

This cannot deploy itself out of the condition it fixes: merging goes through the merge queue whose
deploy is blocked by the very fence in question. An orphaned fence must be cleared manually **once**
before this can land.

Related: #3071 (root cause), #3034 (diagnosis + manual recovery), #3065 (same class, different fence).

devantler:

## Why

When a prod deploy dies holding the GHCR synchronization Lease, it blocks **every
later deploy and its own heal path** — the whole merge queue stops, and clearing it
needs a human with cluster write access. That happened on 2026-08-16 and cost ~50
minutes of blocked merges.

## What

The deploy composite gains a pre-flight step that releases the Lease by itself, but
**only when the holder is provably dead** — the GitHub API reporting its run attempt
terminal, no other fence held, and the heartbeat stopped. Anything it cannot prove,
it refuses and leaves for a human.

This is deliberately *not* "release it when the lease expires": an expired process
can still wake up and write, which is why automatic expiry takeover is switched off
here in the first place. A completed run cannot wake up, which is what makes this
safe.

**Ships default-off.** The step only runs when the new `recover-orphaned-fence`
input is set to `true`, so this merges latent and gets switched on once we've watched
it decline to act on a healthy lane.

Fixes #3183

### DIFF
--- a/scripts/refresh-flux-ghcr-auth.sh
+++ b/scripts/refresh-flux-ghcr-auth.sh
@@ -409,7 +409,12 @@ fence_report_lease() {
   # A second read could land either side of a fresh acquisition, which would let
   # recovery act on a holder the operator never saw evidence for.
   fence_lease_holder="${holder}"
-  if ! cp -- "${state}" "${fence_lease_state_file}"; then
+  # Only recovery consumes the retained copy, and only the mode dispatch sets the
+  # path. Guarded so a caller that reports without recovering cannot turn an
+  # unset path into a failed `cp` and lose the report itself — which is read
+  # precisely when a deploy is already refusing to start.
+  if [[ -n "${fence_lease_state_file}" ]] &&
+    ! cp -- "${state}" "${fence_lease_state_file}"; then
     echo "::error::Could not retain the GHCR synchronization lease state."
     return 1
   fi
--- a/scripts/refresh-flux-ghcr-auth.sh
+++ b/scripts/refresh-flux-ghcr-auth.sh
@@ -19,10 +19,11 @@ require_flux_ghcr_yaml_tool
 check_only=false
 allow_incomplete_fanout=false
 report_fences=false
+recover_fences=false
 record_runtime_proof_path=""
 reuse_runtime_proof_path=""
 usage() {
-  echo "Usage: $0 [--check-only|--allow-incomplete-fanout|--fences|--record-runtime-proof PATH|--reuse-runtime-proof PATH]" >&2
+  echo "Usage: $0 [--check-only|--allow-incomplete-fanout|--fences|--recover-fences|--record-runtime-proof PATH|--reuse-runtime-proof PATH]" >&2
 }
 while (($# > 0)); do
   case "$1" in
@@ -36,6 +37,10 @@ while (($# > 0)); do
       ;;
     --fences)
       report_fences=true
+      shift
+      ;;
+    --recover-fences)
+      recover_fences=true
       shift
       ;;
     --record-runtime-proof | --reuse-runtime-proof)
@@ -66,10 +71,20 @@ fi
 # exits before any credential or proof work runs, so accepting it alongside an
 # operational mode makes that step exit 0 having silently skipped the operation
 # it was configured to perform — an automation failure that looks like a pass.
-if [[ "${report_fences}" == "true" ]] &&
+#
+# --recover-fences is the same kind of alternative mode and carries the same
+# hazard, so it is rejected alongside an operational mode for the same reason.
+# It is also rejected alongside --fences: --recover-fences already emits the
+# full report as the evidence it acts on, so accepting both would leave the
+# caller's intent ambiguous between "report" and "report and then mutate".
+if { [[ "${report_fences}" == "true" ]] || [[ "${recover_fences}" == "true" ]]; } &&
   { [[ "${check_only}" == "true" ]] ||
     [[ "${allow_incomplete_fanout}" == "true" ]] ||
     [[ -n "${record_runtime_proof_path}${reuse_runtime_proof_path}" ]]; }; then
+  usage
+  exit 64
+fi
+if [[ "${report_fences}" == "true" && "${recover_fences}" == "true" ]]; then
   usage
   exit 64
 fi
@@ -325,6 +340,13 @@ runtime_probe_bootstrap_needed=0
 # still held, states whether the holder is provably dead, and prints the exact
 # CAS-guarded release. It performs no mutation by design: an operator running
 # the printed command is the explicit step the fencing model requires.
+# State the fence report hands to --recover-fences. Declared here, beside the
+# report that populates them, so the two cannot drift apart.
+fence_non_lease_held=0
+fence_lease_holder=""
+fence_lease_heartbeat_live=false
+fence_lease_state_file=""
+
 fence_run_segment() {
   if [[ -n "${GITHUB_RUN_ID:-}" ]]; then
     printf 'gh%s.%s' "${GITHUB_RUN_ID}" "${GITHUB_RUN_ATTEMPT:-1}"
@@ -382,6 +404,15 @@ fence_report_lease() {
   holder="$(jq -r '.spec.holderIdentity // ""' "${state}")"
   [[ -n "${holder}" ]] || return 0
   held=$((held + 1))
+  # Retain what --recover-fences needs to decide, so recovery reasons over the
+  # SAME observation the report just printed rather than re-reading the Lease.
+  # A second read could land either side of a fresh acquisition, which would let
+  # recovery act on a holder the operator never saw evidence for.
+  fence_lease_holder="${holder}"
+  if ! cp -- "${state}" "${fence_lease_state_file}"; then
+    echo "::error::Could not retain the GHCR synchronization lease state."
+    return 1
+  fi
   duration="$(jq -r '.spec.leaseDurationSeconds // 0' "${state}")"
   now_epoch="$(date -u +%s)"
   renew_epoch="$(fence_report_epoch "$(jq -r '.spec.renewTime // ""' "${state}")")"
@@ -389,6 +420,7 @@ fence_report_lease() {
   printf '    holder: %s\n' "${holder}"
   if [[ -n "${renew_epoch}" ]] &&
     ((now_epoch - renew_epoch < duration)); then
+    fence_lease_heartbeat_live=true
     printf '    renewed %ss ago, inside its %ss duration — the holder is LIVE. Do not recover.\n' \
       "$((now_epoch - renew_epoch))" "${duration}"
   else
@@ -805,6 +837,12 @@ report_fences_now() {
     printf '\n'
   done <"${fence_report_nodes}"
 
+  # Captured BEFORE the Lease is reported, so it counts only the fences the
+  # Lease's own ordering rule says must be released first. --recover-fences
+  # refuses while this is non-zero rather than duplicating the node-journal
+  # grouping above, which is the part that is genuinely hard to get right.
+  fence_non_lease_held="${held}"
+
   fence_report_lease "${state}" || return 1
 
   if ((held == 0)); then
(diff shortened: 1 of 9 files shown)


devantler (commented):

**Reviewed commit:** `20d4282447a9a6dd8db869ac199f6516d69c3ecb`

Per-lane evidence, each checked directly at this exact head:

- **CodeRabbit** — head commit status `success :: Review rate limited`, posted 12:16:52Z (18s after the request). A quota refusal, not a verdict.
- **Codex** — `You have reached your Codex usage limits for code reviews` at 13:08:24Z. Only the maintainer can lift it.
- **Cursor Bugbot** — check-run `neutral` with `output.title: Error` and `Bugbot couldn't run - usage limit reached` at 13:22:12Z. Only the maintainer can lift it.

All three lanes were tried at `20d42824` and none will deliver, so the local round is admissible under *Local review round*.

**Verdict: 2 findings (P0: 0, P1: 2).**

### What I verified and found sound

- **The liveness proof is a real death proof, not an expiry inference.** `fence_run_is_terminal` refuses on a missing `gh`, a failed query, an empty status, and any status other than `completed` — four distinct NOT-PROVEN paths, each with its own test.
- **The attempt is pinned** (`/attempts/${run_attempt}`), so a finished attempt 2 cannot vouch for a running attempt 1. Asserted on the recorded request path.
- **Single observation.** Recovery reasons over the Lease state the report retained, never a second read, so it cannot act on a holder the report never showed.
- **The CAS ops are op-identical to the command the report prints for an operator** (verified by diffing the two builders) — the automated path is exactly as strong as the documented manual one, today.
- **Ordering is correct in the composite.** The step lands after `🔑 Restore kubeconfig` and `🩺 Verify prod cluster is reachable`, so `kubectl` works; and the Cilium autoscaling gate immediately above it is not one of the counted fences, so it cannot make the pre-flight refuse forever.
- **The `actions: read` grant reaches every caller** — all three jobs that consume the composite (cd.yaml deploy-prod, ci.yaml deploy-prod, ci.yaml heal) carry it. No caller passes the input, so the feature ships latent.

scripts/tests/refresh-flux-ghcr-auth/contracts_test.go
```diff
@@ -265,8 +265,19 @@ func TestDeployActionConsumerStagingPrecedesPublishAndIsReassertedAfterUpdate(t
...
+	// The fourth reference is the orphaned-fence pre-flight. It is asserted
+	// explicitly rather than absorbed into the count, so a future invocation still
+	// has to be accounted for here deliberately — which is what this exhaustive
+	// count exists to force.
+	// Anchored on the step, not its `run:` line: the gating `if:` precedes the
+	// command, so a slice starting at `run:` would miss the very condition that
+	// keeps this default-off.
+	recoverStep := requireIndex(t, action, "- name: 🧯 Recover an orphaned GHCR deploy fence")
+	requireBefore(t, recoverStep, firstRefresh, "fence recovery before staging")
+	requireContains(t, action[recoverStep:firstRefresh], "inputs.recover-orphaned-fence == 'true'")
```
devantler, line 277: P1 — the default-off safety property is not pinned by any test.

Everything that keeps this feature latent in production rests on `default: "false"` in `action.yml`. This assertion pins the *gate* but not the *default*, and the two fail independently: if the default flipped to `"true"`, `if: inputs.recover-orphaned-fence == 'true'` would still be present here, still match this assertion, and would now evaluate **true** on every deploy and every heal — arming automatic Lease mutation against prod with a green contract suite.

That is the one state this PR most needs pinned. *Feature-flag-first delivery* asks for both states tested, and the OFF state currently has no test at all.

Add an assertion on the input default alongside the gate — the input block is already in `action`, so it is a couple of lines here.

scripts/refresh-flux-ghcr-auth.sh
```diff
@@ -818,6 +861,111 @@ report_fences_now() {
...
+  run_id="${run_reference%% *}"
+  run_attempt="${run_reference##* }"
+  fence_run_is_terminal "${run_id}" "${run_attempt}" || return 1
+
+  # The same CAS the report prints for an operator: both `test` ops must still
+  # hold, so a holder that changed under us aborts the patch instead of losing a
+  # live acquisition.
+  resource_version="$(jq -r '.metadata.resourceVersion' "${fence_lease_state_file}")"
+  patch_file="${work_dir}/fence-recovery-patch.json"
+  jq -nc \
```
devantler: P1 — the release patch is now built twice, and nothing binds the copies together.

These four ops are byte-identical to the ones `fence_lease_release_command` builds (I diffed them; only the argument plumbing differs). That duplication is a live drift hazard on a safety-critical control: strengthening the operator-facing CAS later — adding a `test` on `renewTime`, say — leaves this automated copy on the weaker guard, silently. The automated path would then be *less* safe than the manual one it is documented as mirroring, with no test failing.

Extract the patch construction into one helper (e.g. `fence_lease_release_patch <state-file> <holder>`) and have both `fence_lease_release_command` and `recover_fences_now` call it, so the two cannot diverge. Simpler than the current shape, and it makes "the automation uses the same CAS as the runbook" a property rather than a coincidence.

scripts/tests/refresh-flux-ghcr-auth/contracts_test.go
devantler: > 🤖 Generated by the Agentic Engineer

**Fixed** in `038c7be7`.

The default is now read structurally via `yamlScalarAtPath(action, "inputs", "recover-orphaned-fence", "default")` and asserted to equal `false`, so a `default: "false"` belonging to some other input cannot satisfy it.

**Ablation proof:** flipping the value to `"true"` in `action.yml` fails this assertion and only this one — `recover-orphaned-fence default = "true", want "false"` — while the pre-existing `if:` assertion stays green. That is the independence the finding claimed, demonstrated rather than argued.

scripts/refresh-flux-ghcr-auth.sh
devantler: > 🤖 Generated by the Agentic Engineer

**Fixed** in `1612537c`.

Extracted `fence_lease_release_patch <state-file> <holder>`; both `fence_lease_release_command` and `recover_fences_now` now call it, so there is one builder and the automated path cannot drift onto a weaker CAS than the runbook it mirrors.

**Behaviour-preserving:** the full fence test package passes unchanged (120s, `ok`), including the green release that asserts the holder is actually cleared, the CAS-refusal test, and the report tests that pin the printed operator command. `shellcheck` clean; `shfmt -i 2 -ci` reports zero hunks, matching the clean-main baseline measured with the same CI arguments.

devantler (commented):

**Reviewed commit:** `1612537c065e4504121cec0da1b7a25e325e1f8e`

### Why a local round, per lane (direct per-PR checks at this head — not a digest field)

| Lane | State | Evidence |
|---|---|---|
| CodeRabbit | requested, acked, **not delivered** | request marker `5317221235` at 15:03:13Z; ack comment `5317225022` at 15:03:59Z is future tense (``I will review pull request `#3189` at `1612537c` ``) with no verdict line. ~90 min later the only review object with a body sits at the superseded head `eb2bafb5`. Third request today, second consecutive non-delivery. |
| Codex | **usage limit** | `You have reached your Codex usage limits for code reviews` (13:08:24Z). Weekly quota — no retry window. |
| Cursor Bugbot | **usage limit** | `Bugbot couldn't run - usage limit reached` (13:22:12Z), recorded as `outcome=no-gate reason=usage-limit-reached`. Monthly quota; only an admin can raise it. |

The two usage-limited lanes state no retry window, so re-triggering them at this head would post
requests that provably cannot succeed. The head's `CodeRabbit — Review completed` status is a
corroborator only and is **not** treated as a satisfier: no review object with a body exists here.

### What I checked

- **Safety property.** The release is gated on a *positive death proof* (the API reporting the
  holder's run attempt `completed`), never on lease expiry — which the script's own acquisition
  comment explains is unsafe because an expired shell can resume and write under CAS. Every
  unprovable path (`gh` absent, query failure, empty body, unrecognised status, no run reference)
  returns non-zero and refuses. I traced each branch to a `return 1`.
- **Attempt pinning.** `fence_run_is_terminal` pins `/attempts/${run_attempt}`. Unpinned, a finished
  attempt 2 would vouch for a still-running attempt 1 — the dangerous direction for an automatic
  path. `TestRecoverFencesPinsTheAttemptWhenProvingLiveness` asserts this against the *recorded*
  request, so it is a real assertion rather than a restatement of the code.
- **Ordering.** Recovery refuses while `fence_non_lease_held > 0`, so the global exclusion fence is
  never cleared ahead of node/policy fences.
- **Observation consistency.** The CAS patch is built from the state the report *retained*, not a
  fresh read, so the patch tests the same observation the decision was made on. Both `test` ops
  (`resourceVersion` + `holderIdentity`) must hold at apply time.
- **One builder.** `fence_lease_release_patch` is the single source for both the operator-facing
  command and the automated patch, so the automated path cannot drift onto a weaker CAS.
- **Default-off.** `recover-orphaned-fence` defaults to `"false"` and the step carries
  `if: inputs.recover-orphaned-fence == 'true'`. No caller sets it true, so this merges latent. The
  contract test anchors on the step name rather than its `run:` line, so the gating `if:` is inside
  the asserted slice.
- **Exercised, not reasoned about.** `go test ./scripts/tests/refresh-flux-ghcr-auth/` at this exact
  head: **ok, 128.578s**. I confirmed by process inspection during the run that the suite executes
  the real `scripts/refresh-flux-ghcr-auth.sh`, so these are behavioural tests, not mocks of the
  logic under test. 13 added cases cover the happy path and a negative for every refusal branch.

### Non-blocking observations (below the gate; no thread opened)

1. **nit — unanchored attempt assertion.** `strings.Contains(request, "/attempts/1")` also matches
   `/attempts/12`. Harmless with the current fixture (attempt 1), but it would mask an off-by-one if
   the fixture ever moved to a two-digit attempt. Worth tightening when the file is next touched.
2. **nit — suppressed diagnostic.** The liveness query uses `2>/dev/null`, so the operator-facing
   error names *that* the query failed but not *why*. Fail-closed and correct, but this path is read
   exactly when a deploy is already wedged, which is when the cause matters most.

Neither is a correctness defect and neither blocks: raising them inline would open threads whose only
resolution is a push that re-stales every green on a PR whose review lanes are exhausted.

PR #3210, opened 2026-08-18
security(k8s): disposition vendored KubeVirt and CDI operator RBAC as a class

devantler:

### Why

Trivy reports RBAC findings on the KubeVirt and CDI operator bundles, and they are the largest block of high-severity noise on the cluster manifests — 16 of the 25 HIGH/CRITICAL findings on the tree. They are also not ours to fix: both bundles are upstream release artifacts we vendor verbatim and verify against a recorded checksum, and every finding is a permission the operator genuinely needs to run. Editing them would break the operators and break the checksum. While they sit in the count, the findings that *are* ours are buried underneath them.

### What

Records a single, reasoned exception covering that class, so the remaining high-severity findings are the ones on cluster roles this repository actually authors. Trivy goes 109 → 93 overall and 25 → 9 at HIGH/CRITICAL; the 9 left are 7 first-party findings plus 2 deliberately kept reporting because they *are* fixable here.

The exception is deliberately narrow, and a new check fails the build if it ever stops being narrow — the risk with a suppression like this is not that it is wrong today, but that it later goes quiet on our own manifests while the shrinking count reads like progress.

This does not yet let trivy block the build; the medium and low findings remain, and #2787 owns those.

Part of #2990

### DIFF
--- a/scripts/tests/test-trivyignore-vendored-operator-boundary.sh
+++ b/scripts/tests/test-trivyignore-vendored-operator-boundary.sh
@@ -48,7 +48,10 @@ status=0
 
 # ---------------------------------------------------------------- structure --
 for check in "${checks[@]}"; do
-  entries="$(yq "[.misconfigurations[] | select(.id == \"$check\")] | length" "$ignorefile")"
+  # A yq failure here must take the MISSING DISPOSITION path below, not become an
+  # integer-expression error that evaluates false and silently treats the id as present.
+  entries="$(yq "[.misconfigurations[] | select(.id == \"$check\")] | length" "$ignorefile" 2>/dev/null || printf '0')"
+  entries="${entries:-0}"
   if [ "$entries" -lt 1 ]; then
     printf 'MISSING DISPOSITION: %s has no entry in .trivyignore.yaml\n' "$check" >&2
     status=1
@@ -136,10 +139,21 @@ write_probe "$first_party"
 cp "$ignorefile" "$scratch/.trivyignore.yaml"
 
 findings="$scratch/findings.txt"
-(cd "$scratch" && trivy fs --scanners misconfig --exit-code 0 \
-  --ignorefile .trivyignore.yaml --format json --quiet . 2>/dev/null) |
-  jq -r '.Results[]? as $r | $r.Misconfigurations[]?
-         | select(.Status=="FAIL") | "\($r.Target)\t\(.ID)"' |
+trivy_json="$scratch/trivy.json"
+trivy_err="$scratch/trivy.err"
+
+# trivy's stderr is kept rather than discarded: `set -o pipefail` means a trivy failure would
+# otherwise exit this script non-zero with no diagnostic, which is the "guardrail that blocks
+# without naming the fix" shape. Run it on its own so its status is checked directly.
+if ! (cd "$scratch" && trivy fs --scanners misconfig --exit-code 0 \
+  --ignorefile .trivyignore.yaml --format json --quiet . >"$trivy_json" 2>"$trivy_err"); then
+  printf 'trivy failed while scanning the probe fixture; its stderr follows\n' >&2
+  cat "$trivy_err" >&2
+  exit 1
+fi
+
+jq -r '.Results[]? as $r | $r.Misconfigurations[]?
+       | select(.Status=="FAIL") | "\($r.Target)\t\(.ID)"' "$trivy_json" |
   sed 's|^\./||' | sort -u >"$findings"
 
 [ -s "$findings" ] || {
--- a/scripts/tests/test-trivyignore-vendored-operator-boundary.sh
+++ b/scripts/tests/test-trivyignore-vendored-operator-boundary.sh
@@ -17,9 +17,10 @@
 #
 # Verified here in two layers, because they fail for different reasons:
 #
-#   STRUCTURE (always) — every dispositioned check id still carries a non-empty `paths:` list, and
-#   every path in it is one of the two vendored bundles, written literally. This is the layer that
-#   catches the regression above, and it needs nothing but yq.
+#   STRUCTURE (always) — every dispositioned check id is scoped to EXACTLY the two vendored bundles,
+#   written literally: no entry may lose its `paths:` key, gain a path outside them, or drop one of
+#   them. It needs nothing but yq, so it is the layer that still holds when the behavioural control
+#   below is skipped.
 #
 #   BEHAVIOUR (when trivy is installed) — the paired control the KSV-0037 entry describes: the same
 #   offending ClusterRole bytes are written to a vendored path AND to a first-party path, and the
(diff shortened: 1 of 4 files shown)


scripts/tests/test-trivyignore-vendored-operator-boundary.sh
```diff
@@ -0,0 +1,155 @@
...
+  while IFS= read -r path; do
+    [ -n "$path" ] || continue
+    case "$path" in
+      "$vendored_cdi" | "$vendored_kubevirt") ;;
+      *)
+        printf 'WIDENED SKIP: %s is scoped to %s, which is not one of the two vendored operator bundles\n' "$check" "$path" >&2
+        status=1
+        ;;
+    esac
+  done < <(yq ".misconfigurations[] | select(.id == \"$check\") | (.paths // [])[]" "$ignorefile")
```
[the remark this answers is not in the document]
devantler: > 🤖 Generated by the Agentic Engineer

Valid — fixed in `68a000a5`.

Reproduced first: with `KSV-0046` reduced to the CDI path only and `trivy`/`jq` off `PATH`, the
pre-fix test exited **0**. Exactly as described, the structural layer rejected unexpected paths but
accepted a missing expected one, and the behavioural half that would have caught it never ran.

The path loop now tracks each expected bundle and fails when either is absent. Four ablations, all
in structure-only mode:

| Ablation | Result |
|---|---|
| `KSV-0046` loses the kubevirt path | `NARROWED SKIP` → exit 1 (pre-fix test: exit **0**) |
| `KSV-0053` loses the cdi path | `NARROWED SKIP` → exit 1 |
| `KSV-0114` loses its `paths:` key | `UNSCOPED SKIP` **and** both `NARROWED SKIP`s → exit 1 |
| `KSV-0041` widened to `k8s/**` | `WIDENED SKIP` **and** the matching `NARROWED SKIP` → exit 1 |

The last two are regression controls: both pre-existing checks still fire. `.trivyignore.yaml` was
restored byte-for-byte after each ablation (`git status` shows only the test file changed), and the
unablated run is green in both modes — structure-only, and with the behavioural control active.

The header comment claiming this layer "catches the regression" was overstated for the dropped-path
case, so it now states what the layer actually pins.

scripts/tests/test-trivyignore-vendored-operator-boundary.sh
```diff
@@ -0,0 +1,171 @@
...
+}
+
+write_probe "$vendored_cdi"
+write_probe "$vendored_kubevirt"
+write_probe "$first_party"
+cp "$ignorefile" "$scratch/.trivyignore.yaml"
+
+findings="$scratch/findings.txt"
+(cd "$scratch" && trivy fs --scanners misconfig --exit-code 0 \
+  --ignorefile .trivyignore.yaml --format json --quiet . 2>/dev/null) |
```
devantler: > 🤖 Generated by the Agentic Engineer

**nit** — `trivy`'s stderr is discarded here, and `set -o pipefail` is in force, so a trivy failure exits the script non-zero with no diagnostic at all. The vacuity guard below catches an empty result, but a CI reader sees a bare non-zero exit. Capturing stderr to the scratch dir and echoing it on failure would keep the guardrail's "fail with the fix" property.

scripts/tests/test-trivyignore-vendored-operator-boundary.sh
```diff
@@ -0,0 +1,171 @@
...
+readonly first_party='k8s/bases/infrastructure/cluster-roles/boundary-probe.yaml'
+
+# Every check id dispositioned for the vendored bundles. Keep in sync with .trivyignore.yaml.
+readonly checks=(KSV-0041 KSV-0046 KSV-0053 KSV-0056 KSV-0114)
+
+status=0
+
+# ---------------------------------------------------------------- structure --
+for check in "${checks[@]}"; do
+  entries="$(yq "[.misconfigurations[] | select(.id == \"$check\")] | length" "$ignorefile")"
```
devantler: > 🤖 Generated by the Agentic Engineer

**nit** — if `yq` fails here (malformed ignorefile), `entries` is empty and `[ "$entries" -lt 1 ]` is an integer-expression error rather than a clean failure. Inside `if` it does not trip `set -e`, so it evaluates false and the id is silently treated as present. Defaulting to `0` would turn that into the `MISSING DISPOSITION` path it was meant to take.

devantler (commented):

## Self-review (fallback — CodeRabbit, Codex and Cursor Bugbot all unavailable)

**Reviewed commit:** `68a000a56ed0da2b029e88cd0daf07765bbf28b0`

Per-lane evidence, checked directly on this PR at this head rather than inferred from a digest:

| Lane | Result at this head | When |
|---|---|---|
| CodeRabbit | `Review rate limited` (head commit status; `cr:no-gate@68a000a5` recorded) | 2026-08-18T07:18:23Z |
| Codex | account-level *"You have reached your Codex usage limits for code reviews"* | 2026-08-18T09:16:27Z |
| Cursor Bugbot | check-run `neutral` + `output.title: Error`, *"Bugbot couldn't run - usage limit reached"* | 2026-08-18T09:20:12Z |

Two of the three are vendor spend/usage ceilings only a dashboard admin can lift, and the third
stated no retry window, so waiting would park a finished change rather than review it.

### What I checked

The substance here is a **security exception**, so I reviewed it against the fix-vs-except ladder
rather than for style. The disposition covers 16 of 25 HIGH/CRITICAL findings, which is a large
suppression and deserves the scrutiny.

It holds up:

- **Excepting rather than fixing is the right rung.** Both bundles are upstream release artifacts
  fetched at a pinned version and SHA-256 verified by `update-vendored-operators.sh`. Narrowing the
  grants would both break the operators and diverge the bundle from the checksum it is verified
  against, so there is no manifest-level fix short of forking upstream.
- **The exception is scoped, not blanket.** Every entry carries an explicit two-path `paths:` key,
  written literally rather than through a YAML anchor — which is the right call, since an alias
  resolving to an empty list would read as an unscoped skip. That is not a hypothetical: it is the
  exact failure mode recorded on this repo (`n_paths: 0` = blanket skip).
- **The boundary is enforced, not documented.** The new test asserts both structurally (every id
  scoped to exactly the two bundles) and behaviourally (identical offending ClusterRole bytes
  written to a vendored path and a first-party path; the suppression must reach only the first).
- **It cannot pass vacuously.** The behavioural half fails closed when the probe produces no
  findings at all, and the "BOUNDARY LOST" loop requires all five checks to still fire on the
  first-party path — so a blanket suppression fails the test instead of satisfying it.
- **The test is actually wired.** The new script is added to the change-filter path list, so
  editing it triggers the job that runs it. A test the pipeline never invokes would guard nothing.
- **KSV-0014 is deliberately left reporting** on the two operator Deployments, correctly: a
  read-only root filesystem is a container setting a kustomize patch can supply without touching
  upstream rules.

I also checked the two counts that look inconsistent and they are not: 25 total − 16 dispositioned
= 9 remaining, which is the 7 on first-party cluster roles plus the 2 KSV-0014 Deployment findings.
The `109 → 93` note correctly separates a v0.73→v0.74 scanner rule-set difference from real backlog.

### Findings

Two nits, both about diagnosability rather than correctness. Neither blocks.

**Verdict: no P0/P1 findings**

scripts/tests/test-trivyignore-vendored-operator-boundary.sh
devantler: > 🤖 Generated by the Agentic Engineer

**Fixed at `3de08b94`.** trivy stderr is now captured to the scratch dir and printed before the non-zero exit; trivy also runs on its own so its status is checked directly rather than being swallowed by the pipeline.

Verified: shellcheck clean, the test passes with the behavioural layer running (trivy 0.74.0), and the ablation — widening KSV-0041 onto the first-party probe path — still trips both `WIDENED SKIP` and `BOUNDARY LOST` with exit 1.

scripts/tests/test-trivyignore-vendored-operator-boundary.sh
devantler: > 🤖 Generated by the Agentic Engineer

**Fixed at `3de08b94`.** entries now defaults to 0 on a yq failure, so a malformed ignorefile takes the MISSING DISPOSITION path instead of evaluating false.

Verified: shellcheck clean, the test passes with the behavioural layer running (trivy 0.74.0), and the ablation — widening KSV-0041 onto the first-party probe path — still trips both `WIDENED SKIP` and `BOUNDARY LOST` with exit 1.

Issue #3217, opened 2026-08-18
security(kubescape): C-0211's residual population behind the pod-security exception is unsized and unowned


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`pod-security-mutations` suppresses **C-0211** alongside C-0013/C-0016/C-0055, cluster-wide. The
other three have been sized: #3064 measured their residual population against prod and closed
COMPLETED once the three `kubescape` Deployments were hardened in-spec.

C-0211 was explicitly excluded from that sizing. The exception's own annotation says so:

> C-0211 is deliberately NOT sized here. It has no other exception, its field set differs from the
> three above (fsGroupChangePolicy, runAsUser and runAsGroup are supplied by the mutation), and the
> name-scoped `kubescape-privileged.yaml` does not cover it — so node-agent and storage remain
> inside its residual population rather than excepted out of it.
> Sizing it needs its own per-workload pass; #3064 carries that.

**That last pointer no longer holds.** #3064 is CLOSED/COMPLETED, and it never carried the general
per-workload pass — its only C-0211 acceptance criterion was narrow: *"`C-0211` coverage for
`node-agent` is decided explicitly"*. So the pass the annotation defers to does not exist and is
owned by nothing.

## Impact

C-0211 is the one control in this exception with **no** compensating coverage anywhere: it has no
other exception, and `kubescape-privileged.yaml` does not name it, so `node-agent` and `storage` sit
inside its residual population rather than being excepted out of it. Suppression is silent and
indistinguishable from compliance, so nothing here would show a regression — the failure mode #2447
exists to prevent.

This also blocks the narrowing in #2824: the scope is now expressible (`NotIn`, #3083), but the set
that would surface for C-0211 is unknown, so a narrowing would be sized by guess.

## Expected

The residual population for C-0211 is enumerated per workload, and each entry is either hardened
in-spec or carries a narrow, workload-scoped exception whose reason is true of it.

## Acceptance criteria

- [ ] Every workload in the twelve namespaces `add-security-context` excludes is checked against
      C-0211's field set (`fsGroupChangePolicy`, `runAsUser`, `runAsGroup`), and the result recorded.
- [ ] `node-agent` and `storage` each get an explicit decision — hardened, or named in a
      workload-scoped exception stating why.
- [ ] The measurement is re-run rather than inherited: the 2026-08-10 snapshot in the annotation is
      prod-only and the local overlay opts in to a different namespace set.
[...]

Issue #2824, opened 2026-07-27
security: the cluster-wide pod-security exception suppresses controls where its own compensating control does not run


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`pod-security-mutations` ([`k8s/bases/infrastructure/cluster-security-exceptions/pod-security-mutations.yaml`](k8s/bases/infrastructure/cluster-security-exceptions/pod-security-mutations.yaml))
suppresses **C-0013, C-0016, C-0055 and C-0211 cluster-wide** — it carries no `match:` block, and
`scripts/generate-kubescape-exceptions/main.go` maps a missing match to `kind: ".*"` ("No match => the
exception applies to every resource").

Its stated justification is that the compensating control runs at admission:

> Kyverno mutate policy "add-security-context" injects these controls … at pod admission. The
> Deployment spec does not contain them and Kubescape scans the stored spec, not the live pod.

**That justification does not hold for its own scope.** `add-security-context`
([`cluster-policies/best-practices/add-security-context.yaml`](k8s/bases/infrastructure/cluster-policies/best-practices/add-security-context.yaml))
**excludes ten namespaces** from both `add-pod-security-context` and `add-container-security-context`.
In those namespaces Kyverno injects nothing — so the exception suppresses the controls in exactly the
places where the compensating control is absent.

The two are inverses: the exception is justified by a mutation whose scope is the complement of where
the exception matters most.

## Impact

Suppression is silent and indistinguishable from compliance. A workload in an excluded namespace that
genuinely runs as root reads the same as one that is correctly hardened, so the posture score cannot
be used to detect a regression there — which is the failure mode the 100%-and-hold programme (#2447)
exists to prevent. This is the "growing exceptions set is a smell" case from the fix-vs-except ladder:
the exception is doing more work than its rationale claims.

Concretely, #2435 spent effort hardening two workloads on the belief the score would move; it could
not have, and neither the CI gate nor the in-cluster scan would have shown the difference (measured
while reviewing #2822 — both trees "All controls passed").

## Expected

The exception should cover only what its rationale covers: workloads where Kyverno actually injects
the fields. Everything else should either genuinely pass, or carry its own **narrow, workload-scoped**
exception with a reason that is true of it — the `kind`+`name` designator pattern
`kubescape-privileged.yaml` already uses.

## Acceptance criteria

- [ ] `pod-security-mutations` no longer suppresses C-0013/C-0016/C-0055/C-0211 for namespaces excluded
[...]

Issue #3083, opened 2026-08-11
feat(kubescape): support complementary namespace exception scopes


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Parent #2824 cannot narrow `pod-security-mutations` to the namespaces where Kyverno actually mutates workloads with the generator's current contract. `scripts/generate-kubescape-exceptions/convertNamespaceSelector` accepts only `kubernetes.io/metadata.name In [...]`; the required scope is the complement of the mutation policy's excluded namespaces, and a positive `In` inventory would fail every newly onboarded namespace until someone manually extended it.

Kubernetes label selectors already express that scope as `NotIn`. The in-cluster `ClusterSecurityException` can consume that standard operator, but the offline CI and Headlamp generator currently reject it, so the three exception consumers cannot share one source of truth.

## Impact

Without complementary selector support, #2824 must choose between a cluster-wide exception that silently masks genuine posture gaps and a brittle namespace allow-list that turns routine namespace onboarding into a CI outage. That prevents the parent from narrowing suppression safely.

## Expected behavior

The generator accepts `kubernetes.io/metadata.name NotIn [...]` and emits an anchored Kubescape designator whose observable matching behavior rejects exactly the excluded namespaces while accepting existing and newly introduced namespaces. Existing `In` behavior remains unchanged, and every other selector shape still fails closed.

## Acceptance criteria

- [ ] `NotIn` is accepted only for the supported namespace-name key with a non-empty string list.
- [ ] The emitted regex rejects every excluded namespace and accepts ordinary and previously unseen namespaces; metacharacters in excluded values cannot widen or corrupt the match.
- [ ] Unit tests exercise those matching semantics against the real generated designator and first fail against current `main`.
- [ ] Existing `In` output and fail-closed handling for unsupported operators remain unchanged.
- [ ] The generated Headlamp mirror drift check and `go test ./scripts/generate-kubescape-exceptions` pass.
- [ ] This PR lands latent support only. Changing `pod-security-mutations` remains in parent #2824 after the measured C-0055 remediation in #3064.

## Size

Small, one focused generator capability plus tests. No live-cluster mutation and no exception-scope change.

Part of #2824

Issue #3064, opened 2026-08-10
security(kubescape): three Deployments genuinely fail C-0055 behind the blanket pod-security exception


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`pod-security-mutations` suppresses C-0013/C-0016/C-0055/C-0211 cluster-wide, and the compensating
Kyverno mutation does not run in the namespaces it excludes (#2824). Measuring what that actually
hides, against prod on 2026-08-10:

- Four excluded namespaces — `kube-public`, `kube-node-lease`, `local-path-storage`, `chaos-mesh` —
  run **no workloads**, so nothing is suppressed there.
- Of the 24 workloads in the remaining namespaces, **21 already satisfy these controls in their own
  specs** (`flux-system`, `cert-manager`, `keda`, `opencost` and `kubescape/alertmanager` are fully
  hardened container-side).
- **Three do not**: the `kubescape`, `kubevuln` and `operator` Deployments in the `kubescape`
  namespace each set `runAsNonRoot` and `allowPrivilegeEscalation: false`, but carry **no
  `capabilities.drop` and no container-level `seccompProfile`** — so C-0055 is genuinely unmet, not
  merely unscanned.
- `kubescape/node-agent` is legitimately privileged (eBPF) and already covered by name in
  `kubescape-privileged.yaml` for C-0013/C-0016/C-0055.

## Problem

The blanket exception was declared rather than narrowed because the remediation was unsized. It is
now sized, and it is small: three workloads need a genuine in-spec fix, and the rest would pass on
their own. Until they are hardened, the suppression is doing real work for those three — which is
the case #2824 exists to remove.

## Expected

The three Deployments drop all capabilities and set a container-level `seccompProfile`, so the
controls pass genuinely rather than being suppressed. These are upstream chart defaults, so the fix
belongs in the HelmRelease values (or a scoped exception if a component turns out to need a
capability).

## Acceptance criteria

- [ ] `kubescape`, `kubevuln` and `operator` set `capabilities.drop: [ALL]` and a container-level
      `seccompProfile` in their own specs, or carry a workload-scoped exception whose reason is true.
- [ ] Verified against the running workloads, not only the rendered manifests.
- [ ] `C-0211` coverage for `node-agent` is decided explicitly (it is exempted in no CR today).
- [ ] The compliance floor is re-baselined upward if the score moves; never lowered.

## Size

Small. This is the remediation half of the measurement #2824 asked for; narrowing the blanket
[...]

PR #3218, opened 2026-08-18
docs(kubescape): correct three stale pointers in the pod-security exception annotation

devantler:

### Why

The cluster-wide pod-security exception carries a long annotation explaining why it is wider than its
own rationale and what would be needed to narrow it. Three of its statements have since stopped being
true, and nothing sweeps a note written inside a file: it says narrowing is impossible with the tools
we have, and it points at work that is finished as though it were still running.

The cost is real rather than cosmetic. Anyone picking up the narrowing reads "this cannot be done
yet" and stops — which is what happened here, until the pointers were checked against the code and
the issue tracker instead of being believed. A security exception that is documented as un-narrowable
stays un-narrowed.

### What

Corrects the three statements and marks what is now superseded, keeping the original dated
measurement intact as the historical record it is:

- Narrowing **is** now expressible — the missing capability shipped, so the blocker is evidence
  (the surfaced set still has to be re-measured), not tooling.
- The three workloads the annotation lists as unhardened have since been hardened in their own specs,
  and that work is closed.
- The pointer for the one control that was deliberately left unsized aimed at an issue that never
  covered it and has since closed. That work was therefore owned by nobody; it is now filed as #3217
  and linked under the parent.

No behaviour change — the exception itself is untouched, and both overlays validate.

Part of #2824

### DIFF
--- a/k8s/bases/infrastructure/cluster-security-exceptions/pod-security-mutations.yaml
+++ b/k8s/bases/infrastructure/cluster-security-exceptions/pod-security-mutations.yaml
@@ -24,9 +24,12 @@ metadata:
     # C-0013/C-0016/C-0055/C-0211 with nothing injecting them. Six of the twelve
     # are separately and deliberately excepted for C-0013/C-0016/C-0055 in
     # infrastructure-privileged.yaml; C-0211 and the other six are not covered
-    # anywhere. Narrowing it needs a scope the CSE designators cannot yet spell
-    # (an In-list would fail every newly-onboarded namespace), which is why this
-    # is declared rather than fixed here.
+    # anywhere. The complement scope this needs IS now expressible: #3083 added
+    # `NotIn` support to scripts/generate-kubescape-exceptions, rendered by
+    # namespaceNotInPattern, which is exactly what avoids an In-list failing
+    # every newly-onboarded namespace. What still blocks narrowing is evidence,
+    # not expressiveness — the surfaced set must be re-measured first (#2824),
+    # and C-0211's residual population is unsized (#3217).
     #
     # cert-manager, keda and opencost are NOT part of that set even though
     # add-pod-security-context excludes them: add-imageuid-pod-security-context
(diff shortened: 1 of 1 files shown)


devantler (commented):

**Reviewed commit:** `fcb96f02d6bbf48abe1d38464e404653bb7e4267`

Each lane was checked directly on this PR at this head, not inferred from an aggregate:

| Lane | Outcome at this head | When |
|---|---|---|
| CodeRabbit | `Review limit reached` — "all 1 included review currently available", next in ~33 min | 2026-08-18T15:40:05Z |
| Codex | Usage limit for code reviews, **no retry window**; credits are account-admin only | 2026-08-18T15:41:22Z |
| Cursor Bugbot | Check-run `neutral` / `Error`, comment "Bugbot couldn't run - usage limit reached"; admin-only to lift | 2026-08-18T15:42:00Z |

Two of the three state no retry window, so this is a billing/quota condition rather than a slow lane.

### What I checked

**Behaviour.** Comment-only; the CR's `spec` — `reason` and the four `posture` entries — is untouched, so the exception suppresses exactly what it did before. Verified rather than asserted: the exceptions generator's output is byte-identical against `origin/main` (39,345 bytes), and a control confirms the comparison discriminates — perturbing a `reason` field *does* move that output. Both overlays validate, 582 files each.

**Factual claims.** Each corrected pointer was checked against code and issue state rather than the annotation it replaces. `NotIn` support is present on `main` (`namespaceNotInPattern`, and the parser accepts `In`/`NotIn`), #3083 is CLOSED/COMPLETED, #3064 is CLOSED/COMPLETED and its only C-0211 acceptance criterion was `node-agent`, and #3217 is open and linked as a real sub-issue of #2824.

**Security.** No control is suppressed, added, or re-scoped. The change narrows nothing and widens nothing; it removes a false claim that narrowing is impossible, which is the opposite of relaxing a guardrail.

**Historical record.** The dated 2026-08-10 measurement is preserved verbatim with a dated supersession notice beside it, rather than rewritten — so the evidence trail stays intact while the stale forward-looking pointers are corrected.

### Observation (not a finding)

This correction introduces fresh pointers (#3083, #3217) that can go stale the same way these did — this is the second time this file has misdirected work. That is not fixable inside a comment-only PR, and it is already carried as an acceptance criterion on #3217 ("written where the next reader will find it, so this does not become a third stale pointer"). Recording it so the pattern is visible rather than silently repeated.

PR #3220, opened 2026-08-18
chore(security): rule on the vendored CDI and KubeVirt operator RBAC findings

devantler:

### Motivation

Kubernetes misconfiguration scanning still can't be enforced, because the scan reports a backlog of
findings that nobody has ruled on yet. Just under half of what was left came from two files we did
not write: the vendored CDI and KubeVirt operator releases, which we ship exactly as upstream
publishes them.

The findings say those two operators have broad permissions. They do — because that is what the
operators are for. CDI creates the pods that load a disk image; KubeVirt creates and upgrades the
components that make KubeVirt itself run. Narrowing either grant would stop them working, and we
cannot edit the files without breaking the way we take upstream's releases.

### What this changes

Records a ruling for the three permission checks that fire on those two files, each scoped to just
those files, so the checks stay fully enforced everywhere else. Every ruling names why that specific
grant is legitimate and was checked against the actual rules rather than assumed.

The remaining backlog drops from 66 findings to 45. Nothing that gets deployed changes.

Deliberately **not** covered: ten findings on the same two files about the operators' own resource
limits and read-only root filesystem. Those describe real, fixable state — one is HIGH — so they are
left reported and want a proper fix, not a ruling.

Part of #2787

### DIFF
--- a/.trivyignore.yaml
+++ b/.trivyignore.yaml
@@ -181,3 +181,32 @@ misconfigurations:
       Both operators register, update and remove their own validating and mutating admission
       webhooks as part of installing and upgrading themselves. The grant is upstream's, in a
       SHA-256 pinned release artifact vendored verbatim.
+  - id: KSV-0048
+    paths:
+      - k8s/bases/infrastructure/controllers/cdi/cdi-operator.yaml
+      - k8s/bases/infrastructure/controllers/kubevirt/kubevirt-operator.yaml
+    statement: >-
+      Creating and reconciling workloads is what both operators are for. CDI creates the importer,
+      uploader and cloner pods that populate a DataVolume; virt-operator creates and upgrades the
+      virt-api, virt-controller and virt-handler deployments that make up KubeVirt itself. A grant
+      narrow enough to satisfy this check would stop either operator installing or running. The
+      grant is upstream's, in a SHA-256 pinned release artifact vendored verbatim.
+  - id: KSV-0049
+    paths:
+      - k8s/bases/infrastructure/controllers/cdi/cdi-operator.yaml
+      - k8s/bases/infrastructure/controllers/kubevirt/kubevirt-operator.yaml
+    statement: >-
+      Both operators own the configmaps that configure the components they install, including their
+      generated CA bundles, and must create them to install and rotate them. The grant is upstream's,
+      in a SHA-256 pinned release artifact vendored verbatim.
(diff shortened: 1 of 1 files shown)


devantler (commented):

**Reviewed commit:** `8a834dea890047f3cf76aefa279bf4e58193ab32`

Per-lane evidence, each checked directly on this PR at this head:

| lane | outcome | when |
|---|---|---|
| CodeRabbit | `Review rate limited` / `Review limit reached`, next review in 30 minutes | 17:16:09Z |
| Codex | account usage limit for code reviews, **no retry window**; needs credits | 17:17:09Z |
| Cursor Bugbot | check-run `neutral` + `output.title: Error` — the run never started; `usage limit reached`, admin-only to lift | 17:17:54Z |

Two of the three state no retry window, so this is a billing condition rather than a slow lane.

### What I checked

**Are the statements true?** Every asserted grant was read at the cited rule. This caught one real
error pre-push: the first KSV-0113 draft claimed "neither operator is granted update or delete on a
secret", generalised from four of five sites; the fifth (`kubevirt-operator.yaml` @6961) grants
`create/get/list/watch/patch/delete`. Corrected — and the true fact is stronger, since that rule is
`resourceNames`-scoped to sixteen named `kubevirt-*` cert secrets.

**Does it over-suppress?** No. The entries are path-scoped, and the total lands on exactly 66−21=45 —
had a path leaked to other files the count would be lower. Residuals confirm per-check: KSV-0048
14→5, KSV-0049 8→1, KSV-0113 7→2, each the non-vendored remainder. An ablation removing only the
KubeVirt path from KSV-0048 returned 45→50, exactly its 5 findings.

**Does it hide real exposure?** The three checks concern RBAC breadth that is inherent to what these
operators do. Separately, neither bundle is deployed to any cluster today.

**Deliberate design decision, flagged for a future reader:** the statements justify each skip on the
operators' inherent needs and do *not* cite the not-deployed fact. That is intentional — deployment
status can change silently, and a justification resting on it would go stale while still reading as
authoritative. The not-deployed context is recorded on #2787 instead, where it is dated.

**Scope restraint:** the ten workload-hardening findings on these same two files (KSV-0011/0014/0018/
0020/0021, one HIGH) are deliberately left reported rather than swept in with the RBAC class. They
describe real, fixable state.

Verdict: no P0/P1 findings

Issue #3202, opened 2026-08-18
security(vault-backup): retire the CKV_K8S_40 exception by making snapshots group-readable, then raising the writers' UIDs


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`CKV_K8S_40` (*containers should run as a high UID to avoid host conflict*) is the **last** checkov
finding on `k8s/`: a full local reproduction of MegaLinter's invocation reports
`kubernetes … Failed checks: 2`, and both are the vault-backup workloads —
`Job.openbao.vault-snapshot-init` and `CronJob.openbao.vault-snapshot`. #2904 dispositioned them as a
scoped, deferred exception; this issue is the durable fix that retires it.

Three facts were measured on the live cluster while dispositioning them, and together they decide the
shape of the fix:

1. **User namespaces are NOT in effect at these sites, so the risk is a host UID.** The
   `add-user-namespaces` rule matches on the namespace label
   `pod-security.devantler.tech/user-namespaces: enabled`, which live carries exactly
   `backstage`, `crossview`, `headlamp`, `homepage`, `umami`, `whoami` — **not** `openbao`
   (`kubectl get ns openbao -o jsonpath='{.metadata.labels}'` returns no such label). The live
   CronJob's `hostUsers` is unset and the running pod reports `HOSTUSERS <none>`. So container
   UID 100 maps straight through to host UID 100.
   ⚠️ This refutes the basis #2904 originally proposed for accepting the finding — the
   user-namespace rollout is real, but it does not reach these two workloads.
2. **The snapshot lands `-rw-------`.** From the nightly job's own `ls -l`:
   `-rw------- 1 openbao openbao 459839 Aug 17 03:30 /snapshots/openbao-20260817-033025.snap`.
   Owner is UID 100, group GID 1000 — and there is **no group read bit**.
3. **That is why the `mirror` container inherits 100.** `minio/mc` mounts `/snapshots` `readOnly`
   and bakes no uid-100 entry of its own, so it reads the snapshot purely as its **owner**. Drop it
   to a high UID today and the off-cluster R2 mirror stops being able to open the file — silently,
   because the container would still start.

## Why this needs its own slice

`fsGroup: 1000` is not enough on its own. It fixes ownership and mode at **mount** time; the snapshot
is created **afterwards** by `bao operator raft snapshot save` at the process umask, which is what
produces the `0600` above.

## Smallest useful change, in order

[...]

PR #3222, opened 2026-08-18
ci(security): fail when a checkov skip reason is truncated by YAML comment syntax

### LINKED ISSUE #3204: Guard against a checkov skip reason being silently truncated by YAML comment syntax

## Evidence

A `checkov.io/skipN` value that names a tracking issue is silently truncated when the YAML scalar is
unquoted: ` #` starts a comment, so

```yaml
checkov.io/skip2: CKV_K8S_40=deferred to #3202 -- <the whole reason>
```

parses as `CKV_K8S_40=deferred to`. Checkov still suppresses the finding — the check id sits before
the `#` — so nothing fails, and the **stated reason disappears** with no signal at all.

Reproduced while writing #3203 (caught in self-review, fixed in the same PR by quoting). A repo-wide
sweep of the 14 files carrying these annotations found no other truncated value today, so this is a
guard against recurrence rather than a backlog.

## Why it matters

`.mega-linter.yml` states the rule this defeats: *"No suppression lands without its stated reason
being checked against the manifest it describes."* A reason that parses away leaves a bare
suppression that reads as reviewed — the same shape #2898's withdrawn skips had, arrived at by a
different route.

The two ways it goes wrong are both invisible: an unquoted value containing ` #`, and a value whose
text after `=` is empty.

## Smallest useful change

A test alongside the existing `scripts/tests/` that walks every `checkov.io/skipN` annotation in
`k8s/`, parses it as YAML, and fails when the parsed value has no reason text after `=` — or when
the raw line contains a ` #` the parsed value does not.

## Acceptance criteria

- [ ] The test fails on a fixture with an unquoted ` #`-bearing skip value and on one with an empty
      reason, and passes on the current tree.
- [ ] It runs in CI on changes under `k8s/`.

Rough size: small.

devantler:

### Motivation

A security suppression is only acceptable because someone wrote down *why*. Today that written reason can vanish without anything failing: if the annotation value is unquoted and mentions an issue number, YAML treats ` #` as the start of a comment and throws the rest away. The suppression keeps working — the check id survives — so what is left is a bare exception that still reads as reviewed. `.mega-linter.yml` requires the opposite: no suppression lands without its reason being checked.

### What

Makes the failure impossible instead of trying to spot it after the fact: every `checkov.io/skipN` value must now be an explicitly quoted scalar, which CI checks structurally. A quoted `#` is just a character, so a quoted reason cannot be truncated.

That inversion matters because detecting truncation is not merely hard, it is impossible — a complete reason followed by an ordinary comment and a reason cut in half by one are indistinguishable in the file, and only the author knows which was meant. The first attempt here tried anyway, and got it wrong in both directions.

The 21 annotations that were unquoted are quoted in this PR. No reason changes meaning: every parsed value is identical before and after.

Fixes #3204
Part of #2787

### DIFF
--- a/scripts/guard-checkov-skip-reasons.sh
+++ b/scripts/guard-checkov-skip-reasons.sh
@@ -154,7 +154,10 @@ while IFS= read -r file; do
   # Full-line comments are dropped first: a `#`-leading line is never an annotation
   # key, in a comment or inside a block scalar, and this tree does carry prose
   # cross-references shaped like one.
-  raw_keys=$(grep -vE '^[[:space:]]*#' -- "$file" | grep -cE "$key_shape")
+  # `grep -o | wc -l` counts each KEY, not each matching line: a flow map can carry
+  # several annotations on one line, and counting lines there would under-count and
+  # reject valid YAML as uncheckable.
+  raw_keys=$(grep -vE '^[[:space:]]*#' -- "$file" | grep -oE "$key_shape" | wc -l | tr -d ' ')
   raw_rc=$?
   if [ "$raw_rc" -gt 1 ]; then
     die "could not count annotations in $file (grep exit $raw_rc)"
--- a/scripts/guard-checkov-skip-reasons.sh
+++ b/scripts/guard-checkov-skip-reasons.sh
@@ -159,7 +159,14 @@ while IFS= read -r file; do
     -- "$file" 2>&1); then
     die "could not scan $file for block scalars: $blocks"
   fi
-  if printf '%s' "$blocks" | jq -r '.[]' 2>/dev/null | grep -qE "$key_shape"; then
+  # Read the block content through the same fail-closed discipline as the yq call
+  # above: if jq cannot render it, that is a guard that could not check, not a file
+  # with nothing to find. Discarding jq's status here would let the one integrity
+  # check in this file end in the silent pass the whole script exists to prevent.
+  if ! block_text=$(printf '%s' "$blocks" | jq -r '.[]' 2>&1); then
+    die "could not read block scalars from $file: $block_text"
+  fi
+  if printf '%s' "$block_text" | grep -qE "$key_shape"; then
     die "$file: a checkov.io/skipN directive appears inside a block scalar (such as a kustomize 'patch: |'), where the parser cannot reach it — so this file cannot be checked"
   fi
 done <<EOF
--- a/scripts/guard-checkov-skip-reasons.sh
+++ b/scripts/guard-checkov-skip-reasons.sh
@@ -85,7 +85,7 @@ report() { # <file> <key> <problem> <detail>
 # the checkov.io/skip2 rationale"), which is why the tree's own cross-references do
 # not inflate the count.
 sq="'"
-key_shape="(^|[[:space:]]|\"|$sq)checkov\.io/skip[0-9]+[\"$sq]?[[:space:]]*:"
+key_shape="(^|[^A-Za-z0-9_./-])checkov\.io/skip[0-9]+[\"$sq]?[[:space:]]*:"
 
 while IFS= read -r file; do
   [ -n "$file" ] || continue
--- a/scripts/guard-checkov-skip-reasons.sh
+++ b/scripts/guard-checkov-skip-reasons.sh
@@ -1,0 +1,141 @@
+#!/usr/bin/env bash
+#
+# Fail when a `checkov.io/skipN` annotation carries no readable reason.
+#
+# `.mega-linter.yml` states the rule this protects: "No suppression lands without
+# its stated reason being checked against the manifest it describes." Two shapes
+# defeat that rule without failing anything, because checkov reads only the check
+# id, which sits before the damage:
+#
+#   1. TRUNCATION. In a YAML plain (unquoted) scalar, whitespace followed by `#`
+#      opens a comment. So
+#          checkov.io/skip2: CKV_K8S_40=deferred to #3202 -- <the whole reason>
+#      parses as `CKV_K8S_40=deferred to`. The suppression still works and the
+#      stated reason is simply gone. Quoting the scalar is the fix.
+#   2. AN ABSENT REASON. `CKV_K8S_40=` (or a bare id with no `=`) suppresses just
+#      as effectively while asserting nothing.
+#
+# Both leave a bare suppression that reads as reviewed. Nothing else in CI can see
+# either one, which is why this guard exists.
+#
+# Usage: guard-checkov-skip-reasons.sh [root]        (default: k8s)
+# Exit:  0 every annotation carries an intact reason
+#        1 at least one does not (each is named)
+#        2 the guard could not check — missing tool, unreadable root, parse failure
+#
+# Exit 2 is deliberately distinct from both. A guard that cannot run must never be
+# indistinguishable from a clean tree: that is the failure mode where CI goes green
+# having checked nothing.
+
+set -uo pipefail
+
+root=${1:-k8s}
+
+die() { printf 'guard-checkov-skip-reasons: %s\n' "$1" >&2; exit 2; }
+
+command -v yq >/dev/null 2>&1 || die 'yq is required but not on PATH'
+command -v jq >/dev/null 2>&1 || die 'jq is required but not on PATH'
+[ -d "$root" ] || die "not a directory: $root"
+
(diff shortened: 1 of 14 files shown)


devantler (commented):

**Reviewed commit:** `41f2936a4d0278d67022cc205a7578d2ae7e057f`

| Lane | What happened at this head | When |
|---|---|---|
| CodeRabbit | requested; head status reached `Review rate limited` and no review artifact was ever produced at this commit (the only review object sits at the earlier `88596736d3`) | 20:24:15Z |
| Codex | requested; replied that the code-review **usage limit** has been reached — no review ran, and no retry window is stated | 21:07:05Z |
| Cursor Bugbot | requested with a bare trigger; check-run returned `neutral` with `output.title: Error`, zero inline comments, and a `cursor[bot]` comment reporting a **usage limit** | 21:10:02Z |

Two of the three are hard usage limits only an admin can lift, so waiting is not a strategy. This
round is the contract's fallback, held to the same bar as a lane review — not a lighter one.

### What I actually checked

**The design decision is right, and it is the part most worth affirming.** Detecting truncation from
source text is genuinely impossible: `reason  # trailing note` and `reason  # rest of the reason` are
lexically identical, so any source-matching guard must mis-handle one. Requiring an explicitly quoted
scalar *removes* the failure instead of detecting it, and reading yq's node `style` makes that
structural rather than lexical. Four successive false-positive classes were patched out of the old
counter before it was deleted; deleting it was the correct call.

**Shell correctness — the counter-loss traps are all avoided.** `findings`/`checked` are incremented
inside `while … done < <(…)` (process substitution) and the outer loop uses a heredoc, so both run in
the current shell and the counts survive; a pipe would have silently discarded them. `die`'s `exit 2`
therefore also aborts the run for real rather than exiting a subshell. `grep`'s exit status is
correctly split at `>1` so "no matches" stays a clean tree. `-I=0` plus `jq -c '.[]'` handles
multi-document files, and both loops receive a trailing newline so the final entry is not dropped.

**Fail-closed behaviour is pinned, not assumed.** The suite's 24 cases cover all three exit
directions — including a merge key (which would otherwise abort the parse), prose mentioning the
annotation, flow-map placement in three positions, look-alike keys, and a flow-form directive hidden
in a block scalar. A guard that had only ever seen passing input would prove nothing about what it
rejects; this one pins reject and cannot-check too.

**CI wiring verified, because a test the pipeline never invokes guards nothing.** The step runs the
guard over `k8s` *and* the suite, and both `scripts/guard-checkov-skip-reasons.sh` and its test are in
the change filter, so editing either half still runs the check that covers it.

**Verdict: no P0/P1 findings.** One nit inline, which I do not consider blocking and explain why.

scripts/guard-checkov-skip-reasons.sh
```diff
@@ -0,0 +1,175 @@
...
+  # here for exactly the reason the raw pass was removed: a quoted value that
+  # merely MENTIONS the directive
+  #     app.example/note: "See checkov.io/skip1: in the security guide."
+  # is textually identical to a key, and only the parser can tell them apart.
+  if ! blocks=$(yq -o=json -I=0 \
+    '[.. | select(kind == "scalar") | select(style == "literal" or style == "folded")]' \
+    -- "$file" 2>&1); then
+    die "could not scan $file for block scalars: $blocks"
+  fi
+  if printf '%s' "$blocks" | jq -r '.[]' 2>/dev/null | grep -qE "$key_shape"; then
```
devantler: nit — this is the one place the script's own principle ("a guard that cannot run must never be
indistinguishable from a clean tree") is not applied to itself.

If `jq` fails here, `2>/dev/null` swallows the error, the pipeline yields no output, `grep` matches
nothing, and the file is treated as **checkable and clean** — the exact silent-pass shape that
`exit 2` exists to prevent everywhere else in this file.

I am calling it a nit rather than a finding because it is close to unreachable in practice: `$blocks`
is the output of the immediately preceding `yq` call, which is already error-checked at line 157, so
`jq` is parsing JSON that yq just produced. The suppression is defensive rather than load-bearing.

If you want the invariant to hold uniformly, capturing the output and checking the status (as the
`yq` call above already does) would close it without changing any passing behaviour.

devantler: > 🤖 Generated by the Agentic Engineer

**Fixed rather than waved through** — `5f5eccf`.

On reflection the improbability argued for fixing it, not for leaving it: this is the one integrity
check in a file whose entire thesis is that a guard which cannot run must never look like a clean
tree, so an inconsistency there is worth four lines. jq's status is now captured exactly as the `yq`
call above already does it.

Pinned rather than asserted, with the suite's own control discipline:

- The new case stubs jq to fail **only** the block-scan invocation (`-r` with `.[]`) and delegates
  every other call to the real jq, so the fixture still reaches that line with the per-annotation
  checks working normally. Without that isolation a broken jq would abort earlier and prove nothing
  about the scan.
- Run against the **previous** guard the new assertion fails (`expected exit 2, got 0`) — which both
  makes it live coverage and confirms the old path really did pass silently.
- A second assertion checks the stub was actually reached, so exit 2 cannot be satisfied for the
  wrong reason.

26 assertions pass; the guard still reports 40 annotations clean over `k8s`; shellcheck clean.

devantler (commented):

## Self-review (fallback — CodeRabbit, Codex and Cursor Bugbot all quota-blocked)

**Reviewed commit:** `ddde214f7769c6fe4dab0429dc8b9e02cf21bf3d`

| Lane | Evidence | Head | When |
|---|---|---|---|
| CodeRabbit | head status `Review rate limited`; **no review object at this head** | `ddde214f` (this head) | 21:32:37Z |
| Codex | reply reporting the code-review **usage limit** reached — a weekly quota, no retry window stated | `41f2936a` | 21:07:05Z |
| Cursor Bugbot | check-run `neutral` + `output.title: Error`, zero inline comments, `cursor[bot]`: **usage limit reached** — a spend limit only an admin can lift | `41f2936a` | 21:10:02Z |

**On the two lanes evidenced at the previous head.** I did not re-spend them here. Both reported hard
usage limits 22 minutes ago on weekly/monthly windows that cannot have recovered, and re-requesting
would post two more comments that are structurally incapable of succeeding while consuming the very
quotas the lane order exists to protect. Their evidence is a direct per-PR check on this PR, not an
aggregate digest — but it is one head old, and that is stated rather than smoothed over.

CodeRabbit's outcome at this head is worth naming precisely, because the two failures are different:
my first request this round was **prose-heavy and got parsed as chat** (analysis chain + learnings,
no verdict, head status unchanged at the never-reviewed default). The minimal re-trigger then drew
the genuine `Review rate limited`. Only the second is a quota fact.

### The delta since my previous round at `41f2936a`

Two commits: my own fail-closed fix (guard +9, test +47), and `ddde214f` — a MegaLinter auto-fix that
is **`shfmt` cosmetics only** (`: > "$f"` → `: >"$f"` in two redirections). It correctly left the
generated stub's heredoc body alone, and I re-ran the suite at its head rather than assuming: 26/26.

**I am reviewing my own change here, which is a weaker position than an independent lane**, so the
claims below are ones I re-derived by execution rather than by reading:

- The new assertion **fails against the previous guard** (`expected exit 2, got 0`) and passes against
  this one — so it is live coverage, and it independently confirms the old path really did report a
  clean verdict on a file it had never checked.
- Its stub is isolated to the block-scan invocation alone (`-r` with `.[]`). The guard's other jq
  calls are `-c '.[]'` and `-r '.key'`/`.value`/`.style`, none of which match, so the fixture still
  reaches the scan with the per-annotation checks working normally.
- A second assertion requires the stub to have actually fired, so exit 2 cannot be satisfied for the
  wrong reason. Both the RED and GREEN runs show it firing exactly once.
- `exec "$real_jq" "$@"` delegates by absolute path, so the stub cannot recurse into itself.
- Guard over `k8s`: 40 annotations, all clean, exit 0. shellcheck: clean. CI at this head: **0 failing**.

**Exercised as a user, not just reasoned.** Given a genuinely truncated reason —
`checkov.io/skip1: CKV_K8S_40=deferred to #3202 -- the image ships a fixed UID we cannot override` —
YAML parses the value as `CKV_K8S_40=deferred to`, the suppression still works, and the whole
justification is gone. The guard rejects it (exit 1), names the file and key, and states the fix
("wrap the value in double quotes") rather than only reporting a violation.

Issue #3024, opened 2026-08-08
security(kubescape): a framework-load failure must fail loudly, not read as a zero-finding clean result


devantler:
> 🤖 Generated by the Agentic Engineer

## Problem

When the posture scanner cannot load its framework, it aborts and persists nothing — and the surfaces that consume its output cannot tell that apart from a healthy cluster with no findings. **A broken scanner and a compliant cluster read identically.**

That is what made #2933 silent for days: the CI posture gate stayed green throughout, and the stored scan data simply stopped moving. Repairing the immediate cause (#3023) restores scanning, but it does not make the next occurrence visible.

This criterion is called out on #2933 as the durable half, worth more than the immediate repair.

## Expected behaviour

A scan run that fails to load its framework surfaces as a **failure**, not as a zero-finding clean result — loudly enough that it is noticed without anyone going to look.

## Direction

Worth weighing, roughly increasing cost:

1. **Alert on scan staleness.** The cheapest reliable signal — the stored scan data has an age, and an age past roughly one scan interval means the scanner is not working, whatever the reason. Catches framework-load failures and every other abort cause with one rule.
2. **Alert on the scanner's own failure log/metric**, if it exposes one. More precise about cause, but only covers the causes it knows to report.
3. **Fail closed on emptiness.** Any consumer treating "zero findings" as a pass should require positive evidence that a scan ran — this is the same shape as the CI gate's existing liveness problem.

Option 1 gets most of the value and is the suggested starting point. Note the platform already has an alerting path (Alertmanager runs in the `kubescape` namespace).

## Acceptance criteria

- [ ] A framework-load failure produces a visible signal rather than a silent zero
- [ ] The signal fires on scan **staleness** too, so any abort cause is covered, not just this one
- [ ] Proven by exercising the failure — not by reasoning that the rule would fire
- [ ] No consumer treats an empty/zero posture result as a pass without evidence a scan actually ran

Part of #2933.

PR #3238, opened 2026-08-19
security(kubescape): size C-0211's residual population and record the decision

devantler:

### Motivation

A cluster-wide security exception suppresses a Kubernetes hardening control across the whole cluster,
and the note explaining how much it was actually hiding pointed at a closed issue that never answered
the question. Nobody could tell whether the exception was covering a handful of workloads or all of
them, which is the state the posture programme exists to prevent — suppression looks exactly like
compliance.

### What

Measures it, and records the answer where the exception itself lives. It is hiding failures on **every
workload** in the four affected areas — 33 of 33 — so the exception stays as it is for now rather than
being narrowed, which would drop the score without fixing anything. Also corrects two things the note
got wrong: the control checks considerably more than was assumed, and the two workloads it named as
the main concern are not affected at all.

No behaviour change — the exception's own settings are untouched.

Part of #3217

### DIFF
--- a/k8s/bases/infrastructure/cluster-security-exceptions/pod-security-mutations-unscoped.yaml
+++ b/k8s/bases/infrastructure/cluster-security-exceptions/pod-security-mutations-unscoped.yaml
@@ -17,11 +17,38 @@
 # lowers the score to absorb a gap instead of closing it. Setting runAsGroup on
 # those six is tracked in #3223; C-0013 narrows once that lands.
 #
-# C-0211 — its field set differs (fsGroupChangePolicy, runAsUser, runAsGroup), no
-# other exception covers it in the excluded namespaces, and its residual population
-# has never been sized. The name-scoped kubescape-privileged.yaml does not list it,
-# so node-agent and storage sit inside that population rather than excepted out of
-# it. Sizing it needs its own per-workload pass, tracked in #3217.
+# C-0211 — CIS-5.7.3, and it is a twelve-rule control rather than the three-field
+# check its name suggests. The rules this exception suppresses include
+# non-root-containers, set-fsgroup-value, set-fsgroupchangepolicy-value and
+# set-seLinuxOptions, alongside rules C-0016 and C-0055 also carry.
+#
+# Its residual population is every workload in the four excluded namespaces that are
+# actually scanned — flux-system, longhorn-system, observability and velero — all 33
+# of them: 17 Deployments, 7 DaemonSets, 6 CronJobs and 3 StatefulSets. (CronJobs are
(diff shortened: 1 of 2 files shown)


devantler (commented):

**Reviewed commit:** `75ae985db66dc4cda6a2ca6502e66e9c0ffe82a3`

All three external lanes were tried at this head, in priority order, and each refused with a quota-class failure. Per *Local review round*, a provider's billing state is its constraint, not a judgement about this change — so this substitutes for the bot review rather than waiting it out.

| Lane | Result at this head | When |
|---|---|---|
| CodeRabbit | head status `success — Review rate limited`; no review object produced | 07:19:11Z |
| Codex | comment: "reached your Codex usage limits for code reviews" | 07:20:16Z |
| Cursor Bugbot | check-run `completed/neutral`, `output.title: Error` — `Bugbot couldn't run - usage limit reached` (the lane-unavailable shape, zero inline comments) | 07:22:38Z |

### What I checked

The change is documentation and recorded evidence: comments on the exception CR, its `spec.reason`, and the regenerated Headlamp mirror. The exception's behaviour is unchanged — both controls remain `action: ignore` and cluster-wide — so there is no new suppression and no widening of what is excepted.

- **The invariant this PR turns on holds.** Extracted `spec.reason` from the CR and the C-0211 `reason` from the mirror's `exceptionPolicies` independently of the generator, normalised whitespace, and diffed: byte-for-byte identical. This is the desync that `TestCommittedMirrorIsUpToDate` exists to catch, checked without relying on that test's own code path.
- **Arithmetic.** 17 Deployments + 7 DaemonSets + 6 CronJobs + 3 StatefulSets = 33, matching the claimed population.
- **The subset claim is sound.** Two gaps are stated as universal (33/33), so the smaller counts (20 / 18 / 12) are necessarily subsets of them.
- **The mirror was generated, not hand-edited** — regenerated with the command the failing test names; the whole package now passes.
- **Scope discipline.** The counts are explicitly marked prod-only and dated, with an instruction to re-measure rather than inherit — which matters because the local overlay uses a different namespace set.
- **It does not overstate its own evidence.** The comment says plainly that the config scan is not currently producing (#2933, #3024), so this population is real but presently invisible to the posture score. Recording a measured population while naming the reason it cannot yet be seen is the honest form.

### Not findings, recorded for the reader

- `unscanned is not the same as passing` (the `kubescape` namespace exclusion) is the load-bearing distinction here, and it is stated explicitly rather than left implicit. The 384 cluster-scoped scan objects in that namespace are called out as a trap, which is worth keeping.
- The C-0211 residual population stays suppressed. That is the correct call on this evidence: narrowing it now would surface 33 workloads with no fix behind them, which lowers the score to absorb a gap instead of closing it.

PR #3246, opened 2026-08-19
security(vault-backup): make the raft snapshot group-readable

devantler:

### Motivation

Our nightly OpenBao backup writes its snapshot so that **only its owner can read it**. The job that
copies that snapshot off-cluster can therefore only do its work by running as the *same* low user
account — which is the sole reason all three backup workloads are still pinned to a low, host-level
user ID that our security scanner flags.

### Description

Make the snapshot readable by its **group** as well as its owner. The copy job already shares that
group, so it no longer has to impersonate the owner to read the file — which frees those workloads
to move to safe high user IDs in the follow-up step.

Deliberately shipped on its own: the group-readable snapshot has to prove itself on a real nightly
run (and the off-cluster copy has to keep working) before the user IDs move, since that second step
also touches the disaster-recovery restore path.

No behaviour changes today — the copy job still reads the file exactly as before, just no longer
*only* as its owner.

Part of #3202

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -124,6 +124,10 @@ jobs:
               - 'scripts/tests/test-kyverno-umami-mutation-rbac.sh'
               - 'scripts/tests/test-tenant-route-hostname-boundary.sh'
               - 'scripts/tests/test-umami-provisioning-bootstrap.sh'
+              # The snapshot manifests and this test are one contract: the
+              # mirror may only stop depending on OWNING the snapshot while
+              # both writers keep making it group-readable (#3202).
+              - 'scripts/tests/test-vault-snapshot-group-readable.sh'
               - 'scripts/tests/test-kyverno-admission-vpa.sh'
               - 'scripts/tests/kyverno-admission-vpa-rules.yaml'
               # The image-verifier liveness checker and its test are one
@@ -487,6 +491,12 @@ jobs:
         run: |
           shellcheck scripts/tests/test-umami-provisioning-bootstrap.sh
           bash scripts/tests/test-umami-provisioning-bootstrap.sh
+
+      - name: 🔐 Validate vault snapshots stay group-readable
+        if: needs.changes.outputs.k8s == 'true'
+        run: |
+          shellcheck scripts/tests/test-vault-snapshot-group-readable.sh
+          bash scripts/tests/test-vault-snapshot-group-readable.sh
 
       - name: ⚖️ Validate cluster policies
         if: needs.changes.outputs.k8s == 'true'
--- a/k8s/bases/infrastructure/vault-backup/cron-job.yaml
+++ b/k8s/bases/infrastructure/vault-backup/cron-job.yaml
@@ -44,7 +44,7 @@ metadata:
     # is a host UID. That residual is what #3202 removes: it makes the
     # snapshot group-readable first, then raises the UIDs, which the three
     # writers sharing this one RWO PVC have to do together.
-    checkov.io/skip2: "CKV_K8S_40=deferred to #3202 -- the snapshot container's 100 is the openbao image's baked `openbao:x:100:1000` user, and the mirror container must match it to read the -rw------- snapshot from the shared RWO PVC"
+    checkov.io/skip2: "CKV_K8S_40=deferred to #3202 -- the snapshot container's 100 is the openbao image's baked `openbao:x:100:1000` user, and the snapshot is chmod 0640 so the mirror reads it on the shared group 1000 -- only the UID migration itself is outstanding"
 spec:
   schedule: "30 3 * * *"
   # Pin the cron evaluation to UTC instead of inheriting kube-controller-manager's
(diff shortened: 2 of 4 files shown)


devantler (commented):

**Reviewed commit:** `fbc3559a3def8186389972f30d338efd5463856f`

All three external lanes were tried directly at this PR's head and none will deliver:

- **CodeRabbit** — `Review rate limited` at `0b4b0f7f` (2026-08-19T16:23:50Z): *"you've reached your
  PR review limit... Next review available in: 47 minutes."* A stated window makes it predictable,
  not delivered, and 47 minutes parks this past the end of the run.
- **Codex** — `chatgpt-codex-connector[bot]` at 2026-08-19T16:14:08Z: *"You have reached your Codex
  usage limits for code reviews."* No retry window.
- **Cursor Bugbot** — check-run `Cursor Bugbot` `conclusion=neutral`, `output.title=Error` at
  2026-08-19T16:18:37Z, with `cursor[bot]`: *"Bugbot couldn't run - usage limit reached."* That is the
  run-failed shape, not a findings shape. No retry window; only an admin can lift it.

### What I checked

The change makes both vault-snapshot writers `chmod 0640` the snapshot so the off-cluster mirror can
read it as GROUP rather than needing to be its OWNER, which is what pins the writers to one low host
UID (CKV_K8S_40, #3202). I verified the mechanism argument holds — `umask` can only clear permission
bits, so it cannot widen the `0600` that `bao` creates the snapshot with; an explicit `chmod` is
correct under either hypothesis. Both writers (`job.yaml` init container and `cron-job.yaml`) are
covered, and world access stays at `0`.

**One finding was raised and fixed during this review, at the previous head `0b4b0f7f`:**

- **P1 — the chmod side of the binding was unbound.** The guard required exactly one
  `raft snapshot save` and bound the chmod operand to it, but read only the **first** chmod line. A
  later, narrowing `chmod 0600 "$SNAP"` therefore left the snapshot unreadable to the mirror while
  the test still passed — the same unbound-assertion trap the save operand is explicitly guarded
  against, one level down. **Verified by ablation rather than asserted:** with a second narrowing
  chmod injected, the test exited `0` before the fix and exits `1` after it
  (`expected exactly 1 chmod in the snapshot script, found 2`), while the unmodified manifests still
  pass. Fixed in `fbc3559a`.

### Verdict

`Verdict: no P0/P1 findings` at `fbc3559a` — the finding above is fixed at this head, the positive
control still passes, the ablation now fails distinctly, and `shellcheck -S info` is clean.

devantler (commented):

**Reviewed commit:** `3f87488d20e55b67d70c47a2ea793d817706e48f`

All three external lanes were tried **directly at this head** and none will deliver:

- **CodeRabbit** — requested 17:29:0xZ; the head's `CodeRabbit` status returned
  `Review rate limited` within ~40s, with no retry window stated. It had already spent its budget on
  this PR's previous head `c38defa4` at 17:15Z (roughly one review per hour).
  `cr:no-gate@3f87488d`.
- **Codex** — `chatgpt-codex-connector[bot]` at 17:30:44Z: *"You have reached your Codex usage limits
  for code reviews."* No retry window. `codex:no-gate@3f87488d`.
- **Cursor Bugbot** — check-run `Cursor Bugbot` `conclusion=neutral`, `output.title=Error` at
  17:31:2xZ, with `cursor[bot]`: *"Bugbot couldn't run - usage limit reached."* That is the
  **run-failed** shape (zero inline comments, no review object), not a findings shape. No retry
  window; only an admin can lift it. `bugbot:no-gate@3f87488d`.

This is the third time in two days all three lanes have been simultaneously unavailable, so it is a
reachable steady state rather than a coincidence — the spend limits on two of them are worth raising.

## What I reviewed

The change makes both vault-snapshot writers `chmod 0640` the snapshot so the off-cluster mirror can
read it as GROUP rather than by being its OWNER — the coupling that pins the writers to one low host
UID (CKV_K8S_40, #3202). The mechanism argument holds: `umask` can only clear permission bits, so it
cannot widen the `0600` `bao` creates the file with; an explicit `chmod` is correct under either
hypothesis. Both writers are covered, and world access stays `0`.

**Diff scope** — 4 files, `+134/-2`: the two manifests, the new contract test, and its `ci.yaml`
registration (path filter + a `shellcheck`-then-run step). I checked that the `ci.yaml` hunk is this
PR's own and not a duplicate of #3245's, which reached `main` through the branch update: each test
appears exactly the expected number of times in the file.

**Findings raised and fixed during review, across two heads:**

1. **P1 — the chmod side of the binding was unbound** (fixed at `fbc3559a`). The guard required
   exactly one `raft snapshot save` and bound the chmod operand to it, but read only the *first*
   chmod line, so a later narrowing `chmod 0600` would re-close the file with the test still passing.
2. **P1 — a comment counted as a command** (fixed at `3f87488d`). The two bindings were asymmetric:
   `chmod` was matched line-anchored while `raft snapshot save` was an unanchored substring, so a
   comment merely *mentioning* the phrase failed a valid manifest with exactly one real save. Both
   sides now read a comment-free view.
3. **Minor — stale rationale** (CodeRabbit, fixed at `3f87488d`). The `CKV_K8S_40` skip reasons still
   justified the exception by the mirror needing to match the snapshot's owner UID — a constraint
   this change removes. Corrected in both manifests, along with the `/etc/passwd` aside in the test
   header, since the kernel compares numeric UID and GID.

**Control matrix at this head** — each ablation asserted to fail for its *own* reason, with `yq`
confirming the manifest still parses, because my first attempt at (2) failed on a YAML error and that
proves nothing:

| # | Case | Expect | Result |
|---|---|---|---|
| 1 | unmodified | pass | `RC=0` |
| 2 | comment mentioning the save phrase | pass | `RC=0` |
| 3 | **real** second save command | fail | `RC=1` — `binding is ambiguous` |
| 4 | second narrowing `chmod 0600` | fail | `RC=1` — `final mode is ambiguous` |
| 5 | restored | pass | `RC=0` |

3 and 4 are the load-bearing ones: widening the match must not have weakened either guard, and both
still fire. `shellcheck -S info` clean. CI green at this head (18 success / 11 skipped / 0 failed),
including `CodeQL success — No new alerts`.

**What I could not verify here.** The test pins the manifests' *contract*, not the running
container's filesystem: it cannot show that `bao` in the real image honours the chmod, or that the
mirror's GID is 1000 at runtime. Both are asserted from the manifests' `runAsGroup`/`fsGroup` and
from the previously observed `0600` default. The remaining UID migration is #3202's, not this PR's.

### Verdict

`Verdict: no P0/P1 findings` at `3f87488d20e55b67d70c47a2ea793d817706e48f` — all three findings above
are fixed at this head, with the ablations proving the guards still fire.

PR #3257, opened 2026-08-19
security(k8s): disposition trivy KSV-0020 on the vault-config Job, guarded per container

devantler:

### Why

Two scanners were disagreeing about one decision we had already reviewed. checkov accepts that the OpenBao containers run as the user their own image defines; trivy kept reporting it. That difference was holding up the goal of making both scanners block the build again.

Accepting it in trivy carries a risk worth naming: trivy can only silence a check for a whole **file**, while checkov silences it for a single **workload**. So the entry also covers containers whose low user ID would have no justification — and if one were added later, nothing would have gone red. This adds a check that fails the build in exactly that case.

### What

Records the OpenBao user-ID finding as reviewed, and adds a guard that keeps it honest: the build now fails if any other container in that job drops to a low user ID, or if the job's safe default changes. Remaining scanner findings drop 15 → 13, with the highest-severity count unchanged at zero.

The rest of the backlog is now fully accounted for: the OpenBao **backup** workloads (#3202) and one local-only DNS finding. The scanner stays non-blocking until those land.

Part of #2787

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -149,6 +149,9 @@ jobs:
               - 'scripts/megalinter-scan-counts.sh'
               - 'scripts/tests/test-megalinter-scan-counts-ignorefile.sh'
               - 'scripts/tests/test-trivyignore-vendored-operator-boundary.sh'
+              # The vault-config KSV-0020 disposition is path-scoped but the identity it rests
+              # on is PER CONTAINER, so the entry and its premises test move together.
+              - 'scripts/tests/test-trivyignore-vault-config-identity-boundary.sh'
               # The first-party RBAC dispositions and the roles they describe are one
               # contract: the disposition is only valid while the role stays bound and
               # scoped the way it was reviewed. The roles themselves are already covered
@@ -423,6 +426,19 @@ jobs:
         run: |
           shellcheck scripts/tests/test-trivyignore-vendored-operator-boundary.sh
           bash scripts/tests/test-trivyignore-vendored-operator-boundary.sh
+
+      - name: 🧱 Validate vault-config identity disposition boundary
+        if: needs.changes.outputs.k8s == 'true'
+        # The KSV-0020 disposition on the vault-config Job is scoped by PATH, but the reason it
+        # is accepted is per CONTAINER: only the two openbao containers run at the image-baked
+        # uid 100, and the rest take the pod's high 65532 default. trivy cannot express that,
+        # so a container added later at a low UID would be silently suppressed and the count
+        # would simply not rise — which reads exactly like nothing happened. The premise is
+        # asserted per container here; the paired trivy control runs wherever trivy is
+        # installed and skips loudly where it is not.
+        run: |
+          shellcheck scripts/tests/test-trivyignore-vault-config-identity-boundary.sh
+          bash scripts/tests/test-trivyignore-vault-config-identity-boundary.sh
 
       - name: 🧱 Validate first-party RBAC disposition premises
         if: needs.changes.outputs.k8s == 'true'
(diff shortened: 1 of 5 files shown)


devantler (commented):

**Reviewed commit:** `d5d8c89cff2a6231a935f3befba77449380bb547`

### Why this fallback applies — per-lane, all on this PR

| Lane | State | Evidence |
|---|---|---|
| CodeRabbit | **not delivering** — two no-gate outcomes at *this* head | both requests answered with a `🧩 Analysis chain` **chat** reply (comments `5349867796`, `5349901811`); commit status never left `Review skipped: automatic reviews are disabled`. Earlier, at `67125637`, it returned `Review rate limited`. |
| Codex | **not delivering** — usage/billing limit | `You have reached your Codex usage limits for code reviews`, 00:28Z, 47s after its trigger. No retry window; needs account credits. |
| Cursor Bugbot | **not delivering** — usage/billing limit | check-run `Cursor Bugbot` = `neutral` + `output.title: Error`; `cursor[bot]`: `Bugbot couldn't run - usage limit reached`, 00:46Z. Per contract this shape means the review **never ran**. |

I tried the free lane twice at this head, including a re-post in the exact minimal trigger shape that
worked earlier on this PR, before falling back. I am **not** claiming a cause for the chat misparse —
both the prose and minimal forms produced it, so the earlier "prose caused it" reading is refuted and
recorded as unknown.

### What I checked

- **The disposition's premise, empirically.** Inspected the pinned openbao digest directly: config
  `User` is empty, and `/etc/passwd` carries `openbao:x:100:1000::/home/openbao` with `openbao:x:1000`
  in `/etc/group`. Both the prior #2900 finding and this entry's evidence are true — they describe
  different fields. The statement was corrected from *restates* to *defines, not enforces* in
  `67125637` because the original overclaimed.
- **Scoping, behaviourally.** Paired control with the ablation asserted: without the ignorefile
  `job=2 probe=2`; with it `job=0 probe=2`. Identical bytes at a non-dispositioned probe path stay
  reported, so only the path explains the difference.
- **Guard robustness.** Four negative controls fire: non-openbao container at UID 500; pod default
  lowered below the floor; entry stripped of `paths:`; entry widened to an unreviewed path.
- **The reciprocal guard.** `test-trivyignore-vendored-operator-boundary.sh` still passes, accepting
  this as a *reviewed* rather than *widened* skip via the new `reviewed_workload_pairs` array.
- **Effect on the real scan.** `scripts/megalinter-scan-counts.sh --trivy-only`: 15 → 13, KSV-0020
  6 → 4. HIGH/CRITICAL was 0 before and is 0 after.
- **Suite + lint.** All five trivyignore/scan-count tests green; `shellcheck` clean on both changed
  scripts.

### Finding raised and fixed during this round

One real defect, found while preparing this review rather than by any lane:

> **P1 — the new guard could pass vacuously.** Three of its checks read input from
> `done < <(yq ...)`. A failed `yq` feeds the loop nothing, the body never runs, and the check
> reports success having inspected nothing — the exact failure the guard exists to prevent,
> occurring inside the guard, in the dangerous direction (disposition unguarded, CI green). The
> entries check three lines above already guarded with `|| printf '0'`; the loops did not.

Fixed in `d5d8c89c` by capturing each enumeration and status-checking it before its loop, and pinned
by a new negative control: an unparseable `.trivyignore.yaml` now exits 1 with `QUERY FAILED —
refusing to report a boundary this check never inspected`, where it previously passed.

### Residual risk a human should weigh

This disposition does **not** get the "local/CI provider only" premise every other entry in this file
leans on — the Job ships to production. It stands on the image-defined identity alone, which is why
that identity is asserted per container rather than trusted.

**Verdict: no P0/P1 findings at `d5d8c89c`.**

Issue #3071, opened 2026-08-10
The deploy bridge loses its own sync lease mid-run, and each failure fences a node that fails the next deploy


devantler:
## Evidence

Three consecutive prod deploys failed the same way on 2026-08-10, each leaving `prod-control-plane-2` cordoned with a drain fence that failed the next one closed:

| Run | Job window (UTC) | Failure |
|---|---|---|
| `CD` 31416462549 | 18:06:34 → 18:16:15 | `The GHCR synchronization lease was lost; refusing further cluster mutation.` |
| `pr-3066` deploy 31417048057 | 18:16:41 → 18:18:05 | `Could not select Talos nodes requiring GHCR synchronization.` (fails closed on the fence above) |
| `pr-3066` heal 31417048057 | 18:18:51 → 18:31:15 | `The GHCR synchronization lease was lost; refusing further cluster mutation.` |

Each lease-loss run had already cordoned `prod-control-plane-2` and written
`platform.devantler.tech/ghcr-auth-drain-owner`, then refused further mutation — correctly, since it can no longer prove it holds the lease. The fence survives, and the next deploy fails closed on it. That is a self-sustaining loop: **every failure manufactures the precondition for the next failure.**

## What it is not

Ruled out by direct measurement rather than reasoning:

- **Not concurrent deploys.** Job windows do not overlap, and the `prod-deploy` concurrency group is correctly applied to both `cd.yaml` (workflow level) and `ci.yaml`'s `deploy-prod` job. The heal ran alone from 18:18:51 to 18:31:15 and still lost the lease at 18:31:07.
- **Not Flux overwriting the Lease.** `ghcr-auth-refresh` appears in no manifest under `k8s/`, carries no Flux labels or annotations, and its only `managedFields` manager is `kubectl-patch` — the bridge itself.
- **Not a stale-holder takeover.** Takeover of a non-empty holder is refused by design, so nothing could have seized it from a live holder.

## Why it matters

The fail-closed behaviour is right: a process that cannot prove it holds the lease must not keep mutating nodes. The defect is that losing the lease is apparently reachable while running **alone**, and that the resulting fence has no automatic release — so a single occurrence wedges the lane until a human clears it, and the heal that is supposed to restore `main` re-creates the condition instead.

## Where to look

[...]

PR #3344, opened 2026-08-23
fix(deploy): enable orphaned-fence recovery at every deploy call site

### LINKED ISSUE #3343: Orphaned GHCR fence recovery is implemented but never enabled at any call site

## Evidence

The prod merge queue has been wedged since **2026-08-23T14:57Z** (5h+ at time of filing) on an orphaned GHCR synchronization Lease.

`flux-system/ghcr-auth-refresh` is held by `4384f07cc5a864b0-gh32646748943.1-2831-12212`, `renewTime 14:56:33Z`, `leaseDurationSeconds 120` — expired at 14:58:33Z and never released.

The holder is CI run [32646748943](https://github.com/devantler-tech/platform/actions/runs/32646748943). Its `🚀 Deploy to Prod` job **failed at 14:56:37Z**, four seconds after the lease's last heartbeat, without releasing. The `🩹 Heal Prod` job then ran and failed at 14:57:07Z with:

> Another GHCR synchronization transaction holds the synchronization lease; automatic expiry takeover is disabled because Talos writes cannot be fenced.

## The defect

The recovery for exactly this state is **already implemented and already permissioned — it is simply never switched on.**

`.github/actions/deploy-prod` has a `🧯 Recover an orphaned GHCR deploy fence` step running `./scripts/refresh-flux-ghcr-auth.sh --recover-fences`, gated on `if: inputs.recover-orphaned-fence == 'true'`. The input defaults to `"false"`.

**No call site sets it.** All three — `ci.yaml` `deploy-prod`, `ci.yaml` `heal-prod-on-failure`, and `cd.yaml` `deploy-prod` — grant `actions: read` with the comment *"read a prior run attempt status, to prove an orphaned fence holder is dead (deploy-prod recover-orphaned-fence)"*, and none passes the input. In the run above the step is logged `outcome=skipped`.

So the permission, the `GH_TOKEN` wiring, the step, and the script support are all in place for a feature that can never execute.

## Why it is safe to enable

`recover_fences_now()` refuses unless every condition is proven, and releases via a compare-and-swap built from the retained observation:

- refuses while any non-Lease fence is held (ordering);
- refuses while the heartbeat is inside its duration (holder live);
- refuses when the holder carries no run reference (the local-run shape);
- refuses unless the **attempt-pinned** run status is `completed`;
- the release patch is a CAS, so it fails if the Lease moved after the report.

Verified against the current holder: `runs/32646748943/attempts/1` → `status: completed`, `conclusion: failure`, and `ghcr-auth-refresh` is the only stale lease. Every guard would have passed, and the wedge would have cleared automatically.

## Impact

Each affected merge-group deploy fails and is ejected from the queue, and the heal job that exists to restore `main` fails the same way. Recovery currently needs a human with cluster write access. Three merge-ready PRs (#3320, #3340, #3342) are parked behind it.

## Expected behaviour

An orphaned fence whose holder is provably terminal is recovered automatically by the next deploy or heal, without a human.

## Acceptance criteria

- [ ] `recover-orphaned-fence: "true"` is passed at all three call sites.
- [ ] `scripts/validate-merge-group-heal` pins the flag so it cannot silently regress, with a test that fails without it.
- [ ] The composite's refusal paths are unchanged — no guard is relaxed.

Size: S. Part of #3071.

devantler:

### Motivation

A prod deploy that dies while holding the GHCR synchronization Lease wedges the merge queue — and also wedges the heal job that exists to recover from exactly that. Clearing it needs a human with cluster write access.

The recovery for this was already built, and already permissioned: every call site grants `actions: read` specifically so the composite can prove the dead holder is dead. It was simply never switched on, so the recovery step logged `skipped` and could never run. That is what has left the queue wedged since 14:57Z today.

### What this changes

Turns the existing recovery on at all three deploy call sites, and pins it in the workflow contract validator so it cannot quietly revert.

No guard is relaxed. The composite still refuses to release unless no other fence is held, the heartbeat has stopped, and the holder's run attempt is proven terminal through the API — then releases via a compare-and-swap on that same observation. Anything unproven still refuses.

### Merge note — this PR clears the current wedge itself

`ci.yaml` and `cd.yaml` are both inside the `k8s` path filter, so this PR **does** run `deploy-prod` in the merge queue. The merge group builds the speculative merge commit, so that deploy runs *this* PR's workflow — with the flag on. It should therefore recover the orphaned Lease and deploy normally, rather than failing on it the way every deploy since 14:57Z has.

If the recovery declines for any reason, the deploy fails and this PR is evicted with nothing released — the same state as today, no worse.

Fixes #3343
Part of #3071

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -192,3 +192,9 @@ jobs:
           ghcr-token: ${{ secrets.GHCR_TOKEN }}
           hcloud-token: ${{ secrets.HCLOUD_TOKEN }}
           wg-server-private-key: ${{ secrets.WG_SERVER_PRIVATE_KEY }}
+          # Release a synchronization Lease that an already-dead deploy left
+          # held, so one crashed run cannot wedge every later deploy and its own
+          # heal. This relaxes no guard: the composite still refuses unless no
+          # other fence is held, the heartbeat has stopped, and the holder's run
+          # attempt is proven terminal via the API (#3343).
+          recover-orphaned-fence: "true"
--- a/scripts/validate-merge-group-heal/main.go
+++ b/scripts/validate-merge-group-heal/main.go
@@ -34,6 +34,9 @@ const expectedHealCondition = "always() && " +
 // step reachable, so it is pinned rather than left to a reviewer to notice.
 const recoveryOptIn = `          recover-orphaned-fence: "true"`
 
+// The shared composite both prod-deploy paths call.
+const deployCompositePath = "./.github/actions/deploy-prod"
+
 func validateWorkflowContract(workflow string) error {
 	healJob, ok := extractJob(workflow, "heal-prod-on-failure")
 	if !ok {
@@ -56,11 +59,6 @@ func validateWorkflowContract(workflow string) error {
 		{line: "      group: prod-deploy", description: "shared production lock"},
 		{line: "      cancel-in-progress: false", description: "non-preempting production lock"},
 		{line: "          ref: main", description: "current-main checkout"},
-		// Without this, a deploy that dies holding the GHCR synchronization Lease
-		// wedges the queue AND this heal job — the exact state the heal exists to
-		// clear (#3343). Opting in relaxes no guard: the composite still refuses to
-		// release unless the holder is provably terminal and its heartbeat stopped.
-		{line: recoveryOptIn, description: "orphaned-fence recovery"},
 	}
 	for _, requirement := range requirements {
 		if !containsExactLine(healJob, requirement.line) {
@@ -78,15 +76,27 @@ func validateWorkflowContract(workflow string) error {
 		)
 	}
 
-	// The heal job is the last line of defence; the deploy job reaching the same
-	// composite is what stops the wedge arising at all. Pin both, or a dead deploy
-	// still fails the NEXT deploy before any heal gets the chance to clear it.
-	deployJob, ok := extractJob(workflow, "deploy-prod")
-	if !ok {
-		return errors.New("missing deploy-prod job")
-	}
-	if !containsExactLine(deployJob, recoveryOptIn) {
-		return errors.New("deploy job is missing orphaned-fence recovery")
+	// Both jobs reach the same composite, and the opt-in is checked against the
+	// STEP rather than the job: a line accepted anywhere in the job would still
+	// pass after an edit moved it onto an unrelated step, leaving the composite
+	// on its default "false". The validator would then vouch for precisely the
+	// state it exists to catch. The heal job is the last line of defence; the
+	// deploy job reaching the composite is what stops the wedge arising at all.
+	for _, target := range []struct{ key, label string }{
+		{"deploy-prod", "deploy"},
+		{"heal-prod-on-failure", "heal"},
+	} {
+		job, ok := extractJob(workflow, target.key)
+		if !ok {
+			return fmt.Errorf("missing %s job", target.key)
+		}
+		step, ok := extractDeployStep(job)
+		if !ok {
+			return fmt.Errorf("%s job does not reach the shared deploy composite", target.label)
+		}
+		if !containsExactLine(step, recoveryOptIn) {
+			return fmt.Errorf("%s job is missing orphaned-fence recovery", target.label)
+		}
 	}
 
 	return nil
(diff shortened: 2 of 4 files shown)


scripts/validate-merge-group-heal/main.go
```diff
@@ -50,6 +56,11 @@ func validateWorkflowContract(workflow string) error {
 		{line: "      group: prod-deploy", description: "shared production lock"},
 		{line: "      cancel-in-progress: false", description: "non-preempting production lock"},
 		{line: "          ref: main", description: "current-main checkout"},
+		// Without this, a deploy that dies holding the GHCR synchronization Lease
+		// wedges the queue AND this heal job — the exact state the heal exists to
+		// clear (#3343). Opting in relaxes no guard: the composite still refuses to
+		// release unless the holder is provably terminal and its heartbeat stopped.
+		{line: recoveryOptIn, description: "orphaned-fence recovery"},
```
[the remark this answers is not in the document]
devantler: > 🤖 Generated by the Agentic Engineer

Valid, and fixed in `e94216b3`.

The check is now bound to the step that uses `./.github/actions/deploy-prod` rather than to the job. `extractDeployStep` walks back from the `uses:` line to the list-item start, so a step declaring `with:` before `uses:` is handled, and it accepts both `- uses: …` and `uses:` under `- name:`.

Added the negative test you asked for — `heal job relocates the opt-in off the deploy step` moves the line onto the neighbouring checkout step and expects `heal job is missing orphaned-fence recovery`. I also split "does not reach the composite" from "reaches it without the opt-in", since those are different breaks and were previously conflated.

Ablation confirming the new binding is what catches it: reverting just that one call to the job-wide match (kept compiling, so the control is not vacuous) fails **exactly** `heal_job_relocates_the_opt-in_off_the_deploy_step` and nothing else. Restored byte-identical afterwards; full suite green and the real `ci.yaml` still passes.

scripts/validate-merge-group-heal/main.go
```diff
@@ -97,6 +130,50 @@ func extractHealJob(workflow string) (string, bool) {
...
+			target = i
+			break
+		}
+	}
+	if target < 0 {
+		return "", false
+	}
+
+	start := 0
+	for i := target; i >= 0; i-- {
```
devantler, line 153: nit — the walk-back to the list-item start takes the nearest preceding line beginning with `- `, which is right for the two step spellings the comment above names. It would mis-anchor on a step that declares a **sequence-valued input before `uses:`**, e.g.

```yaml
- name: X
  with:
    args:
      - foo
  uses: ./.github/actions/deploy-prod
```

where `- foo` is found first. Worth noting that this fails **closed**: a mis-anchored block will not contain the opt-in line, so the validator errors loudly on the PR rather than vouching for a step it did not check. That is the safe direction, and no such step exists today — so this is a note for whoever next edits the deploy step, not a change request.

.github/workflows/cd.yaml
```diff
@@ -192,3 +192,9 @@ jobs:
           ghcr-token: ${{ secrets.GHCR_TOKEN }}
           hcloud-token: ${{ secrets.HCLOUD_TOKEN }}
           wg-server-private-key: ${{ secrets.WG_SERVER_PRIVATE_KEY }}
+          # Release a synchronization Lease that an already-dead deploy left
+          # held, so one crashed run cannot wedge every later deploy and its own
+          # heal. This relaxes no guard: the composite still refuses unless no
+          # other fence is held, the heartbeat has stopped, and the holder's run
+          # attempt is proven terminal via the API (#3343).
+          recover-orphaned-fence: "true"
```
devantler, line 200: nit — this opt-in is the one of the three that the validator does **not** pin. CI invokes `go run ./scripts/validate-merge-group-heal .github/workflows/ci.yaml`, so only `ci.yaml`'s two call sites are contract-checked; deleting this line from `cd.yaml` would leave the CD deploy path silently back on the composite's `default: "false"` with nothing failing.

Not a change request here — `validateWorkflowContract` requires a `heal-prod-on-failure` job, which `cd.yaml` has no equivalent of, so covering it needs a second entry point rather than a second invocation. That is a larger change than this PR should carry, and the gap predates it. Filing it as a follow-up.

devantler (commented):

**Reviewed commit:** `e94216b3b4705771303af9c95a1e7cf7f5460a71`

**Per-lane state at this head — all three tried, none delivered:**

| Lane | Outcome at this head | When |
|---|---|---|
| CodeRabbit | `Review rate limited` — reply to the 20:49:15Z request, and the head's `CodeRabbit` status reads `Review rate limited`, not `Review completed`. The auto-summary refreshed at 20:49:33Z and names this head, but a head-naming summary carrying a rate-limit marker is a refusal, not a review. | 20:49:29Z |
| Codex | `You have reached your Codex usage limits for code reviews` — no retry window; only the maintainer can lift it. | 20:55:48Z |
| Cursor Bugbot | check-run `neutral` with `output.title: "Error"` (`Bugbot run failed`, 1s), plus `Bugbot couldn't run - usage limit reached`. Zero inline comments, no review object — a failed run, not findings. | 20:57:50Z |

Two of the three are spend/usage limits with no retry window, so waiting would park a finished change indefinitely.

### What I verified, by observation rather than reasoning

**The change enacts.** Traced end-to-end at this exact head: the composite declares `recover-orphaned-fence` with `default: "false"`; its step is gated `if: inputs.recover-orphaned-fence == 'true'` and runs `refresh-flux-ghcr-auth.sh --recover-fences`; the step sits immediately **before** `🔑 Stage Flux and consumer GHCR pull credential`, which is the step that actually fails on a held Lease. All three call sites pass the string `"true"`, matching the `== 'true'` test, and each already grants `actions: read` for the liveness proof.

**All three recovery guards pass against the live cluster right now**, checked independently of the script:

| Guard | Observation | Verdict |
|---|---|---|
| Ordering — no other fence held | exactly **1** `ghcr-auth-refresh` Lease cluster-wide | passes |
| Liveness — heartbeat stopped | `renewTime` 14:56:33Z against `leaseDurationSeconds: 120` — **stale by ~6.1h** | passes |
| Terminal holder | holder `…-gh32646748943.1-…` → run 32646748943 attempt 1 = `completed` / `failure` | passes |

So this change would clear the wedge that is live at this moment, rather than waiting on a human with cluster write access.

**CI at this head:** 29 success, 4 skipped, 0 failing (34 check-runs, matching `total_count`); 0 unresolved review threads.

### Findings

Two non-blocking observations, both filed inline. Neither is a defect introduced by this change.

scripts/validate-merge-group-heal/main.go
devantler: > 🤖 Generated by the Agentic Engineer

**Refuted as a change request; no code change.** The mis-anchor case needs a step that declares a sequence-valued input *before* `uses:`, and no such step exists in either workflow today. It also fails **closed** — a mis-anchored block will not contain the opt-in line, so the validator errors on the PR rather than vouching for an unchecked step. Tightening the walk-back would add parsing surface to guard a shape that has never occurred, in the safe direction. Recorded here for whoever next edits the deploy step.

.github/workflows/cd.yaml
devantler: > 🤖 Generated by the Agentic Engineer

**Deferred to a filed follow-up; no change in this PR.** Tracked as devantler-tech/platform#3345. Covering `cd.yaml` needs a second entry point rather than a second invocation, because `validateWorkflowContract` requires a `heal-prod-on-failure` job that `cd.yaml` has no equivalent of — a larger change than this PR should carry, and the gap predates it.

PR #3656, opened 2026-09-07
fix(policy): generate the CloudNativePG operator's tenant ingress path

### LINKED ISSUE #3655: Tenants with a CloudNativePG database cannot express the operator's ingress path

### Evidence

`restrict-tenant-network-policies` refuses any tenant `fromEndpoints`/`toEndpoints` selector whose key names a namespace (`k8s:io.kubernetes.pod.namespace`). The CloudNativePG operator runs in `cnpg-system` and connects directly to each instance pod on 8000 (instance-manager status) and 5432 (PostgreSQL). A tenant namespace carries the generated `default-deny`, so that ingress is blocked unless something allows it.

A tenant cannot write that allow at all. A selector carrying no namespace key is scoped to the policy's own namespace, so it can never select the operator's pods; and naming the namespace is refused. Measured against live `main` with the tenant template's shipped scaffold, the rule fails `cnp-ingress-sources-must-be-constrained`.

Platform-shipped artifacts are unaffected, because `flux-system:kustomize-controller` is excluded from the policy. Only artifacts reconciled by a tenant's own impersonated ServiceAccount reach this boundary — which is exactly how a real tenant adopts the template.

### Audience and impact

Any tenant adopting `platform-tenant-template` with a database. Its scaffold CiliumNetworkPolicy is refused at admission, leaving two bad options: drop the rule and the operator can no longer manage the cluster, or keep it and the tenant cannot reconcile.

### Expected

The platform provides the cross-namespace path it requires, exactly as it already does for DNS. `add-default-deny` generates `allow-dns` in every namespace precisely because DNS is a platform-required cross-namespace path a tenant may not name. The operator path is the same shape and belongs in the same place.

This keeps the tenant boundary intact — it does not widen what a tenant may name.

### Acceptance criteria

- `add-default-deny` generates an `allow-cnpg-operator` CiliumNetworkPolicy in every tenant namespace: ingress-only, from `cnpg-system`, to CNPG instance pods on 8000/5432.
- The destination is scoped to CNPG instance pods (`cnpg.io/cluster` present), not to every tenant pod.
- `allow-cnpg-operator` joins `default-deny` and `allow-dns` in `cnp-generated-policy-names-are-reserved`, so a tenant cannot patch it into a permissive body.
- Negative control: `restrict-tenant-network-policies` still refuses a tenant-authored selector naming `cnpg-system`, so the boundary is unchanged.

Size: small.

devantler:

### Why

A tenant with a database cannot currently get onto the platform. The scaffold every new tenant starts from ships a network policy the platform then refuses, so the tenant either cannot reconcile at all, or drops the rule and its database stops being managed.

The cause is a genuine gap rather than a mistake on either side. The platform requires tenants to keep their network rules inside their own namespace, which is right — but the CloudNativePG operator lives outside it and has to reach each database pod to run it. There is no wording a tenant can use that satisfies both.

### What

The platform now provides that one path itself, the same way it already provides DNS for every tenant. Tenants keep exactly the boundary they have today — nothing they are allowed to write changes — and the database path simply exists without them having to ask for it.

It is scoped to database pods only, so a tenant without a database gains nothing reachable, and the new policy is protected from being edited into something more permissive.

Fixes #3655


k8s/bases/infrastructure/cluster-policies/best-practices/restrict-tenant-network-policies.yaml
```diff
@@ -110,7 +110,8 @@
 # settle it. A tenant allow-list has no reason to set it either way, so the
 # ambiguity is resolved fail-closed rather than left open.
 #
-# Ownership: the generated `default-deny` and `allow-dns` names are reserved.
+# Ownership: the generated `default-deny`, `allow-dns` and `allow-cnpg-operator`
+# names are reserved.
```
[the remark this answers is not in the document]
devantler: > 🤖 Generated by the Agentic Engineer

Valid, fixed in `492e0ca3`.

Both sites were stale in the way described. The carve-out prose still named the generated floor as
`default-deny`/`allow-dns` only, though the background-controller exclusion now protects
`allow-cnpg-operator` too, and the author-identity fixture header still described two resources
where it now carries three.

One correction to the finding while fixing it: the fixture header did not only undercount, it also
asserted something that is no longer true of all of them. It said both resources use an empty
top-level `endpointSelector` — true of `default-deny` and `allow-dns`, but `allow-cnpg-operator`
selects on `k8s:cnpg.io/cluster`. The header now separates what all three share, the reserved name,
from what only two share.

`test-restrict-tenant-network-policies.sh` passes unchanged, and the change is comments only — no
policy behaviour moves.

k8s/bases/infrastructure/cluster-policies/best-practices/restrict-tenant-network-policies.yaml
```diff
@@ -494,7 +497,8 @@ spec:
         # of its own defaults to Audit and these violations would be ADMITTED.
         failureAction: Enforce
         message: >-
-          `default-deny` and `allow-dns` are generated and re-synchronised by the
+          `default-deny`, `allow-dns` and `allow-cnpg-operator` are generated and
```
devantler, line 500: `nit` The reflow here leaves a very short continuation line (`re-synchronised by the`) before `platform (...)` on the next. It reads awkwardly in the rendered admission message a tenant actually sees. Not worth another CI cycle on its own — worth folding into the next touch of this block.

devantler (commented):

**Reviewed commit:** `492e0ca37e5aef78eaa0ed1553bdde4f3b14f671`

All three review lanes were tried at this exact head and none will deliver:

- **CodeRabbit** — `rate-limited` at 18:54:03Z. Its command-invocation reply carries `⚠️ Action not completed` / `Review rate limited`, and the head status agrees. This round posted its request marker first, so the refusal belongs to this round rather than a spent earlier one.
- **Codex** — `usage-limit` at 18:55:20Z: "You have reached your Codex usage limits for code reviews." No retry window; only the maintainer can lift it.
- **Cursor Bugbot** — `run-failed` at 19:0xZ: check-run `conclusion: neutral` with `output.title: Error`, summary "Bugbot failed to run" after 8s, zero inline comments and no review object.

That is a provider throttle, not a judgement about this change, so the PR is reviewed here rather than parked. Held to the same bar as a bot lane.

### What I checked

**Does it do what it claims?** The fourth generator adds an ingress path from `cnpg-system` to CNPG instance pods on 8000/5432. Its `endpointSelector` requires `k8s:cnpg.io/cluster` (Exists), so a namespace running no database gains no reachable path. Placement of `allow-cnpg-operator` in the reserved-name list is under `cnp-generated-policy-names-are-reserved`, which is the correct rule — a tenant may not author that name, and the background-controller carve-out lets Kyverno create it.

**Does the tenant boundary move?** No. A tenant-authored selector naming `cnpg-system` is still refused; the test asserts that as a negative control. The fixture summary moves `pass: 305, fail: 77` → `pass: 309, fail: 78`, i.e. one more shape *denied*, none newly admitted. Fail counts going up is the safe direction here.

**Does the Crossplane egress contract still hold?** The new rule is generated into every matched namespace including `crossplane-system`, which is why it had to enter the reviewed set rather than be waved through. It grants ingress where that contract governs egress, and selects only pods carrying `cnpg.io/cluster`, which no Crossplane pod does. I also added it to the mutation guard's reviewed-name set — without that, a chart mutating the new policy in `crossplane-system` would not have been refused while one mutating `allow-dns` is.

**Are the guards real?** Ablation, not just a green run: widening the new rule's `endpointSelector` key fails the *contract* assertion (`must remain the exact default-deny and DNS-only contract`), not the inventory one, so the new expected line pins the body rather than merely lengthening a list. Neutralising the tenant policy's key filter reddens the shell test. `test-restrict-tenant-network-policies.sh` and `test-crossplane-egress-policy.sh` both pass; `shellcheck` clean.

**Migration.** `generateExisting: true` means the policy is created in every existing matched namespace. A tenant that already had a CiliumNetworkPolicy literally named `allow-cnpg-operator` would now find it reserved and Kyverno-owned. The name is new to the platform and nothing in-tree uses it, so I judge this a non-issue, but it is the one behaviour here that is not purely additive and it belongs in the record.

### Verdict

`Verdict: no P0/P1 findings` — one `nit` below.

k8s/bases/infrastructure/cluster-policies/best-practices/restrict-tenant-network-policies.yaml
devantler: > 🤖 Generated by the Agentic Engineer

Deferring this one rather than fixing it now, with the reasoning on the record.

It is a cosmetic reflow of a comment-adjacent message string. Fixing it would stale the current-head
review round and require the full lane walk again — and all three providers are down at this head
(CodeRabbit rate-limited, Codex account usage-limited with no window, Bugbot failing to run), so
re-establishing a review would mean re-requesting all three and spending Bugbot's monthly quota a
second time on a line-wrap change.

That trade is not worth it while this PR gates a red default branch: `platform-tenant-template`
`main` is failing `Platform Admissibility`, and `platform-tenant-template#182` cannot go green until
this merges. Folding the reflow into the next touch of this block costs nothing and blocks nothing.

Resolving as deferred, not as invalid — the observation stands.

PR #3684, opened 2026-09-08
fix(talos): collect unused images before storage warns

### LINKED ISSUE #3683: Keep kubelet image GC below the storage warning boundary

## Problem

The kubelet's image garbage-collection trigger is above the platform's storage warning boundary. A node can therefore remain in a warning state while retaining superseded workload images.

## Acceptance criteria

- Configure kubelet image garbage collection declaratively in the shared Talos machine patch layer.
- Start collection below the storage warning boundary and leave enough hysteresis to avoid repeated collection.
- Validate the rendered Talos control-plane and worker configurations.
- Deploy through the normal platform pipeline.
- Verify image data is reclaimed, affected storage returns below the warning threshold, Longhorn remains schedulable, and all volumes remain healthy.
- Keep the setting applied to future static and autoscaled nodes.

devantler:

The kubelet's default image-GC trigger sits above the platform storage warning boundary, allowing superseded workload images to keep a node warning while Longhorn refuses further replica scheduling. This adds a shared Talos patch that starts image collection at 75% and reclaims to 70%, preserving kubelet's five-point hysteresis and applying the guard to existing and future nodes.

Validation:

- RED: the threshold assertion failed because no declarative patch existed.
- GREEN: the policy asserts below the warning boundary with low < high.
- Checksum-verified Talos 1.13.9 rendered and validated the complete patch stack for both control-plane and worker machine configurations.
- Both rendered roles carry the exact 75/70 settings.

Closes #3683

### DIFF
--- a/talos/cluster/gc-unused-images-before-pressure.yaml
+++ b/talos/cluster/gc-unused-images-before-pressure.yaml
@@ -1,0 +1,22 @@
+# Keep container image garbage collection ahead of the platform's storage
+# warning boundary.
+#
+# Kubelet defaults to collecting above 85% image-filesystem use and stopping at
+# 80%. Coroot warns at 80%, and Longhorn can stop scheduling replicas before the
+# kubelet's default trigger because both share /var on the static workers. That
+# leaves a node indefinitely warning and its Longhorn disk unschedulable while
+# months of superseded workload images remain locally cached.
+#
+# Start at 75% and reclaim to 70%. The five-point hysteresis matches kubelet's
+# default gap, while the lower boundary leaves ten percentage points before the
+# Coroot warning and room for Longhorn replica growth. This is a cluster patch,
+# so new static and autoscaled nodes inherit the same protection.
+#
+# No maximum image age is set: recently unused images remain cached while space
+# is healthy, and kubelet selects reclaim candidates only when the shared
+# filesystem crosses this capacity guard.
+machine:
+  kubelet:
+    extraConfig:
+      imageGCHighThresholdPercent: 75
+      imageGCLowThresholdPercent: 70


talos/cluster/gc-unused-images-before-pressure.yaml
```diff
@@ -0,0 +1,22 @@
...
+# default gap, while the lower boundary leaves ten percentage points before the
+# Coroot warning and room for Longhorn replica growth. This is a cluster patch,
+# so new static and autoscaled nodes inherit the same protection.
+#
+# No maximum image age is set: recently unused images remain cached while space
+# is healthy, and kubelet selects reclaim candidates only when the shared
+# filesystem crosses this capacity guard.
+machine:
+  kubelet:
+    extraConfig:
```
devantler, line 20: **P1 — the render check proves the new keys landed, but not that the existing ones survived.**

`talos/cluster/evict-pods-before-oom.yaml` already sets `machine.kubelet.extraConfig`, with `systemReserved.memory` and `kubeReserved.memory`. This patch sets the *same* `machine.kubelet.extraConfig` map. Two cluster patches now write one key.

The PR body's validation says the rendered control-plane and worker configs "carry the exact 75/70 settings". That is a one-directional check: it confirms this patch won, which is exactly what you would also observe if `extraConfig` were replaced wholesale rather than deep-merged. In that case `systemReserved`/`kubeReserved` would be silently dropped from every node — removing the OOM guard that its own patch exists to provide, with no signal at render or validate time.

Talos most likely deep-merges here, so this is probably fine — but "probably" is the wrong standard for a change that could remove a memory guard fleet-wide, and the failure is invisible precisely because the assertion that was run passes either way.

**What would settle it:** re-render the same stack and assert **both directions in one output** — that the rendered `kubelet.extraConfig` contains `imageGCHighThresholdPercent: 75` **and** `imageGCLowThresholdPercent: 70` **and** the pre-existing `systemReserved.memory` / `kubeReserved.memory` values. A negative control makes it airtight: drop this patch, re-render, and confirm the reserved values are unchanged from that baseline.

If the render already shows all four keys coexisting, paste that fragment and this resolves immediately — no code change needed.

devantler (commented):

Reviewed commit: `378c16822e515cc2dd688f54b06afbd7381510a7`

This is a taken-over PR from a sibling lane, reviewed under the local-review-round fallback because all three external lanes are provably not delivering at this exact head:

- **CodeRabbit** — `Review rate limited` at 2026-09-08T19:23:09Z; the free allowance is spent. `cr:no-gate@378c16822e` recorded.
- **Codex** — usage limit reached for code reviews at 2026-09-08T20:03:17Z, stating no retry window (cause class `quota/billing`). `codex:no-gate@378c16822e` recorded.
- **Cursor Bugbot** — check-run `102218969938` returned `neutral` / `Error` ("Bugbot run failed") 2 seconds after starting at 2026-09-08T20:07:48Z, with a usage-limit explanation and no retry window. `bugbot:no-gate@378c16822e` recorded.

### What I verified myself

**The patch is genuinely wired in, not a silent no-op.** This was the main correctness risk, since a Talos patch that nothing consumes validates green while changing nothing. `talos/cluster/` holds 16 sibling patches and no configuration anywhere lists any of them by name, so the directory is consumed by convention rather than an explicit list — a new file in it is picked up. I confirmed the absence of a name-based listing with a working positive control (my first search apparatus returned a false empty and was discarded).

**Threshold arithmetic is right.** `high (75) > low (70)` satisfies kubelet's ordering requirement, the five-point gap matches kubelet's own 85/80 default hysteresis, and 75 sits below the 80% Coroot warning boundary the issue is about.

### Verdict: 1 finding (P0: 0, P1: 1)

The finding is a missing conservation proof rather than a demonstrated defect — see the inline comment.

talos/cluster/gc-unused-images-before-pressure.yaml
devantler: > 🤖 Generated by the Agentic Engineer

**Refuted by render — resolving. `extraConfig` deep-merges; nothing is dropped.**

I raised this and I own settling it, so I produced the render myself rather than leaving the PR parked. Both directions asserted in one output, on the full patch stack at head `378c16822e`, rendered with `talosctl gen config --talos-version v1.13.9` (17 cluster patches + the role patches, filename order).

**Rendered `machine.kubelet.extraConfig` — identical for `controlplane` and `worker`:**

```yaml
evictionHard:
  memory.available: 100Mi
evictionMaxPodGracePeriod: 60
evictionMinimumReclaim:
  memory.available: 200Mi
evictionSoft:
  memory.available: 500Mi
evictionSoftGracePeriod:
  memory.available: 1m30s
imageGCHighThresholdPercent: 75
imageGCLowThresholdPercent: 70
kubeReserved:
  memory: 256Mi
mergeDefaultEvictionSettings: true
systemReserved:
  memory: 256Mi
```

All four keys coexist. The OOM guard survives.

**Negative control** — same stack with this patch dropped, diffed against the above per role:

| assertion | controlplane | worker |
|---|---|---|
| keys added by this patch | exactly `imageGCHighThresholdPercent`, `imageGCLowThresholdPercent` | same |
| keys dropped | none | none |
| shared keys whose value changed | none | none |

**And the conservation check is itself sensitive** — an empty "nothing dropped" is worthless unless the check would have spoken. Ablating the render to omit `evict-pods-before-oom.yaml` (the shape this finding feared) makes it report all 8 keys dropped, `systemReserved` and `kubeReserved` among them. So the empty result above is a real conservation, not a blind check.

Conclusion: Talos strategic-merges `machine.kubelet.extraConfig` across cluster patches, so two patches writing that map compose rather than overwrite. No code change needed and the original one-directional assertion, while incomplete as evidence, described a correct outcome.

One caveat worth stating rather than hiding: this was rendered with a v1.14.0 client targeting `--talos-version v1.13.9`. The generated config targets 1.13.9, but if you want the assertion pinned to the exact CI toolchain, the same two commands under CI's checksum-verified 1.13.9 binary would close that last gap.

Issue #3465, opened 2026-08-29
A constant seLinuxOptions level: s0 would flatten per-Pod MCS isolation if SELinux were ever enforcing


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`add-security-context.yaml` assigns a **constant** `seLinuxOptions.level: s0` to every container it
touches. On prod that is **116 containers** today.

`s0` is the lowest sensitivity level with **no MCS categories**. Under SELinux, per-workload
isolation comes from the *categories* (`s0:c123,c456`), which are normally allocated uniquely per
container by the runtime. A blanket `s0` flattens that: every container shares one label, and
MCS no longer separates them.

## Why this is not currently a live vulnerability

Measured 2026-08-29 on `admin@prod`:

| fact | value |
|---|---|
| node OS | **Talos v1.13.9**, kernel `6.18.44-talos` |
| SELinux | **not the active LSM** (Talos hardens via KSPP/AppArmor) |
| pods carrying pod-level `seLinuxOptions` | **0** |

`runc` ignores SELinux labels when SELinux is not enforcing, so the setting is inert here — there
are no per-Pod MCS categories to flatten, because nothing is allocating any. The policy comment
already records this rationale, and the field is retained as declared CIS-5.7.3 posture.

## The latent risk

The premise "SELinux is not enforcing" is a **property of the cluster, not of the policy**, and
nothing in the repository asserts it. If SELinux were ever enabled — on this cluster, or on any
consumer of these shared `k8s/bases` policies — the constant `s0` would silently become an
isolation regression across the whole fleet at once, and no existing test would notice.

This is a property of the **pre-existing** rules, not of any one recent change: #3460 extends the
same established pattern to 12 more namespaces without introducing a new risk class.

## Expected

Decide, and record the decision, between:

1. **Assert the premise.** Add a check that fails if SELinux is enforcing on any node, so the
   policy's own assumption is pinned rather than assumed.
2. **Stop setting `level` at all** and satisfy C-0211 another way, since on this cluster the field
   is inert and only changes declared posture.
3. **Keep it and document the constraint** explicitly as "these policies require a non-SELinux
   cluster", so a consumer of the shared bases is warned.

## Acceptance criteria

- [ ] A recorded decision among the three, with rationale.
- [ ] If (1): the check exists and fails when SELinux is enforcing.
- [ ] If (2) or (3): the change or the documented constraint lands with the reasoning.

[...]

PR #3691, opened 2026-09-08
fix(metrics-server): declare baseline security defaults

### LINKED ISSUE #3676: security(metrics-server): declare baseline security defaults without changing runtime identity
Part of #3239.

The metrics-server controller template still omits fsGroupChangePolicy and container seLinuxOptions. Add only OnRootMismatch and an empty SELinux object through the existing chart values. Preserve the current user, capabilities, seccomp, volumes, image, two-replica strategy and disruption budget.

Acceptance:

- [ ] Verify chart omitted/enabled/rollback output and Helm field preservation.
- [ ] Preserve rendered identities and authorization.
- [ ] Deploy through the normal GitOps queue.
- [ ] Verify the same controller/release identities, full readiness and working metrics after rollout and later reconciliation.

This closes only metrics-server; the parent measurement remains open.

devantler:

### Motivation

Kubescape scores metrics-server as missing two baseline security declarations. The
cluster-wide mutation that normally supplies them rewrites the *running* pod, while
Kubescape reads the *stored* workload spec — so the finding persists no matter how the
cluster behaves at runtime. Declaring them in the chart values is the only route that
moves both surfaces.

### What

metrics-server now declares `fsGroupChangePolicy: OnRootMismatch` at the pod level and
an empty `seLinuxOptions` on the container. Neither changes what the workload can do:
the first is inert while the deployment mounts only `emptyDir` and sets no `fsGroup`,
and the second keeps the node's per-container SELinux label allocation rather than
pinning one. Everything that defines the runtime identity — the pinned user, dropped
capabilities, read-only root filesystem, seccomp profile, image, volumes, replica count
and disruption budget — is unchanged, which the chart render confirms field by field.

The EKS authorization aggregate moves with any rendered change and is re-approved in
this PR. Its own error set is the conservation proof in both directions: one error
class only, and zero unapproved, missing or duplicate rendered identities, so no
subject, grant, binding or ServiceAccount moved.

Fixes #3676
Part of #3239

### DIFF
--- a/k8s/bases/infrastructure/controllers/metrics-server/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/metrics-server/helm-release.yaml
@@ -19,13 +19,13 @@ spec:
   # https://github.com/kubernetes-sigs/metrics-server/blob/master/charts/metrics-server/values.yaml
   values:
     replicas: ${metrics_server_replicas:=2}
+    # fsGroupChangePolicy and seLinuxOptions are the two non-privilege C-0211
+    # fields the cluster-wide add-security-context mutation cannot supply here:
+    # that mutation rewrites the running POD, while Kubescape scores the STORED
+    # workload spec (#3239). Writing them into the chart values is the only
+    # route that moves both surfaces. They land under two different chart value
+    # keys: podSecurityContext renders pod-level, securityContext container-level.
     podSecurityContext:
-      # fsGroupChangePolicy and seLinuxOptions are the two non-privilege C-0211
-      # fields the cluster-wide add-security-context mutation cannot supply here:
-      # that mutation rewrites the running POD, while Kubescape scores the STORED
-      # workload spec (#3239). Writing them into the chart values is the only
-      # route that moves both surfaces.
-      #
       # Inert today, which is the point: this Deployment mounts only emptyDir and
       # sets no fsGroup, so OnRootMismatch changes no ownership on any volume. It
       # is a declared default that holds if a future chart revision adds one.
--- a/k8s/bases/infrastructure/controllers/metrics-server/helm-release.yaml
+++ b/k8s/bases/infrastructure/controllers/metrics-server/helm-release.yaml
@@ -19,6 +19,26 @@ spec:
   # https://github.com/kubernetes-sigs/metrics-server/blob/master/charts/metrics-server/values.yaml
   values:
     replicas: ${metrics_server_replicas:=2}
+    podSecurityContext:
+      # fsGroupChangePolicy and seLinuxOptions are the two non-privilege C-0211
+      # fields the cluster-wide add-security-context mutation cannot supply here:
+      # that mutation rewrites the running POD, while Kubescape scores the STORED
+      # workload spec (#3239). Writing them into the chart values is the only
+      # route that moves both surfaces.
+      #
+      # Inert today, which is the point: this Deployment mounts only emptyDir and
+      # sets no fsGroup, so OnRootMismatch changes no ownership on any volume. It
+      # is a declared default that holds if a future chart revision adds one.
+      fsGroupChangePolicy: OnRootMismatch
+    securityContext:
+      # Preserve runtime SELinux/MCS label allocation. An empty object satisfies
+      # C-0211, which accepts any non-absent seLinuxOptions, without pinning a
+      # level the node would otherwise allocate per-container (#3465).
+      #
+      # Helm merges this map into the chart's own securityContext, so the pinned
+      # user, dropped capabilities, read-only root filesystem and seccomp profile
+      # are all preserved; the rendered container gains this one field.
+      seLinuxOptions: {}
     podDisruptionBudget:
       enabled: true
       # Stays minAvailable: 1: ksail bootstrap-installs metrics-server (ksail.yaml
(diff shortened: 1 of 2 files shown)


k8s/bases/infrastructure/controllers/metrics-server/helm-release.yaml
```diff
@@ -19,6 +19,26 @@ spec:
   # https://github.com/kubernetes-sigs/metrics-server/blob/master/charts/metrics-server/values.yaml
   values:
     replicas: ${metrics_server_replicas:=2}
+    podSecurityContext:
+      # fsGroupChangePolicy and seLinuxOptions are the two non-privilege C-0211
```
devantler: nit: this comment block sits under `podSecurityContext` but describes *both* fields — "fsGroupChangePolicy and seLinuxOptions are the two non-privilege C-0211 fields" — while `seLinuxOptions` is actually set under the separate `securityContext` key nine lines below. A reader landing here reasonably expects both to be in this block. Splitting the shared rationale (the mutation-rewrites-the-pod-but-Kubescape-scores-the-stored-spec point) from the per-field note, or moving the shared half above `podSecurityContext:`, would put each explanation next to the field it governs. Non-blocking — the rendered result is correct and verified.

devantler (commented):

**Reviewed commit:** `dccb06ee53a463322d2f57e0e6ff41e3292cbca0`

All three external lanes were tried at this head, in priority order, and each was checked directly on its own surface rather than inferred from a digest:

- **CodeRabbit** — 18:01:26Z. Command-invocation reply `⚠️ Action not completed` / `Review rate limited`, plus an auto-summary stating the free OSS review allowance is exhausted. No review object, no findings.
- **Codex** — 18:03:23Z. Issue comment: "You have reached your Codex usage limits for code reviews." No review object, no findings.
- **Cursor Bugbot** — 18:10:00Z. Check-run `Cursor Bugbot` `conclusion: neutral` with `output.title: "Error"` and summary `Bugbot run failed`, plus a `cursor[bot]` comment: "Bugbot couldn't run - usage limit reached." Per the contract's table that is a *failed run*, not findings — zero inline comments, zero review objects, 7s elapsed.

None of the three will deliver a usable review at this head, and all three are quota/billing limits rather than a judgement about this change, so this round substitutes for the external green under *Local review round*.

### What I verified

**The change renders, and renders exactly.** The decisive risk in a Helm-values change is a silently ignored key — Helm does not error on an unknown value, so a wrong key name is a no-op that still moves the fingerprint. Both keys are real in `metrics-server` 3.14.0 (`podSecurityContext` and `securityContext` are declared chart values), and rendering the chart at that pinned version with and without the added values differs by **exactly** two lines:

- pod-level `spec.template.spec.securityContext.fsGroupChangePolicy: OnRootMismatch`
- container-level `seLinuxOptions: {}`

`seLinuxOptions: {}` **survives** into the rendered manifest rather than being pruned as an empty map, which was the specific no-op risk worth checking. Every pre-existing hardening field is preserved by the map merge — `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `readOnlyRootFilesystem: true`, `runAsNonRoot: true`, `runAsUser: 1000`, `seccompProfile.type: RuntimeDefault` — and image, volumes, replicas and the disruption budget are untouched. That satisfies the issue's "Verify chart omitted/enabled/rollback output and Helm field preservation" criterion directly.

**Authorization is conserved, and that is independently checkable right now.** `expectedRenderedSurfaceSHA` is the one constant whose update could launder an authorization change, so it deserves the scrutiny. The validator pins per-identity fingerprints *alongside* the aggregate, and `🔐 Validate Production Authorization` is **green at this head** — so every per-identity pin still matches its original value while only the aggregate moved. No subject, grant, binding or ServiceAccount changed, which is consistent with a Deployment not being a grant-bearing kind.

**Security reading.** No secrets, credentials, permissions, workflow triggers or CI configuration are touched. `seLinuxOptions: {}` is not a weakening: it declares the field without pinning a level, so the runtime keeps allocating MCS labels exactly as before — pinning a level would be the riskier choice, which is what #3465 settled. `fsGroupChangePolicy: OnRootMismatch` is inert here (emptyDir only, no `fsGroup`) and is honestly documented as such.

**Scope.** The two remaining acceptance criteria — deploy through the GitOps queue, and confirm identities/readiness/metrics after rollout — are inherently post-merge, and belong to the 📊 Verifying stage. `Fixes #3676` / `Part of #3239` is the right pairing: the issue itself states it closes only metrics-server while the parent measurement stays open.

### Verdict

**Verdict: no P0/P1 findings.** One `nit` is posted inline.

k8s/bases/infrastructure/controllers/metrics-server/helm-release.yaml
devantler: > 🤖 Generated by the Agentic Engineer

Fixed in `3c0f3b667cdf9c6f7d8aaf23b14f4166787c517b`. The shared rationale (the mutation rewrites the running pod while Kubescape scores the stored spec) now sits above both value keys, and names which key renders where — `podSecurityContext` pod-level, `securityContext` container-level — which is precisely the thing that had to be verified independently. Each per-field note stays with its own field.

Comment-only: the parsed YAML is identical before and after (compared as a JSON projection, which drops comments; a deliberate `OnRootMismatch` → `Always` mutation was detected by the same comparison, so it is not vacuous). The rendered surface, and therefore `expectedRenderedSurfaceSHA`, is unaffected.

Issue #3790, opened 2026-09-14
security(platform): declare CPU limits on the datapath DaemonSets through the rollout gate


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Measured on #3567 (2026-09-14): the kube-system datapath DaemonSets — `cilium`, `cilium-envoy`, `tetragon`, `hcloud-csi-node` — declare no CPU limit in their templates and rely on the kube-system `default-limitrange` (`default.cpu: "2"`) at admission, so on a fresh cluster they can start uncapped.

## Audience and problem

Operators rebuilding prod. Unlike the non-datapath add-ons, changing these templates rolls the node datapath, which this repository only does through the Cilium rollout gate.

## Smallest useful change

Declare `resources.limits.cpu: "2"` (identical to the LimitRange default, so running pods keep the same effective limit) in these charts' values, and land it **only** through the existing rollout-gate machinery (or folded into the next gated Cilium/Tetragon chart bump) — never as a plain rollout.

## Acceptance criteria

- [ ] Each DaemonSet's rendered template carries `limits.cpu`.
- [ ] The change reaches prod through the rollout gate, with the gate's own completion evidence recorded.
- [ ] A static check fails if one of them loses the declaration.

Rough size: M.

Part of #3567.

devantler:
> 🤖 Generated by the Agentic Engineer

## Read this before implementing: the agent limit contradicts a recorded decision

`k8s/bases/infrastructure/controllers/cilium/helm-release.yaml` sets requests on the agent and `cilium-envoy` DaemonSets and states that **limits are intentionally unset**, citing Cilium's performance guidance against capping the agent.

The kube-system `default-limitrange` (`default.cpu: "2"`) overrides that intent: every agent and envoy container admitted after the `bootstrap` layer already carries a 2-core CPU limit on prod. So the live state already contradicts the chart comment, and this issue's plan — declare the same value in the template — would make that contradiction permanent.

Decide one of these before writing any values change:

1. **Keep the cap** (it is compressible and ~10× the agent's 200m request) — declare `limits.cpu: "2"` here **and** update the chart comment so it no longer claims limits are unset.
[...]

PR #3863, opened 2026-09-16
feat(ci): add a dispatch-only datapath CPU throttling measurement

devantler:

### Why

We decided in #3790 to keep the CPU cap on the network datapath (Cilium, Envoy, Tetragon, the storage node plugin), but only if it does not throttle them. Nobody can measure that today: the in-cluster Prometheus does not collect the right counters, and agent access cannot reach them.

### What

A workflow the maintainer can start by hand that reads the throttling counters straight from each node, waits, reads them again, and reports the share of throttled time per container against the 1% bar from #3790. It only reads, prints no node or pod names, and says INCONCLUSIVE instead of guessing if anything moves during the window. Run it once before and once after the limit change.

Part of #3790


devantler:


@coderabbitai review

devantler:

## Readiness at `0489186e2b7cf0c7930d876d8143960094ce511e`

- **Tested:** CI 24/24 green; the new `📏 Validate datapath CPU throttling measurement` step ran in the `🧪 Validate Manifests` job and printed its pass line. Locally the suite also passes under macOS bash 3.2, and each guard was ablated to confirm the test fails without it: the pod-name filter, the namespace filter, the counter-reset check, the node-set check, the stderr discard, and the zero-periods rule in both directions.
- **Reviewed:** CodeRabbit reviewed this head with no actionable comments (summary updated 18:39Z, status `Review completed`), and there are 0 review threads. Its pre-merge docstring-coverage warning is a generic percentage heuristic on shell helpers such as `usage` and `fail`. Neighbouring scripts don't document helpers like these either, so I'm not treating it as a finding.
- **Evaluated as its user:** this is a dispatch-only workflow that can only run from `main` with the prod environment credentials, so it can't be run against the cluster before merge. What I did instead: ran the script end to end against a faked kubelet that serves real cAdvisor exposition lines, and read the output the way an operator would. The table gives per-container ratios and the 1% bar, and prints no node or pod names. Every way the reading could go wrong ends as INCONCLUSIVE (exit 3), never as a number. The script only issues `get` calls, so a wrong assumption about the live endpoint costs one INCONCLUSIVE run and changes nothing in the cluster. The first real run is the before-change baseline #3790 asks for.

Issue #3848, opened 2026-09-15
Redeploy main after a merge when prod has not converged on it


devantler:
> 🤖 Generated by the Agentic Engineer

Part of #2838. It follows #3846, whose resolver merged in #3847.

## Evidence

- Since #3847, `scripts/resolve-prod-convergence` answers from signed evidence whether prod's published artifact matches `main`. The answer is `CONVERGED`, `BEHIND`, `DIVERGED` or `UNKNOWN`.
- Nothing runs it. After the interleaving #2838 describes, prod still stays behind `main` until an unrelated deploy happens to land.
- The resolver needs registry read access to verify the attestation. That exists only on a runner with `packages: read`, so its last acceptance point (a live `CONVERGED` read) is proven here, not in #3846.

## Audience and impact

This serves the platform operator. It closes the last path by which prod can silently disagree with `main` after a merge.

## Smallest useful change

After a merge lands on `main`, run the resolver against the digest `:latest` names:

- `CONVERGED`: do nothing.
- `BEHIND` or `DIVERGED`: run the normal production deploy of `main`, through the same composite and the shared `prod-deploy` lock.
- `UNKNOWN`: do not deploy; fail visibly, so a broken check is never mistaken for a converged prod.

Because this job reaches production, it inherits every gate the existing deploy paths carry: the publication gate, the authorization and signing contracts, and the per-workflow coverage guards. It lands behind an opt-in switch, defaulting off, and is activated separately.

## Acceptance criteria

- [ ] On a merge where prod already matches `main`, the job reads `CONVERGED` and does not deploy (proven on a real merge).
- [ ] A `BEHIND` or `DIVERGED` verdict triggers exactly one deploy of `main`, and a second evaluation afterwards reads `CONVERGED`.
- [ ] `UNKNOWN` never deploys and fails the job with the resolver's reason.
- [ ] The job holds the shared `prod-deploy` lock and cannot run concurrently with a merge-group deploy or heal.
- [ ] #3846's live `CONVERGED` acceptance point is recorded from this job's first real run.

## Size

M: one workflow job plus contract-test updates, behind an opt-in switch.

devantler:
> 🤖 Generated by the Agentic Engineer

[...]

Issue #3846, opened 2026-09-15
Tell whether prod has converged on main from the published artifact's signed provenance


devantler:
> 🤖 Generated by the Agentic Engineer

Part of #2838. Also resolves the race #2828 described, which was closed as its duplicate.

## Evidence

- Every production deploy promotes one digest to `ghcr.io/devantler-tech/platform/manifests:latest`, and that digest carries a signed SLSA provenance attestation that names the commit it was built from.
  - Measured 2026-09-15 on the merge-group deploy for #3845: `latest@sha256:6cb49cc4…` attests commit `eb8b5d0386a8…`, built from `gh-readonly-queue/main/pr-3845-…`.
  - That commit is exactly `main`'s tip after the merge (the compare reads `identical`).
- So the Git state prod is running is already recorded and signed. Nothing reads it.
- #2828's proposed guard, "skip the heal when the published revision is reachable from `main`", cannot close the race. When the older group's heal runs, the newer group has not merged yet, so its artifact is not reachable from `main` either and the two look identical. The gap only becomes decidable after the newer group lands, which is what #2838 describes.

## Audience and impact

This affects the platform operator and every service on the cluster. After that interleaving, prod silently runs an older revision than `main` until some unrelated change triggers another deploy. Nothing reports it, so the platform's GitOps promise ("prod is what `main` says") is quietly false.

## Smallest useful change

Build a read-only resolver. Given `:latest`'s digest, it verifies the provenance against the platform's own deploy workflow identity, takes the source commit, and reports one verdict:

| Verdict | Meaning |
|---|---|
| `CONVERGED` | the commit is `main`'s tip, or an ancestor of it with no deploy-input change since |
| `BEHIND` | an ancestor of `main`, but deploy inputs changed since (the #2838 state) |
| `DIVERGED` | not reachable from `main`: an ejected merge-group artifact (the #2026 state the heal already handles) |
| `UNKNOWN` | any read or verification failure; never reported as `CONVERGED` |

Deploy inputs are the `k8s` path filter in the `changes` job of `.github/workflows/ci.yaml`. The resolver reads them from that file rather than copying the list, so the two cannot drift.

This child is not wired into any workflow and changes nothing in production. The follow-up child uses the verdict to run a convergence deploy after a merge.

## Acceptance criteria

- [ ] A test proves each verdict against a fixture repository, including ablations that fail when the ancestry check or the deploy-input check is removed.
[...]

Issue #2838, opened 2026-07-28
A merge-group heal can leave prod behind a merge that lands just after it


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Raised as a P1 by Codex on [#2825](https://github.com/devantler-tech/platform/pull/2825) and confirmed
as a real system property while reviewing that PR — but it is **not** introduced by that PR, so it is
filed here rather than folded into it.

## The problem

`deploy-prod` and its `heal` job share the `prod-deploy` concurrency lock. A merge group's deploy is
**speculative** — it deploys code that has not merged yet. That admits this interleaving:

1. Merge group A's deploy fails; A is evicted and its heal is queued.
2. Merge group B's deploy runs and succeeds, putting B's (unmerged) content on prod.
3. A's heal acquires the lock, checks out the current tip of `main` — which still predates B — and
   deploys it, overwriting B's deployment.
4. B then merges. Nothing re-deploys it, so prod sits behind `main` until the next deploy.

Step 3 is individually correct: the heal deliberately deploys the last state that actually landed in
Git, and if B were later evicted instead of merged, that overwrite would be exactly right. The defect
is the **absence of a deploy after step 4**, not the heal's choice of ref.

## Why this is not a #2825 regression

Without `queue: max` the same interleaving occurs whenever the heal is not dropped; `queue: max` only
stops the heal being silently discarded, which is the failure that job exists to prevent. The gap is
in the heal contract, and it predates the queue setting.

## Expected behaviour

After a merge group lands, prod converges on `main` without requiring an unrelated later deploy.

## Acceptance criteria

- A merged revision is always deployed, even when a heal overwrote its speculative deploy beforehand.
- The heal keeps deploying `main`'s tip (no reintroduction of `base_sha`).
- Covered by `scripts/validate-merge-group-heal` or an equivalent executable check.

Rough size: S–M. Likely shapes: a post-merge convergence deploy on `push: main`, or a heal that
records the sha it deployed and re-fires when `main` advances past it.

devantler:
> 🤖 Generated by the Agentic Engineer

## Decision: fix this with a convergence check after merge, not in the heal

I selected this as the oldest open bug with nothing parked on it, together with #2828, which describes the same race from the heal's side. I closed #2828 as a duplicate of this issue.

[...]

PR #3864, opened 2026-09-16
feat(ci): report whether prod has converged on main after each push

devantler:

### Why
Production can quietly run an older revision than `main` after a merge (#2838), and nothing reports it. The checker that can tell (#3846) exists but nothing runs it.

### What
After every push to `main`, a read-only job asks that checker whether production matches `main` and puts the answer on the run: a note when it matches, a warning when it does not, and a failure when it cannot tell. It never deploys. Acting on the answer is the next slice of #3848.

Part of #3848


devantler:


@coderabbitai review

devantler:

**Readiness at `ab0d2303`**

1. **Tested.** `test-report-prod-convergence.sh` (15 assertions) ran and passed in this PR's CI (job 104963971895, step "Validate prod convergence report"). It uses stub `docker`/`gh`/`git`/`go`. I removed each guard in turn (the verdict/exit pairing, the digest shape, stderr kept out of the digest, the skip for commits already present, and the check on the attested commit's format), and each removal failed exactly the case that covers it. All 30 checks are green. `actionlint` is clean, and `zizmor` reports nothing new versus `main`.
2. **Reviewed.** CodeRabbit reviewed this head with no actionable comments. Its summary names `ab0d2303`, and its status reads `Review completed`. No review threads.
3. **Tried as a user, as far as it can be before merge.** The job only runs on a push to `main`. It needs the Actions token's package read to reach the private manifests artifact. My local token cannot read it (GHCR returned 403), so the live read cannot happen before merge. The change only reads and adds annotations: it never deploys, and its worst failure is a red report job on `main`. **After merge:** read the first `🧭 Observe Production Convergence` run. It should show `CONVERGED`, which is also #3846's last open acceptance point. If it shows anything else, I'll fix it as breakage.

PR #3873, opened 2026-09-17
fix(security): declare CPU limits on the datapath DaemonSets

devantler:

### Why

The node datapath DaemonSets (Cilium, Cilium Envoy, Tetragon, and the Hetzner storage node plugin) only get a CPU limit from a namespace default applied when their pods start. On a freshly built cluster they can come up before that default exists, and run with no limit at all.

### What

Declares the same CPU limit they already receive today directly in each chart's values, so a rebuilt cluster gets it from the start. The requests are pinned to today's values, so the running pods' effective limits do not change — this only rolls each DaemonSet once. The existing CPU-limit guard now covers these containers too.

**Rollout decision:** this rolls through the normal rolling update, the same path every Cilium chart upgrade takes. The staged rollout gate is reserved for Cilium device-set changes; here nothing about the datapath configuration changes. Throttling was measured before the change (issue comment) and will be measured again after deploy.

Part of #3790


devantler:


@coderabbitai review

devantler:


@coderabbitai review

devantler:

## Readiness at `55530e8a215cf46ef617594cc199fbea69eefb7f`

**1. Programmatically tested**
- `scripts/tests/test-kube-system-addon-cpu-limits.sh` now covers 19 containers. Run against `main` 1dff6239, it fails on exactly the 11 new assertions (the seven datapath containers). On this branch it passes, and all 8 injected defects are caught for the named reason.
- `ksail workload validate` passes for the local and prod overlays. All 30 checks are green, including `🔐 Validate Production Authorization` after the re-approval.
- Authorization conservation: all five authorization overlays rendered on `main` 1dff6239 and on this branch give 573 identities per side, none added or removed. The grant-bearing RBAC and ServiceAccount documents hash identically. The only difference is 28 added CPU lines in the controllers overlay. Both negative controls fire (a synthetic ClusterRole changes the hash, and a dropped identity is reported).

**2. Reviewed**: CodeRabbit reviewed this head with no actionable comments. Its summary covers 1dff6239..55530e8a, and the head status reads `Review completed` at 11:29:44Z. There are no review threads.

**3. Tried as a user (rendered output + live read-only cluster)**: nothing can run pre-merge, because this only changes chart values. So I checked the two things that decide the outcome:
- `helm template` of cilium 1.20.1, tetragon 1.7.1 and hcloud-csi 2.23.0 with these values. The `cilium`, `cilium-envoy`, `tetragon` (both containers) and `hcloud-csi-node` (all three containers) DaemonSets now carry `limits.cpu: "2"` with the pinned requests.
- Live prod, read-only: the running pods already have these ratios from the LimitRange, scaled by the VPA. cilium-agent is 143m/1430m, which is 200m:2. cilium-envoy is 50m/2, tetragon 50m/1, which is 100m:2, and csi-node and export-stdout are 50m/6666m, which is 15m:2. The declared pairs keep every ratio, so the effective limits after rollout are unchanged. The only effect is one rolling restart per DaemonSet.

**After deploy:** re-dispatch `Measure Datapath CPU Throttling` and record the result on #3790 against the 06:13Z baseline.

PR #3874, opened 2026-09-17
fix(security): supply the C-0211 baseline context to audit-log-forwarder

devantler:

### Why

The security scan still reports the audit-log forwarder as missing two baseline security settings. Its namespace is left out of the cluster-wide rule that normally adds them, because the forwarder needs to read a root-owned host file. These two settings are not privilege settings, so the exclusion hides a gap it was never meant to cover.

### What

Adds both settings to the forwarder through the chart's own options. Everything it already enforces stays as it is. It follows KEDA as the next workload to close.

Part of #3239

**Merge order:** this likely changes the production authorization fingerprint. If another fingerprint change merges first, this PR must be re-approved on top of it.


devantler:


@coderabbitai review

devantler:

**Readiness at `1ef085d013cc601c052ca79c8459dd24d1d4ae74`**

- **Tested:** CI 23 pass / 7 skipped, including `🔐 Validate Production Authorization` after the fingerprint re-approval. `ksail workload validate` passes locally for both the local and prod overlays (635 files each).
- **Reviewed:** CodeRabbit review at this head, with no actionable comments; 0 review threads.
- **Tried as a user:** this change has no runtime behaviour to run before it deploys, so I checked its rendered output instead. `helm template` of `opentelemetry-collector` 0.172.1 with this release's values, before and after the change: the DaemonSet gains exactly `securityContext.fsGroupChangePolicy: OnRootMismatch` and container `seLinuxOptions: {}`, and nothing else changes. The existing root/`DAC_READ_SEARCH` context is untouched, and no `fsGroup` is set, so no volume ownership changes.
- **Authorization conservation:** rendered all five authorization overlays on this branch and on main `48e8bec3`. Both sides have the same 569 identities. The 94 RBAC and ServiceAccount documents hash identically, and a synthetic ClusterRole changes that hash. The only diff is 3 value lines.

After deploy, I'll check that the live `audit-log-forwarder` DaemonSet template carries both fields and its pods roll out Ready.

PR #3875, opened 2026-09-17
fix(security): supply the C-0211 baseline context to the kubescape workloads

devantler:

### Why

The Kubescape namespace is excluded from the automatic security-context policy, so its seven workloads never received the two C-0211 fields that every other namespace gets. They are among the last gaps keeping the C-0211 population from closing.

### What

Adds the two fields to the Kubescape scanner, vulnerability scanner, operator, storage, both scan schedulers and the node agent. The node agent keeps the SELinux type it needs to watch host processes. Nothing else about the workloads changes, and the pods roll once.

Part of #3239


devantler:


@coderabbitai review

devantler:

**Readiness at head `3af2e57843ea209c2e0628bce53b9400070367dc`**

- **Tested:** CI green (23 passed, 7 skipped), including production authorization after the fingerprint re-approval. `ksail workload validate` passes for both local and prod.
- **Reviewed:** CodeRabbit reviewed this head and posted no actionable comments. 0 review threads.
- **Tried as a user:** there is no cluster to run, so I inspected the rendered output. I rendered chart kubescape-operator 1.40.4 with this HelmRelease's values and applied its post-renderer patches with `kubectl kustomize`. All seven workloads now carry `fsGroupChangePolicy: OnRootMismatch`. The four Deployments and both scheduler CronJobs also carry `seLinuxOptions: {}`, and node-agent keeps `spc_t`. The authorization conservation check shows the only rendered difference is these patches: 569 identities, and grant documents hash identically to main.
- **Post-deploy:** I will confirm the live workload templates and rollout health after merge.

PR #3876, opened 2026-09-17
fix(security): supply the C-0211 baseline context to longhorn-manager and the driver deployer

devantler:
> 🤖 Generated by the Agentic Engineer

### Why

Kubescape's C-0211 control still flags the Longhorn storage workloads, because Longhorn's namespace is excluded from the cluster-wide security-context mutation and the chart has no setting for these two fields. This is the Longhorn slice of #3239.

### What

Adds the two missing security-context fields to the two remaining workloads the Longhorn chart renders: the storage manager and the driver deployer. Neither pod changes file ownership or SELinux behaviour. The other eight flagged Longhorn workloads are created by Longhorn itself at runtime, so a chart patch cannot reach them; they stay open on #3239.

**Operational note:** the storage manager restarts node by node when this deploys. Volumes stay attached during that restart.

Part of #3239

PR #3878, opened 2026-09-17
fix(security): declare the C-0211 baseline context on the Hetzner CSI workloads

devantler:

### Motivation

Our cluster-security scan flags every workload in `kube-system` for two missing
declarations, because that namespace is deliberately excluded from the mutation that
supplies them everywhere else. The exclusion exists to protect privilege settings, but
these two fields are not privilege settings — so the scan has been reporting a gap we
never needed to have.

### Changes

Declares the two fields on both Hetzner CSI workloads (the node agent and the controller),
clearing 2 of the 13 remaining `kube-system` workloads. The values are chosen so nothing
about how these pods run actually changes: one field is provably a no-op for the volume
types these pods use, and the other declares the setting while leaving SELinux entirely to
the runtime, exactly as today.

No operational action needed. Pods restart once as the change rolls out.

Part of #3239


devantler:


@coderabbitai review

devantler:


@coderabbitai review

devantler:

## Readiness at `aec94a3f29f57ef1600adc59b45ebcc67989d286`

**1. Programmatically tested.** All required checks green at this head (30 total: 23 success,
7 skipped, 0 failing). The three that matter here:

- `🔐 Validate Production Authorization` = SUCCESS — it ran and passed, i.e. the in-PR
  fingerprint re-approval (`a4781e58`) is accepted, not skipped.
- `🧪 Validate Manifests` = SUCCESS
- `CI - Required Checks` = SUCCESS

**2. Reviewed.** CodeRabbit green **at this exact head**: its auto-generated summary names
`aec94a3f29f57ef1600adc59b45ebcc67989d286` and states "No actionable comments were generated in the
recent review"; the head's `CodeRabbit` commit status reads `Review completed` (20:17:58Z), so a
review genuinely ran rather than being skipped or rate-limited. Its own note on the change:
"The Hetzner CSI security-context declarations and their approved rendered-surface fingerprint
update are consistent with the stated change. No actionable merge risk remains."
Review threads: **0 total, 0 unresolved** (paginated, `fetched == totalCount`). No non-thread
finding sections.

**3. Tried and evaluated as a user.** For a GitOps manifest change the cluster applies exactly the
rendered output, so the patched render is the surface that actually observes the effect, and it was
exercised rather than reasoned about — recorded earlier on this PR: the patched-vs-unpatched render
diff is **exactly the 10 intended additions** (`fsGroupChangePolicy: OnRootMismatch` plus per-container
`seLinuxOptions: {}` across the node DaemonSet's 3 containers and the controller Deployment's 5),
with `privileged: true` preserved on the node plugin container, and `ksail workload validate` rc=0 for
both the local and prod configs.

One thing worth stating plainly, because it is the part a reader should not take on trust:
**`OnRootMismatch` is inert for these two workloads today.** The node DaemonSet mounts only
`hostPath` volumes, and the kubelet never applies fsGroup ownership to hostPath; the controller's
single volume is an `emptyDir`, which is root-owned at every start so the GID always mismatches and
the same chown happens either way. The field is declared so the C-0211 baseline is explicit and a
future volume type inherits the right policy — it is not expected to change behaviour now.

Live post-merge verification (DaemonSet/Deployment template fields present, rollouts Ready, restart
counts) follows the same pattern as #3874 / #3875 / #3876 and will be recorded on #3239.

PR #3881, opened 2026-09-17
fix(security): pin the Kubescape relevancy capability instead of inheriting it

### LINKED ISSUE #3707: [feature]: Enable relevancy scan in Kubescape
### User Story

**As a** maintainer of the platform,
**I want** to isolate the findings kubescape makes to only the ones that have relevance (are in use),
**So that** I can fix what matters and not irrelvant stuff.

### Acceptance Criteria

- [ ] Criteria 1: Describe the first acceptance criterion here.
- [ ] Criteria 2: Describe the second acceptance criterion here.
- [ ] Criteria 3: Describe additional criteria as needed.

devantler:

### Why

Relevancy scanning is what makes our vulnerability reports actionable: the node agent watches which
libraries a container actually loads, so unreachable CVEs get marked `not_affected` instead of
competing for attention with the ones that matter. We rely on it — but we never asked for it. It is
on purely because the chart happens to default it on, so a future chart bump that flips that default
would switch a security capability off silently, with nothing in our config to notice or review.

### What

Declares `relevancy: enable` in the Kubescape capabilities we pin, alongside the ones we already pin
there. No behaviour changes today — this makes the current state intentional and reviewable instead
of inherited, so it can only change deliberately.

Fixes #3707


devantler:


@coderabbitai review

devantler:

**Readiness at `edee95c3b697d1c88cb05702ae5f1fb0fd29c616` — promoting.**

| Condition | Evidence |
|---|---|
| Programmatically tested | All required checks green, including `🔐 Validate Production Authorization` = SUCCESS. The re-approval carries a 573-identity conservation with **two negative controls that both fired** (a synthetic ClusterRole changes the grant hash *and* adds an identity), and the aggregate was read from CI's own approved renderer, not this host's unapproved one. |
| Reviewed | CodeRabbit at this exact head — auto-summary refreshed 23:26:38Z after the 23:23:13Z request, names the head, **zero actionable finding sections** (only the non-actionable `🔇 Additional comments (3)`), no rate-limit or did-not-run marker on the newest same-head reply. Threads 0/0, `fetched == totalCount`. |
| Tried and evaluated as a user | Live on the production cluster: `ks-capabilities` reads `relevancy=enable`, with **1133** `vulnerabilitymanifests` labelled `kubescape.io/context: filtered` against **235** non-filtered — i.e. relevancy is not merely configured, it is demonstrably producing the filtered verdicts that make CVE triage actionable. That is the behaviour this PR pins. |

### One thing I checked before promoting, because it would have failed in the queue

`main` moved **twice** after this branch's conservation was computed against `eea5453d` — #3880 (22:40Z) and #3884 (23:07Z). A retained aggregate is only valid while `main` has not moved, and #3884 touched `k8s/bases/apps/fleetdm/helm-release.yaml`, which sits under `appsOverlayPath` — **one of the five `authorizationOverlayPaths`**. On its face that stales `previousRenderedSurfaceSHA`.

It does not, and the proof is structural rather than a re-render:

1. **#3884 merged through the queue without touching `scripts/validate-eks-ci-role-policy/main.go`.** The gate runs on `merge_group` whenever `k8s` changes, so had that merge moved the aggregate, #3884's own gate would have failed. It passed.
2. **Mechanism:** `fleetdm/` is commented out of `k8s/bases/apps/kustomization.yaml` (*"fleetdm disabled 2026-06-03 — not in use"*), so nothing under `bases/apps/fleetdm/` is rendered at all. Changing that HelmRelease is inert to the authorization surface.

#3880 is separately inert here — it changed only `.github/workflows/{ci,cd}.yaml` and a test script, i.e. no manifests.

So both constants stand as written, and the merge-group gate re-evaluates against the real base regardless.

Issue #3894, opened 2026-09-19
Eliminate remaining Coroot operational warnings


devantler:
> 🤖 Generated by the Agentic Engineer

## Problem

Coroot still reports operational warnings after the main remediation landed. Fresh production evidence identifies one actionable probe defect and three signals that need narrowly scoped declarative disposition without hiding neighboring failures. This issue is separate from #3812 and does not change its availability-risk scope.

## Root-cause evidence

- Kubescape emitted one liveness failure during the daily scan because its 1-second timeout was too tight under scan load. The pod remained Ready with zero restarts and memory returned to its baseline.
- Cilium aggregates DNS-proxy observations; Kubernetes resolver search-path candidates return NXDOMAIN before the canonical service or external name succeeds. DNS latency and server-error checks remain healthy.
- kube-controller-manager logs the ReplicaSet consistency guard when its informer read version trails a version it just wrote. The controller rate-limits and requeues the key, and every observed Deployment converged.
- Alertmanager carries a two-event DNS-server alert last updated August 27. The current pod has no DNS or error recurrence; future DNS failures must remain visible.
- The transient Kubescape memory-growth warning resolved after the scan and must not be suppressed.

## Acceptance criteria

- [ ] Increase only the Kubescape liveness timeout enough to tolerate scan load while retaining the probe.
- [ ] Add exact declarative Coroot exemptions for the Cilium NXDOMAIN rule and ReplicaSet consistency-requeue message.
- [ ] Retire only the immutable historical Alertmanager alert record; do not suppress future Alertmanager DNS failures.
- [ ] Add regression tests proving exact matches are suppressed and near matches remain visible.
- [ ] Render and validate the production manifests.
- [ ] Merge through protected GitOps delivery and verify the deployed revision, workload health, and fresh Coroot alert state.

devantler:
> 🤖 Generated by the Agentic Engineer

Closed after live production verification.

- PRs #3895, #3897, #3902, #3903, and #3905 are merged; #3905 deployed as `4cac44f6ef1f2ae73f540c01a02a850df5409bb1` / signed OCI `latest@sha256:97aa7991543ad6c9478cf727cf6954afa71f5eae57108be3b8f8be94f8247e9e`.
- All Flux layers are Ready on the deployed revision.
[...]

PR #3902, opened 2026-09-19
fix(observability): close remaining Coroot alert gaps

devantler:

## Summary

- make Alertmanager StatefulSet peer DNS available before readiness while keeping the client Service readiness-gated
- give the Flux kustomize-controller a native ten-second pre-stop grace period so terminating endpoints drain before the process exits
- make reviewed Coroot exemptions survive alert-ID rotation by binding stable fingerprints to exact applications, rules, and current representatives; changed representatives are reopened automatically
- recognize only the exact Flux `HealthCheckFailed` cancellation emitted during the protected image-credential handoff
- increase the Coroot node-agent L7 perf ring patch from 128 to 512 pages per CPU after production still dropped 164 samples

## Root causes and treatment

- Alertmanager peer lookups returned NXDOMAIN during HA startup because its headless gossip Service did not publish not-yet-Ready endpoints. This is fixed declaratively on that Service only.
- kustomize-controller exited before endpoint removal propagated, creating a transient readiness refusal and canceling an in-flight health check. Shutdown now has a native pre-stop grace period; the exact cancellation is declaratively exempted as the expected handoff signal.
- Coroot alert IDs rotate when a resolved fingerprint fires again. The previous ID-pinned exemptions therefore missed semantically identical API-server and kubelet signals. Exemptions now use stable fingerprints plus narrow semantic guards and fail open by reopening changed representatives.
- Talos kubelet logs prove the runtime signals are CRI `NotFound` cleanup races or the exact unsupported `UpdatePodSandboxResources` call, so those remain reasoned exemptions rather than product fixes.
- The Coroot node agent's enlarged 128-page L7 buffer still overflowed. The reviewed patch now builds 512 pages per CPU; the immutable digest pin follows after this change reaches main and the main-only publisher signs it.
- The old node-agent readiness event is not exempted. It is allowed to expire naturally; the already-deployed lifecycle fix prevents recurrence.

## Evidence

- RED: focused contracts initially rejected the missing Alertmanager peer-DNS patch, Flux pre-stop grace, 512-page buffer, rotated alert IDs, and exact health-check cancellation.
- GREEN: all four Coroot remediation contract suites pass.
- full `k8s/providers/hetzner/infrastructure` Kustomize render passes.
- the exact-base EKS authorization validator passes against main `d4442f10428caf2c4d3849bfdd009a770a3cdcff` using the repository-approved kubectl v1.36.2 binary.
- the Coroot patch applies cleanly to the exact pinned upstream source. Package tests are CI-authoritative because the upstream package has no Darwin implementation.
- `git diff --check` and shell lint pass.

## Authorization conservation

The rendered authorization surface moves only because `kubescape/alertmanager` is a selected HelmRelease. No ClusterRole, Role, binding, ServiceAccount, identity, verb, wildcard, AWS identity, or permission changes.

Relates to #3894. The issue remains open until the separately published node-agent image is pinned, deployed, and the live Coroot census is clear.


devantler (commented):

Exact-head self-review of `136b20c1084c305b0536737d51a73ca204703b45` completed with no findings.

Reviewed the rendered Alertmanager Service selector and readiness boundary, Flux lifecycle patch semantics, fingerprint/application/rule/representative matching and fail-open reopen path, exact event label constraints, 512-page memory tradeoff, authorization-surface conservation, and focused regression coverage. The branch remains draft until all terminal checks finish.

PR #3905, opened 2026-09-19
fix(observability): classify Flux handoff cancellations

devantler:

## Why

Coroot still reports three warnings created when the protected production rollout deliberately hands Flux reconciliation to a restarted controller. The affected reconciliations retried successfully, so these records obscure real operational alerts without indicating failed desired state.

## What

Classify only the exact cancellation records emitted by that handoff, including the live unattributed event identity and post-build cancellation shape. Keep permission, validation, timeout, dependency, and mismatched event records visible through explicit negative contracts.

Refs #3894


devantler:

Exact-head self-review completed for `f5829fe4e670278f3a0c95df29d23b696f117662`: no findings.

Reviewed the complete diff against current base `e0cce893896718a764ec55127220da6f9fb42a2c`. The exemption is fail-closed on the unattributed Coroot event application, complete Flux label set, source, and per-reason message shape; permission, validation, timeout, wrong-reason, and wrong-application near matches remain visible. No authorization surface changes; the pinned kubectl 1.36.2 exact-base authorization comparison and focused remediation contracts pass.

PR #3906, opened 2026-09-19
fix(kyverno): record reviewed RBAC exclusions as a current skip

### LINKED ISSUE #3900: Reviewed privileged-RBAC exclusions leave 12 stale warnings in cluster policy reports

## Evidence

Read-only against prod on 2026-09-19: of 281 ClusterPolicyReports, **12 hold a `warn` result that is 129 hours old**, all from `audit-privileged-rbac` / `privileged-rules`. The same rule's other **229** results are `pass` and less than an hour old, so the background scan is running and rewriting results for this rule. Only these 12 are frozen.

The 12 subjects are exactly ClusterRoles that #3791 later moved into the policy's reviewed `exclude` list (for example `cluster-admin`, `admin`, `edit`, `longhorn-role`, the Crossplane and KRO roles). As `docs/policy-reports.md` already explains, an `exclude` produces no replacement result, so each audit-era warning stays in its report indefinitely.

## Who this affects

Anyone reading the Policy Reporter dashboard. It shows 12 privileged-RBAC warnings for roles the platform has already reviewed and accepted, so the count stays wrong, and a real new warning is harder to spot next to it.

## Why the report-pruning work does not cover it

The `prune-stale-policy-reports` DeletingPolicy (platform#3899, part of #2573) matches namespaced `policyreports` only, and its delete grant is scoped the same way. Cluster-scoped reports are out of scope by design.

## Expected behaviour

A reviewed exclusion shows as a current `skip`, or its report no longer lists a result for that rule, and nothing reports stale failures for roles that were deliberately accepted.

## Options

1. **Root cause (preferred):** express the exact-identity exclusions as a false `preconditions` entry rather than `exclude`, which is the shape `docs/policy-reports.md` already recommends. The result becomes a current `skip`. The exact-identity matching in `scripts/tests/test-audit-privileged-rbac.sh` must keep its guarantees (name and kind for ClusterRoles, namespace and name for Roles).
2. Extend the prune policy and its grant to `clusterpolicyreports`. This is broader, and it only clears the current symptom while the `exclude` keeps producing new stale results.

## Acceptance criteria

- [ ] Prod shows 0 `audit-privileged-rbac` results older than 6h in ClusterPolicyReports, after one background scan.
- [ ] The existing privileged-RBAC tests pass, and a new test proves an excluded role yields `skip` rather than no result.
- [ ] Non-excluded privileged roles still produce `warn`/`fail`.

## Size

Small to medium: one policy's exclusion shape plus its test.

devantler:

### Why

The policy-report dashboard shows 14 privileged-RBAC warnings (12 cluster-wide, 2 namespaced) for roles the platform has already reviewed and accepted. They were reported before being accepted, and the way the policy exempts them records nothing new, so the old warnings never go away. They inflate the count and make a real new warning harder to spot.

### What

The reviewed exemptions now record a current "skipped" result on every background scan, so an accepted role shows as accepted and its old warning is replaced. The list of exempted roles, and the rule that each exemption covers exactly one role in one place, are unchanged. Deployment is unchanged too: Flux applies it and the next hourly scan clears the 14 warnings.

Fixes #3900


devantler:


@coderabbitai review

devantler:

**Tried as a user, at head `7922d006`.**

- **Prod, before (read-only, 2026-09-19 ~16:00Z):** `audit-privileged-rbac` holds **12** stale `warn` results in ClusterPolicyReports and **2** in namespaced PolicyReports, every one older than 6h; the rule's other 270 results are fresh `pass`.
- **The background-scan result the change produces:** `kyverno apply --policy-report` (Kyverno 1.19.1, the CI version) on the policy's fixtures records `ClusterRole||crossplane-rbac-manager` and `Role|velero|velero-server` as `skip` with `process: background scan`, while the same-name lookalikes `Role|sample|velero-server`, `ClusterRole||velero-server` and `ClusterRole||crossplane-rbac-manager-lookalike` still `fail`. Census: `pass: 15, fail: 24, warn: 0, error: 0, skip: 2`.
- **RED:** the unchanged policy fails the new shape check, and with shape and census neutralised it fails the new report assertion alone (`did not skip exactly the reviewed exclusions`).
- `ksail workload validate` passes for both overlays; `test-chart-rendered-rbac-policy.sh` passes against the new precondition inventory.
- **After merge:** confirm on prod that the 14 stale results are replaced by fresh `skip` results after one hourly background scan (acceptance criterion on #3900).

devantler:


@codex review

PR #3907, opened 2026-09-19
fix(observability): reconcile Coroot application health

### LINKED ISSUE #3894: Eliminate remaining Coroot operational warnings

devantler:

## Why

Coroot’s alert feed is clear, but its Applications view still reports warning and error health. The remaining signals combine one real stale control-plane watcher with bounded or third-party behavior that should not keep otherwise healthy applications degraded.

## What

Refresh the controller managers so retired-resource watchers stop, emit platform-owned lifecycle logs as structured information, and declaratively reconcile narrow per-application thresholds only while every live signal matches its reviewed reason. Any new log pattern or network upstream automatically restores Coroot’s fail-visible default.

Fixes #3894


devantler:

Exact-head readiness review for `452fa2773695ef4969326fed40a941e83a435eb5`:

- Re-read the complete diff against the current base and found no remaining correctness, security, operability, or GitOps convergence issues.
- Verified that every application-level exception is exact-pattern scoped, finite, reasoned in the manifest, and fails closed: a new or aged-out pattern resets the app to the global zero-error policy.
- Verified that the actionable sources are fixed rather than hidden: lifecycle output is structured as `info`, and the controller-manager static pods receive a declarative restart while preserving the tuned PodGC threshold.
- Reproduced the CI ShellCheck portability failure locally, replaced the ambiguous boolean chain with an explicit conditional, and reran ShellCheck plus the executable reconciler contract successfully.
- Validated both cluster renders, the production infrastructure render, YAML parsing, the focused Coroot contracts, and the full Talos patch stack using the repository-pinned Talos 1.13.9 toolchain.

The exact-head protected CI run is green. No GitHub Copilot review was requested or used.

PR #3908, opened 2026-09-19
fix(observability): clear kubelet baseline warning

### LINKED ISSUE #3894: Eliminate remaining Coroot operational warnings

devantler:

## Why

The first production reconciliation from #3907 removed every Applications-view log warning, but kubelet remained warning even though Coroot showed the reviewed count of two failed static-pod probe upstreams and the configured threshold of two. Live readback proves Coroot warns at equality despite displaying a greater-than condition. The refreshed census also surfaced a Crossplane memory-growth warning that was not present in the earlier one-hour view.

## What

Set the narrowly-scoped kubelet application threshold to three, the smallest value that clears the exact reviewed count of two. The existing fail-closed upstream-set check is unchanged: any third failed upstream removes the exemption instead of accepting it.

Add a Crossplane-only memory threshold of 35 after a seven-day review showed a bounded daily cache sawtooth: its 22-day-old, restart-free pod repeatedly returned to 270–300 MiB after 451–520 MiB peaks while Coroot's short-window regression reported 27% per hour. The finite threshold continues to expose sustained growth outside that cycle. Update the executable contract to lock both live-observed boundaries.

Fixes #3894


devantler:

Exact-head readiness review for `9d339f753b47aa0ab4ef389d91247799e51dc496`:

- Re-read the complete two-file diff against merge commit `aa344e8575a96856689291f8c280b48824dba3ff`; no remaining correctness, security, operability, or fail-closed issues found.
- Confirmed from the live Coroot report that the reviewed count is two, the active upstream set is exactly kube-scheduler plus kube-controller-manager, and Coroot still warns when its displayed threshold is also two. Threshold three is therefore the smallest effective boundary, while the existing third-upstream guard still removes the override.
- Measured Crossplane's seven-day RSS history: its 22-day-old, restart-free pod repeatedly returned to 270–300 MiB after bounded 451–520 MiB daily peaks. A 35% per-hour app-only threshold clears the observed 27% short-window regression while preserving a warning for sustained growth outside that cycle.
- Added both expectations to the executable contract first, observed the test fail, then implemented the policies and reran the focused test, ShellCheck, diff validation, and production render successfully.

The exact-head protected CI run is green. No GitHub Copilot review was requested or used.

PR #3909, opened 2026-09-19
fix(observability): classify Talos state timeout

### LINKED ISSUE #3894: Eliminate remaining Coroot operational warnings

devantler:

## Why

The production reconciliation from #3908 cleared the reviewed kubelet and Crossplane warnings, then the next Coroot census surfaced one Talos init log pattern emitted during that rollout: a containerd state lookup for one container. The pattern text alone is not safe to exempt because a different underlying error can use the same message shape. The raw Coroot entry proves this occurrence was a bounded context deadline during the rollout and came from the Talos init service.

## What

Extend the existing finite Talos runtime log policy to classify a container state lookup only when every matching raw entry carries both exact evidence attributes: error=context deadline exceeded and service.name=/talos/init. Missing logs, mixed causes, query failure, a different service, or any other error removes the application override so the signal remains visible.

Add executable positive and negative coverage. The negative case starts with the override installed and proves that a same-shaped permission-denied event declaratively removes it.

Fixes #3894


devantler:

Exact-head review completed for `80f0703469bd9da6e7fdd5e873e48fb5810f7f78`.

Reviewed the complete two-file diff and verified the trust boundary end to end:

- the container-id-shaped message is only a candidate selector;
- every distinct matching pattern must join to at least one raw Coroot entry;
- every joined entry must carry both `error=context deadline exceeded` and `service.name=/talos/init`;
- missing, mixed, differently attributed, or unavailable evidence removes the override;
- the finite Talos cap remains unchanged, so known-message storms still surface;
- positive and already-installed negative-removal paths are executable in the contract test.

Local evidence: focused behavior contract green, embedded POSIX shell green apart from the pre-existing intentional curl retry-word splitting, production Kustomize overlay renders, and diff hygiene is clean. No blocking findings. CI is still running; this PR remains draft until terminal exact-head checks are green.

PR #3910, opened 2026-09-19
fix(observability): align control-plane probe listeners

### LINKED ISSUE #3894: Eliminate remaining Coroot operational warnings

devantler:

## Why

After #3909 deployed, Coroot's Applications view retained one kubelet warning: two failed upstream services even with an application threshold of three. Live API readback and Coroot's implementation prove the threshold cannot suppress this class of warning: item-based checks fire whenever their item set is non-empty.

The actual source is a Talos control-plane mismatch. Talos generates kube-scheduler and kube-controller-manager health probes with host `localhost`, which resolves to IPv6 loopback first on these nodes, while both components listen only on IPv4 loopback. Every healthy probe therefore records a failed `::1` attempt before succeeding on `127.0.0.1`.

The first protected deployment attempt exposed a second production-safety defect: `ksail cluster update` returned after applying the in-place Talos configuration but before the resulting static control-plane pod restarts began. The Kubernetes API became unavailable 23 seconds later, so the next credential-fencing operation failed and the merge queue correctly restored main.

The second protected attempt proved that stability gate catches the delayed restart and waits through recovery, then exposed a separate Flux handoff race. Suspending a Kustomization stops later reconciliations but does not cancel one already running; taking the suspension fence while the root reconciliation was active could preserve its `Reconciling=True` condition indefinitely and block the final GHCR policy handoff. The merge queue again restored main.

The third protected attempt proved both of those repairs in production: the API gate reset after six successful probes when the delayed restart arrived, then required 20 consecutive successes; all six control-plane components became Ready on `::1`; and the pre-fence Flux reconciliation quiesced. That reconciliation had, however, started an asynchronous rollout of the `kustomize-controller` Deployment. The policy handoff sampled its Pods once during replacement and failed even though the Deployment subsequently converged to 2/2 Ready. The protected heal restored main and all fences were released.

## What

Bind the scheduler and controller-manager health endpoints to `::1`. The endpoints remain loopback-only, but the kubelet's first `localhost` attempt now reaches them without a failed fallback connection.

Remove the obsolete Coroot threshold declaratively by reconciling the exact kubelet override back to inheritance. Update the executable contract to require both loopback listener settings and to prove the stale threshold is removed rather than widened.

Make the production deploy wait for 20 consecutive Kubernetes `/readyz` probes after any attempted Talos machine-config update before resuming cluster mutations. The GHCR synchronization bridge also retries only a classified API transport interruption during its initial read-only Lease inspection; authorization errors, malformed state, and conflicting ownership continue to fail closed.

Wait for the owning Flux Kustomization to quiesce before acquiring its exact-resource-version suspension fence, and retain the post-fence quiescence check for a reconciliation that starts between observation and the patch. This closes both sides of the handoff race without weakening ownership, lease, or compare-and-swap protections.

Before intentionally restarting `kustomize-controller` for the credential-policy handoff, wait for its declarative Deployment rollout and every selected Pod to stabilize. Sample the Deployment resource version on both sides of the Pod proof, then retain the existing exact-resource-version restart CAS. A rollout that remains unready still times out, releases both policy fences, and blocks mutation.

## Validation

- Coroot application-health behavior contract
- Kubernetes/Talos compatibility validation
- production deployment action ordering contract
- RED→GREEN API interruption and stable-window tests
- RED→GREEN Flux pre-fence quiescence and stale-condition regression test
- RED→GREEN Flux controller declarative-rollout stabilization and timeout tests
- full GHCR bridge Go suite (all runnable local cases)
- Bash syntax and ShellCheck
- diff hygiene

The repository's pinned Talos v1.13.9 CI gate remains authoritative for the full machine-config render and the protected merge-group deployment proves the production behavior.

Fixes #3894


devantler:

Exact-head review completed for `46d32288f1145bdfa9cb1b8afafe145aa434b06c`.

Reviewed the full three-file diff against the live topology and the upstream implementations:

- Coroot's `CheckTypeItemBased` path fires on any non-empty item set and does not consult the displayed threshold, so removing the old override is required;
- the live Talos static pods probe `localhost` while both components currently bind `127.0.0.1`, matching Coroot's two failed-upstream series;
- `::1` retains the same loopback-only exposure while matching the first address selected for `localhost` on these nodes;
- both scheduler and controller-manager move together, avoiding a partial fix;
- the reconciler actively deletes the stale exact-app override and leaves all other bounded policies unchanged;
- the executable contract locks the listener settings and the cleanup behavior.

Local behavior, shell, render, version-compatibility, and diff checks are green. The installed local talosctl is v1.14.0, so the repository's pinned v1.13.9 CI machine-config render is the authoritative schema check. No blocking findings. The PR remains draft until exact-head CI is terminal green.

devantler:

Exact-head self-review completed for `2718b608ed5221099f6242af3e767e8025e7b5b6`.

No blocking findings.

Reviewed the whole diff against current `main`, including:

- Talos scheduler/controller-manager listeners remain loopback-only and match the IPv6-first `localhost` health probes.
- The obsolete Coroot item-check threshold is removed declaratively rather than widened.
- The post-Talos deployment gate requires a sustained Kubernetes API-ready window before later mutations and also runs after a partial update failure.
- A skipped machine-config update preserves the existing reassert path; an attempted update requires the stability proof.
- Initial synchronization-Lease recovery is restricted to classified transport interruptions; authorization, malformed state, and conflicting ownership remain fail-closed.
- RED→GREEN coverage exercises the observed pre-claim connection refusal and the delayed readiness interruption/reset path.

Local evidence: Coroot reconciler behavior tests, Kubernetes/Talos compatibility test, Bash syntax, ShellCheck, standalone GHCR safety tests, API-stability behavior test, deployment action contract test, and the full runnable GHCR Go suite are green. One pre-existing macOS sandbox-only test cannot write through `/dev/stdout`; the rest of that package passed, and CI will run the complete suite on Linux.

No GitHub Copilot review was requested or used.

devantler:

Exact-head self-review completed for `7b97b280466f03b767ebb2cabcdd989ac1162b6b`.

No blocking findings.

Reviewed the complete diff against current `origin/main`, including the Talos loopback listener change, declarative Coroot override removal, post-update API stability gate, classified pre-claim API retry, and the Flux pre-/post-suspension quiescence fence. The new fence preserves the existing Lease ownership, resource-version CAS, and release protections while preventing an in-flight Flux reconciliation from being frozen with a stale `Reconciling=True` condition.

Current-head evidence:

- targeted Flux ownership, contention, release, and new stale-condition regression tests passed;
- production API stable-window behavior test passed;
- GHCR safety regressions passed;
- Coroot application-health behavior contract passed;
- Bash syntax, source-complete ShellCheck, and diff hygiene passed.

GitHub CI remains the authoritative Linux/full-manifest gate. No GitHub Copilot review was requested or used.

devantler:

Exact-head self-review completed for `6ccff50feef7bb004ddce909d016fa05726285a3` against current `origin/main` (`c3680b017b037577ab20603c865ffeeb7532806f`).

No blocking findings.

The third protected merge-group run supplied production evidence for the remaining edge: after the root Flux reconciliation quiesced, its asynchronous `kustomize-controller` Deployment rollout was still replacing Pods. The previous single Pod sample failed safely. This head now waits for observed generation, updated/ready/available replicas, zero unavailable replicas, and every selected Pod to be Ready and non-terminating. It samples the Deployment resource version on both sides of the Pod proof, then retains the exact-resource-version restart CAS.

Safety properties retained:

- a stuck declarative rollout times out and blocks policy mutation;
- both Flux suspension fences are released on failure;
- Lease ownership, parent/child quiescence, restart adoption, and pre-handoff process-replacement proofs remain enforced;
- no threshold broadening or imperative production exception was introduced.

Current-head evidence:

- RED captured from the production failure shape; the new stabilization and timeout cases are GREEN;
- full runnable GHCR bridge suite passed in 173.5 seconds;
- API stable-window, GHCR safety, and Coroot application-health behavior suites passed;
- Bash syntax, source-complete ShellCheck, and diff hygiene passed;
- post-heal live readback showed main restored, all five Flux Kustomizations Ready and quiescent, `kustomize-controller` 2/2 Ready with current observed generation, no residual Lease/fence, and all six control-plane pods Ready.

GitHub CI remains the authoritative Linux/full-manifest gate. No GitHub Copilot review was requested or used.

Issue #3919, opened 2026-09-20
C-0211 on the nine control-plane static pods cannot be mutated by Kyverno, and labelling kube-system would report them falsely compliant


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Nine pods in `kube-system` fail C-0211 on `fsGroupChangePolicy` and `seLinuxOptions` and **cannot be fixed by the Kyverno mutation** that #3239 uses for every other namespace. Measured against live prod specs 2026-09-20:

```
kube-apiserver-prod-control-plane-{2,4,5}
kube-controller-manager-prod-control-plane-{2,4,5}
kube-scheduler-prod-control-plane-{2,4,5}
```

All nine lack both fields. Each carries `ownerReferences[0].kind: Node` and:

```
kubernetes.io/config.mirror: 4f1ad59b7ebf15d444e823aed8c9081d
kubernetes.io/config.source: http
```

They are **kubelet mirror pods** for static pods delivered from Talos machine config. The API object is a read-only projection: kubelet owns the real spec and rebuilds the mirror from it, so an admission-time mutation edits the projection and never the running container.

`kube-system`'s full C-0211 residual is 34 of 61 pods. 25 are ordinary controller-owned pods that #3239's existing opt-in label handles normally; **these 9 are the part that mechanism cannot reach.**

## Who this affects and how

Nobody at runtime — both fields are inert, as #3239's policy documents at length. The impact is on **trustworthiness of the posture signal**, which matters because that signal is what the security work is steered by:

- Leave them as they are, and `kube-system` can never reach the 0-of-N target #3239 sets, so a real and complete rollout looks unfinished forever.
- Label `kube-system` without excluding mirror pods, and Kyverno mutates the 9 projections. Kubescape scores the API object, so they flip to **passing while the actual `kube-apiserver`, `kube-controller-manager` and `kube-scheduler` containers are unchanged.** That is a posture number improving without the cluster improving, and it is invisible from the score alone.

The second outcome is the worse one, and it is the default if `kube-system` is simply labelled like the namespaces before it.

## Expected

C-0211 on the control-plane static pods is either genuinely satisfied, or explicitly and justifiably excepted — never silently reported as passing on a mutated mirror pod.

## The two candidate remedies

[...]

PR #3921, opened 2026-09-20
fix(policies): never mutate kubelet mirror pods into false C-0211 compliance

### LINKED ISSUE #3920: security(kyverno): the C-0211 opt-in can mutate kubelet mirror pods into false compliance

## Evidence

`add-security-context.yaml`'s two namespace opt-in rules (`add-baseline-context-optin-namespaces`, `add-baseline-context-optin-containers`) match `Pod` on `CREATE` with no precondition on how the pod was created. In `kube-system` that set includes the kubelet **mirror pods** for the Talos-served static control plane.

Measured live 2026-09-20 against prod: `kube-system` holds **61** pods, of which **exactly 9** carry `kubernetes.io/config.mirror` — `kube-apiserver`, `kube-controller-manager` and `kube-scheduler` on `prod-control-plane-{2,4,5}` — and **52** do not.

## Who this affects and how

A mirror pod's API object is a read-only projection of a spec kubelet owns and rebuilds from machine config. An admission mutation edits the projection; the running container is untouched. Kubescape scores the API object.

So the moment `kube-system` is labelled for the #3239 rollout, those nine flip to **passing C-0211 while the real control plane is unchanged** — a posture number improving without the cluster improving, and invisible from the score alone. That is strictly worse than the honest FAIL it replaces, because the FAIL is at least true.

Supplying the fields at the source is not an alternative: Talos machine config exposes no `securityContext` on any control-plane component, verified on #3919 against both v1.13 (the version prod runs) and v1.14, on both configuration surfaces and with a strict-decoder negative control.

## Expected

Labelling `kube-system` cannot mutate a mirror pod, so the rollout's remaining 25 addressable pods are fixed genuinely and the 9 keep an honest verdict.

## Acceptance criteria

- Both namespace opt-in rules carry a precondition excluding pods annotated `kubernetes.io/config.mirror`.
- The rendered policy output carries the guard on exactly those two rules.
- Verified against live `kube-system`: the 9 mirror pods are excluded and the other 52 are not.

## Size

XS — a precondition on two rules, plus the reasoning recorded where the next person will read it.

Part of #3919.

devantler:

## Why

The security rollout that has been fixing a posture control namespace by namespace is about to reach the cluster core namespace, where nine of the pods are the Kubernetes control plane itself and are run directly by the operating system. What the cluster stores for those nine is only a read-only copy, so the rollout would edit the copy while the scanner grades that same copy, flipping all nine to passing without the real control plane changing at all. A posture number that improves on its own is worse than an honest failure, because nothing in the score says it happened.

## What

Excludes those nine pods from the rollout, so labelling the cluster core namespace can no longer produce a false pass. Checked against the live cluster, exactly the nine control-plane pods are excluded and the other fifty-two are untouched, so the namespace still gets the genuine fix everywhere it actually applies. It also records why setting the value properly is not an option, since the operating system offers no way to set it on the control plane in the version production runs.

Fixes #3920


devantler:

## Evaluation record — exercised at `e9d27dea`

This repo's "tried as a user" is the cheapest observation that needs no cluster: render the changed policy and check the intended effect, plus read-only checks against the running cluster. I did both, and the second one is the part that actually proves the predicate is right.

**Rendered.** `kubectl kustomize k8s/bases/infrastructure/cluster-policies/` builds clean and the guard appears **exactly twice** in the output — once per patched rule, and nowhere else. I checked the render rather than the source file because the file has no `kustomization.yaml` of its own; building the two cluster overlays would **not** have exercised it (they are 125-line Flux pointers), so a green overlay build would have proved nothing about this change. Both overlays still build too.

**Verified against live prod — the control that matters.** A precondition that excludes the wrong set is the failure mode here, in both directions: miss the mirror pods and the false-compliance outcome stays reachable; over-match and the 25 genuinely addressable pods silently stop being fixed. So I measured the predicate itself against `kube-system`:

| | count |
|---|---|
| pods in `kube-system` | 61 |
| carrying `kubernetes.io/config.mirror` (guard skips) | **9** |
| not carrying it (guard still applies) | **52** |

The nine are exactly the nine this work is about — `kube-apiserver`, `kube-controller-manager` and `kube-scheduler` on `prod-control-plane-{2,4,5}`. Nothing else in the namespace matches, so the opt-in keeps working everywhere it legitimately applies. That is a positive and a negative control on real data rather than on a fixture.

**Judged as the change's user:** what I wanted was to be able to label `kube-system` for #3239 without the posture score lying to me afterwards. With this in place the remaining 25 get the genuine mutation and the 9 keep an honest verdict, which is the outcome #3919 asks for.

**Not claimed:** that the nine now *pass* C-0211 — they do not, and should not. They remain covered by the existing cluster-wide C-0211 exception in `pod-security-mutations-unscoped.yaml`; accounting for them permanently when that exception is eventually narrowed is the rest of #3919, not this PR.

Staying a draft until CI settles and a current-head review lands.

devantler:

Current-head review at `e9d27dea21d82b782513c8ba8ad6781af64c75d7` found one blocking readiness gap: the mirror-pod boundary has no repository-owned behavioral regression test.

The existing `tests/add-baseline-context` suite contains no Pod annotated `kubernetes.io/config.mirror`, so `kyverno test ./tests` remains green if both new preconditions are deleted. I independently reproduced the missing behavior against the same policy: with a labelled `kube-system` fixture, the unguarded rules injected both `fsGroupChangePolicy: OnRootMismatch` and container `seLinuxOptions: {}`; with this head's preconditions, both rules correctly skip it.

Please add the regression at the existing boundary:

- declare `kube-system` as `baseline-context: enabled` in the test values, so a namespace-selector miss cannot pass vacuously;
- add a Pod fixture carrying `kubernetes.io/config.mirror`;
- expect both opt-in rules to skip that fixture; and
- assert the rendered Pod still lacks both mutation fields.

That test fails specifically when this protection is removed or widened incorrectly, while the current live 9/52 positive/negative control remains the runtime evidence.

devantler:

## Resolution of current-head review finding at `d7c1d6f9`

Added the repository-owned behavioral regression test for the mirror-pod exemption at the existing `tests/add-baseline-context` boundary:

- **Test values**: declared `kube-system` as `pod-security.devantler.tech/baseline-context: enabled` in `tests/add-baseline-context/values.yaml` so the namespace-selector matches and evaluation is live rather than reporting "Excluded" (preventing vacuous pass).
- **Pod fixture**: added `kube-system/kube-apiserver-mirror` in `tests/add-baseline-context/resources.yaml` carrying annotation `kubernetes.io/config.mirror: "1234567890abcdef"`.
- **Expected results**: declared `result: skip` for both `add-baseline-context-optin-namespaces` and `add-baseline-context-optin-containers` on `kube-system/kube-apiserver-mirror` in `tests/add-baseline-context/kyverno-test.yaml`.
- **Negative control verification**: tested against the policy; inverting the precondition caused `kyverno test tests/add-baseline-context` to fail specifically on `kube-system/kube-apiserver-mirror` (`Fail | Want skip, got pass`). With the guard in place, all 45 test assertions in `tests/add-baseline-context` pass, and all 344 Kyverno tests across the platform suite pass clean.

@coderabbitai review

PR #3926, opened 2026-09-20
fix(kubescape): declare the scanned-namespace scope instead of inheriting it

### LINKED ISSUE #3923: security(kubescape): the scanned-namespace scope is an inherited chart default the repo neither declares nor verifies

## Evidence

Kubescape's scanned scope on this cluster is decided by one value:

```
excludeNamespaces: "kubescape,kube-system,kube-public,kube-node-lease,kubeconfig,gmp-system,gmp-public"
```

read from `configmap/operator` in the `kubescape` namespace, live prod 2026-09-20.

**That value is not set anywhere in this repository.** `spec.values` in
`k8s/bases/infrastructure/controllers/kubescape/helm-release.yaml` has no `excludeNamespaces` key —
the only mention of it in the whole tree is prose, in a comment on that same file and in two
comments inside `k8s/bases/infrastructure/cluster-security-exceptions/`. The effective value is an
upstream chart default we inherit silently.

Measured consequences today:

| namespace | scanned? | `workloadconfigurationscans` |
|---|---|---|
| `observability`, `velero`, `longhorn-system`, `flux-system` | yes | 145 / 108 / 97 / 79 |
| `kube-system`, `kube-public`, `kube-node-lease` | **no** | **0 / 0 / 0** |
| `kubescape` | **no** | 748, but cluster-scoped kinds only — no workload objects |

(Unfiltered control: 2529 scan objects across 39 namespaces, so the surface is producing.)

## Who this affects and how

The posture signal is what the platform's security work is steered by, and its **scope** is
currently invisible from the repository. Two effects, both already observed:

1. **Work gets planned against namespaces that cannot produce a verdict.** #3919 was opened on the
   premise that labelling `kube-system` would make Kubescape flip nine control-plane mirror pods to
   passing. It cannot — `kube-system` is unscanned — but nothing in the tree said so, and the
   measurement had to be made against the live cluster to find out. #3239 likewise names
   `kube-system` as "the obvious next namespace" for a rollout whose read-back is the posture signal.
2. **An upstream chart bump can silently change what is scanned.** If a future chart version adds a
   namespace to its default exclusions, every posture claim about that namespace quietly becomes
   vacuous, and nothing here fails. A namespace removed from the default is the same problem
   inverted: findings appear that no one expected to own.

This is the same class the repo already guards elsewhere — `guard-pod-security-exception-scope.sh`
exists because two scopes drifted apart in silence, and both directions of that drift read as
compliance. This one has no guard at all, and its input is not even declared.

## Expected behaviour

The set of namespaces Kubescape does not scan is declared in this repository, so it is reviewable,
diffable, and readable without cluster access — and a divergence between the declared set and what
the chart actually applies is a failing check rather than a discovery made months later while
planning unrelated work.

## Acceptance criteria

- [ ] `excludeNamespaces` is set explicitly in the kubescape HelmRelease values, pinned to the
      value currently in effect so the change is behaviour-preserving on merge. Verify the chart's
      values key before setting it — a key that does not reach the rendered ConfigMap would look
      like a fix while changing nothing, which is the failure mode this issue is about.
- [ ] The rendered/applied value is read back and confirmed equal to the pre-change live value.
- [ ] A comment at that key records why the scope is load-bearing: every posture claim is scoped by
      it, and "unscanned" is not "passing".
Split out: the guard that would catch an opt-in inside the unscanned set is #3925, which is blocked
on this issue because it needs the declared set to compare against.

## Size

Small. A few lines once the values key is verified.

Part of #3239

devantler:

## Why

The security posture score only covers part of the cluster, and nothing in this repository said
which part — the skipped-namespace list came from the upstream chart's default, so the scope could
only be learned by reading a live setting on the running cluster. That is how two pieces of security
work came to be planned against a namespace that is never scanned, which from the score alone looks
exactly like a namespace with nothing wrong.

## What

Writes the current list down explicitly, so the scope of every posture claim is visible in the
repository and reviewable in a diff, with a note recording that "not scanned" must never be read as
"passing". The list is exactly what production runs today, verified against both the live cluster
and the chart, so nothing about what gets scanned changes.

Fixes #3923


devantler:

## Evaluation record — exercised, not just reasoned about

This repository's convention for "tried as a user" is the cheapest observation that needs no
cluster: inspect the rendered output of the changed overlay for the intended effect. Here that is
unusually easy to get wrong, because the claim being made is that **nothing changes** — and a
setting that is silently ignored produces exactly that. So the effect was measured in both
directions.

### The pin is behaviour-preserving

| Source | `excludeNamespaces` |
|---|---|
| live prod `configmap/operator` | `kubescape,kube-system,kube-public,kube-node-lease,kubeconfig,gmp-system,gmp-public` |
| chart 1.40.4 default (`helm show values`) | identical |
| this change, rendered | identical |

`helm template` with this repository's own HelmRelease values, before and after, produces a
**byte-identical** operator `config.json` (compared as canonical JSON, since the chart mints
self-signed certs per render and a whole-manifest diff is therefore noise). The changed overlay path
renders the pinned value; no cluster overlay patches these values, so the base is the only source.

### The key is load-bearing, not decorative

The ablation is the part that matters. Rendering a deliberately different value
(`zzz-ablation-namespace`) **does** change the operator `config.json`, which is what proves the key
reaches the rendered config rather than being silently dropped — the exact failure mode that would
make this change look like a fix while fixing nothing.

### Static validation

`ksail workload validate` and `ksail --config ksail.prod.yaml workload validate` both pass — 762 and
51 validated resources respectively, zero failures, `kubescape` validated explicitly in each. Both
overlays build under `kubectl kustomize`. `guard-pod-security-exception-scope.sh` and
`guard-pod-security-exception-documented.sh` both pass unchanged.

### The authorization re-approval

The second commit re-approves the rendered authorization-surface fingerprint, which moved because a
HelmRelease is a controller-RBAC emitter. Its conservation is exact and came from CI, not from me:
the approval-base comparison names **one** moved entry, `HelmRelease kubescape/kubescape`, and no
RBAC, identity, binding, verb, wildcard or AWS-permission source changed. The new aggregate was
printed by CI's checksum-verified kubectl v1.36.2 renderer (job 106055927834); a local render on
v1.36.1 printed the same value, so two independent renderers agree on it.

⚠️ Worth a reviewer's attention: I deliberately did **not** self-attest that re-approval locally. The
runtime declined to let me run the validator over my own edit to an approved security constant, which
is the right call — an agent changing a pinned fingerprint and then reporting its own test green is
self-approval. CI is the authority here, and this repository's own convention already says so: the
local toolchain is refused as unapproved, and the value must come from the required job.

devantler:


@coderabbitai review

PR #3944, opened 2026-09-20
fix(deploy): fence GHCR rollout from autoscaler node deletion

### LINKED ISSUE #3942: fix(deploy): fence GHCR rollout from autoscaler node deletion
## Problem

A production merge-group deployment can select a freshly autoscaled Talos worker for GHCR credential synchronization after Cluster Autoscaler has already begun removing that node. The node disappears while `talosctl patch machineconfig` is in flight, so an otherwise valid pull request is evicted from the merge queue.

## User and product impact

- Unrelated platform changes fail deployment and cannot merge.
- The automatic heal job must run another full production rollout, creating avoidable workload churn and Coroot lifecycle noise.
- The failure is intermittent because it depends on the autoscaler's timing, so retrying without a fix can either hide or reproduce it.

## Live evidence

Merge-group run `35524936547`, deploy job `106116096528`, failed on 2026-09-20:

> Talos node autoscale-cx43-23c57ebb850ee3f8 did not accept the Git/SOPS GHCR registry auth.

The Kubernetes event timeline proves the race:

- 17:18:32 UTC — the node registered.
- 17:29:14 UTC — Cluster Autoscaler emitted `ScaleDown`, cordoned the node, and began removal.
- 17:29:35 UTC — the autoscaler reported the node removed with drain.
- 17:30:19 UTC — the credential bridge's Talos patch failed for that same node.
- 17:30:28 UTC — Kubernetes deleted the Node because it no longer existed at the cloud provider.

The bridge then reported that it preserved a pre-existing cordon. This means it admitted the already-cordoned autoscale node after scale-down had started. PR #3185 intentionally ignores only the advisory `DeletionCandidateOfClusterAutoscaler` taint; this is the later, destructive phase and needs a separate race-safe path.

## Intended outcome

Make GHCR credential synchronization cooperate with Cluster Autoscaler without weakening node identity, ownership, deletion, or drain safety. A node already entering scale-down must not be mutated, and a node claimed by the credential bridge must not be removed underneath the Talos operation. The bridge must re-enumerate or otherwise prove that every surviving node has the desired credential before publishing.

## Acceptance criteria

- [ ] A failing regression test reproduces a node that is cordoned/removed by Cluster Autoscaler between inventory and Talos mutation.
- [ ] The bridge never applies a Talos mutation to a node already in the destructive scale-down phase.
- [ ] A bridge-owned autoscale node is protected from scale-down only for the bounded operation and the protection is removed with ownership; externally owned scheduling intent remains untouched.
- [ ] A disappearing node is handled only with positive identity/deletion evidence, and surviving nodes are re-enumerated before success.
- [ ] Existing tests still prove advisory removal-candidate churn does not block a deploy and real scheduling/deletion changes fail closed.
- [ ] Exact-head non-Copilot review and required CI pass.
- [ ] The change merges through the protected queue and a merge-group production deployment succeeds under live autoscaling.

## Related

- #3188 fixes a distinct parent-Kustomization quiescence timeout.
- #3185 fixed false failures caused by the advisory removal-candidate taint.

devantler:

## Problem

The protected GHCR credential rollout selected an autoscaled node, then the cluster autoscaler cordoned and deleted that node before the bridge reached its Talos mutation. The bridge treated the pre-existing cordon as operator intent and later failed after the node disappeared, blocking the merge-group deployment.

## User and product impact

A safe production rollout could fail even though the cluster remained healthy, leaving the merge queue blocked and requiring a heal deployment. The same race could recur whenever autoscaler scale-down overlaps the credential bridge.

## Intended outcome

Make node selection and mutation ownership atomic with cluster-autoscaler scale-down, reject nodes already leaving service, and recover bridge-owned guards without ever removing an operator-owned guard.

## What changed

- Reject autoscaled nodes that are already cordoned, deleting, or marked for autoscaler removal before any Talos mutation.
- Atomically add the standard scale-down-disabled annotation and a bridge owner annotation when claiming an eligible autoscaled node.
- Preserve pre-existing operator scale-down guards and remove only a guard owned by the current bridge transaction.
- Treat a node that disappears immediately after the owned claim as safely deselected only after an ignore-not-found read and a fresh full inventory prove there is no same-name, same-UID, or same-address replacement.
- Keep every later destructive-edge disappearance fail-closed.
- Reclaim an orphaned bridge-owned autoscaler guard together with its proven pre-mutation drain fence, while deliberately leaving the node cordoned.
- Extend manual fence recovery to test and remove only the exact bridge-owned autoscaler guard in the same CAS patch.

## Why this is safe

Ownership is explicit and node-scoped. Every release tests node identity, resource version, drain owner, scale-down owner, and exact guard value before one atomic patch. Changed, mismatched, operator-owned, or ambiguous state fails closed. No global autoscaler setting is weakened.

## Validation

- RED: focused autoscaled-node tests initially showed the claim omitted the owned scale-down guard and did not reject a pre-cordoned target.
- GREEN: focused autoscaler, orphan-reclaim, and fence-report regression tests.
- GREEN: complete refresh-flux-ghcr-auth Go integration suite on the rebased exact head (207.673 seconds).
- GREEN: standalone GHCR safety regression suite.
- GREEN: ShellCheck with sourced-file following, Bash syntax checks, and git diff checks.

## Live evidence

Failed protected run 35524936547 cordoned and removed autoscale-cx43-23c57ebb850ee3f8 between inventory/claim and the Talos patch. The bridge logged that it preserved the pre-existing cordon. Deployment readback after merge will verify the protected queue, Flux convergence, and healthy node state.

Closes #3942


devantler:

@coderabbitai review

Please review exact head 573d3f30a1fd6e243dec2700cb3d313e59853d9e, focusing on fail-closed node identity handling, atomic cluster-autoscaler guard ownership, preservation of pre-existing operator guards, orphan recovery, and destructive-edge behavior. Do not review an earlier head.

devantler:

@coderabbitai review

Please review exact head 3e10869e482bb6b0fa2e00cc69d513cf6a10c884. The prior P1 is addressed by recording scale-down guard provenance at claim time, carrying it through durable bootstrap journals, requiring the exact bridge owner and true value at every destructive edge and release, and rejecting owner-only removal before any Talos mutation. The new regression leaves the node cordoned and proves no Talos or root mutation occurs. Do not review an earlier head.

devantler:

@coderabbitai review

Please review exact head 15d16fdcb5eec775275395ec709c42e340bdc360. The remaining P1 is addressed by passing guard provenance into the final fence transition and atomically testing the exact bridge scale-down owner plus scale-down-disabled=true in the same JSON patch that marks the fence mutating. A new regression removes the owner inside that transition and proves the patch fails before Talos, drain, reboot, uncordon, or root mutation. The complete integration suite, standalone safety suite, ShellCheck, Bash syntax, and diff checks pass. Do not review an earlier head.

PR #3948, opened 2026-09-20
fix(coroot): expose clean-state counters

### LINKED ISSUE #3947: Expose exact Coroot clean-state counters in scheduled job logs
## Problem

Coroot remediation can be deployed and validated from OIDC-only Kubernetes reads, but the exact Alerts and Applications legend counters are only visible through the authenticated browser UI. That makes post-merge verification depend on desktop browser automation, which is disruptive on the maintainer workstation and not suitable for the scheduled verifier.

## Outcome

Emit one machine-readable clean-state summary from the existing Coroot alert autosuppressor after its exact, declarative suppressions run. The summary must use Coroot's own v1.22.2 API responses so it matches the UI:

- Alerts: active Critical and Warning counts from response context.
- Applications: counts of critical (SLO violation), warning, info (errors in logs), unknown (integration required), and OK statuses from the applications overview.
- Keep every current exemption exact and fail-visible; this is evidence only, not a broader suppression or weakened check.

## Acceptance criteria

- A non-vacuous test fails before implementation and covers the exact JSON counter mapping.
- The scheduled job logs a stable JSON object suitable for OIDC-only readback.
- Any API/query/schema failure fails the job instead of reporting a false zero.
- Existing autosuppressor matching and negative controls remain green.
- Delivery uses exact-head non-Copilot review, protected CI/merge queue, and production readback.

devantler:
## Why

Coroot’s exact alert and application health counters could only be confirmed through the desktop browser, making production verification disruptive and unsuitable for unattended runs.

## What

The existing scheduled Coroot reconciliation now emits the same counters as structured read-only evidence and fails closed if Coroot returns an unexpected response.

Fixes #3947


devantler:

@coderabbitai review

Please review exact head `ae7662c33c511b51389956a88c976389404b1155`, focusing on fail-closed Coroot response validation, exact parity with the Alerts and Applications UI counter semantics, stable structured output, and preservation of every existing suppression boundary. Do not review an earlier head.

devantler:

@coderabbitai review

Please review exact head `e45f632168d30a053f6f6f96b8f1172981b79941`. Confirm the P1 boolean-counter finding is fixed, missing zero-valued keys still match Coroot’s UI semantics, malformed counters fail closed, and existing suppression boundaries remain unchanged. Do not review an earlier head.

devantler:

@coderabbitai review

Please review exact head `96128ffe255096e49232ec947e228d3bcc2c2d58`. Confirm the remaining P1 is fixed: present alert counters must be non-negative integers, while absent zero-valued keys still map to zero; malformed values and unknown application statuses fail closed. Confirm existing suppression boundaries remain unchanged. Do not review an earlier head.

Issue #3918, opened 2026-09-20
security(longhorn): clear the pre-opt-in C-0211 residual in longhorn-system without detaching volumes unintentionally


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

`longhorn-system` is opted in to `pod-security.devantler.tech/baseline-context`, and the mutation
works — but 26 of its 41 pods still carry neither `fsGroupChangePolicy` nor `seLinuxOptions`.
Measured against live prod specs 2026-09-20, the split is temporal and completely clean:

- every **clean** pod was created on or after `2026-09-06T20:46:02Z`
- every **residual** pod was created on or before `2026-08-26T13:14:47Z`

#3636 landed the opt-in on 2026-09-06. The rule matches `CREATE` only, so this is the documented
behaviour working as designed: pods created after the opt-in carry both fields, and pods created
before it keep their original spec until something recreates them.

Nothing in `longhorn-system` is going to recreate them on its own:

```
12x DaemonSet/engine-image-ei-*        three engine versions, oldest pod 2026-06-16
 4x DaemonSet/longhorn-csi-plugin
 4x InstanceManager/instance-manager-*
 6x ReplicaSet/csi-{attacher,provisioner,resizer}
```

Longhorn keeps an `engine-image` DaemonSet per engine version for as long as any volume references
that version, and an `InstanceManager` per node for as long as volumes are attached there.

## Audience and problem

This is the last chunk of the C-0211 residual in an **already opted-in** namespace, so it will not
close by itself and it is invisible in the rollout state — the namespace looks done.

It is deliberately **not** folded into #3239's per-namespace opt-in work, because the risk class is
different. Opting a namespace in is a posture change with no runtime effect on running pods. Clearing
this residual means **recreating storage-plane pods**, and recreating an `InstanceManager` requires
detaching the volumes it serves. That is a storage availability decision, and it should not ride
along inside a security sweep.

## Expected behaviour

Every pod in `longhorn-system` carries both fields, reached by a sequenced roll that never detaches a
volume without intent.

## Acceptance criteria

- [ ] The three workload classes are sequenced by disruption, least first: the CSI sidecar
      ReplicaSets and `longhorn-csi-plugin` DaemonSet roll normally; the `engine-image` DaemonSets
      are checked for whether any volume still references each engine version before touching them;
      the `InstanceManager` pods are handled last and explicitly, against attachment state.
[...]

PR #3950, opened 2026-09-20
fix(coroot): accept autoscaled Longhorn manager risk

devantler:
## Why

The autoscaler added a healthy worker, which caused Longhorn to create a new node-scoped instance manager. Coroot correctly left its generated application identity visible for review, but the resulting Availability warning is not actionable: each node intentionally has one manager, while volume durability comes from replicas across nodes.

## What

Add the exact generated manager identity to the reviewed GitOps acceptance list with the existing recovery rationale. The behavioral contract now requires all 37 unique, reasoned declarations and proves this new identity is present; undeclared risks still remain visible.

Fixes #3949

PR #3953, opened 2026-09-20
security(longhorn): retrofit C-0211 onto the four CSI sidecar Deployments

### LINKED ISSUE #3946: security(longhorn): clear the C-0211 residual on the four CSI sidecar Deployments

## Evidence

Measured 2026-09-20T21:20Z against live PROD stored controller specs — the surface Kubescape reads. `longhorn-system` is opted in to `pod-security.devantler.tech/baseline-context` and the mutation works, but 8 of its 11 controllers still carry neither `fsGroupChangePolicy` nor per-container `seLinuxOptions`, because the rules match `CREATE` only and these controllers predate the opt-in (#3636, 2026-09-06).

Four of the eight are CSI sidecar Deployments:

| Deployment | `fsGroupChangePolicy` | containers missing `seLinuxOptions` |
|---|---|---|
| `csi-attacher` | MISSING | 1 / 1 |
| `csi-provisioner` | MISSING | 1 / 1 |
| `csi-resizer` | MISSING | 1 / 1 |
| `csi-snapshotter` | MISSING | 1 / 1 |

## Audience and problem

#3918 sequences this namespace's residual into three classes by disruption and requires the least disruptive to go first. These four are that class: they are stateless CSI controller sidecars that roll normally, and unlike the `engine-image` DaemonSets and the `InstanceManager` pods they need **no** volume-reference or attachment-state reasoning before being touched. Bundling them with the storage-plane work would put a safe change behind an availability decision.

Nothing recreates them on their own, so this does not close by itself and the namespace reads as done while 8 workloads still fail.

## Hypothesis

Bringing the four stored Deployment specs under the existing baseline-context mutation closes 4 of the namespace's 8 residual workloads with no volume detached and no storage-plane pod recreated. Because the rules are `CREATE`-only, reaching an already-created controller needs a `mutate-existing`-shaped rule (the repo already uses that pattern in `label-existing-baseline-workers.yaml`) rather than a pod restart — restarting yields clean pods but leaves the stored spec dirty, which is the surface Kubescape scores.

## Success signal

- **Baseline:** 4 of 4 missing `fsGroupChangePolicy`; 4 of 4 with every container missing `seLinuxOptions`.
- **Target:** 0 of 4 on both, measured from stored specs.
- **Guardrail:** no change to privilege, user or group fields on any of the four; the other 7 controllers in the namespace unchanged; no pod outside these four Deployments recreated; no Longhorn volume detached or degraded.

## Smallest useful change

A rule scoped to these four Deployments' namespace and kind that supplies only `fsGroupChangePolicy` and container-level `seLinuxOptions` — the same two non-privilege fields #3239 established, and nothing else.

## Acceptance criteria

- [ ] The four Deployments' **stored** specs carry `fsGroupChangePolicy` and container-level `seLinuxOptions`.
- [ ] Verified with the discriminating predicate: `fsGroupChangePolicy` present **AND** every container has `securityContext.seLinuxOptions`. Do **not** test `seLinuxOptions.level` (never set; every object is `{}`) and do **not** test at pod level (absent on all 11 controllers by design, #3469) — either reports every workload failing and never registers progress.
- [ ] No privilege, user or group field changed on any of the four, checked before/after.
- [ ] The `engine-image` DaemonSets, `longhorn-csi-plugin`, `longhorn-manager`, `longhorn-ui` and `longhorn-driver-deployer` are untouched, and no Longhorn volume is left detached or degraded.
- [ ] Both overlays build (`kubectl kustomize k8s/clusters/local/` and `.../prod/`).

## Size

Small. The care is in scoping the rule so it reaches exactly these four and does not roll the storage-plane DaemonSets in the same change.

## Not in scope

The `engine-image` DaemonSets, `longhorn-csi-plugin` and the `InstanceManager` pods — they stay on #3918, which carries the volume-reference and attachment-state sequencing. The `kube-system` opt-in stays on #3239.

Part of #3918

devantler:

## Why

Four of Longhorn's storage sidecars fail the cluster's pod-hardening check and
cannot fix themselves: the opt-in that hardens workloads only applies when
something rewrites them, and nothing ever rewrites these four. They have sat
failing since the namespace was opted in and would stay that way indefinitely,
and restarting them does not help because the check reads the stored definition
rather than the running pod.

## What

Applies the two hardening settings directly to those four existing sidecars,
with a permission scoped to exactly them; each restarts once as it picks the
settings up. Nothing else in the namespace is touched — the storage components
that need careful sequencing before a restart are deliberately left for #3918,
and no privilege, user or group setting is changed.

Fixes #3946
Part of #3918


devantler:


@coderabbitai review

devantler:


@coderabbitai full review

devantler:


@coderabbitai full review

devantler:

**Readiness record for `df414f6ccca218c06c03041aa23d4c6cd005ac32`.** All three conditions hold; promoting.

**1 — Programmatically tested.** `scripts/tests/test-csi-sidecar-baseline-context.sh` drives the real `kyverno` binary at **1.19.1**, the version the cluster runs, over committed trigger and target fixtures, and asserts the mutated output rather than the policy text. RED→GREEN plus two ablations, each of which failed with its own assertion rather than an error: removing `resourceNames` from the mutation Role, and widening the rule's match to `longhorn-manager`.

I verified the test **actually executes in CI at this head**, rather than inferring it from a green tick. Run [35542234368](https://github.com/devantler-tech/platform/actions/runs/35542234368), job `🧪 Validate Manifests`, step `⚖️ Validate cluster policies` — its log carries both `shellcheck scripts/tests/test-csi-sidecar-baseline-context.sh` and `bash scripts/tests/test-csi-sidecar-baseline-context.sh`, and the policy validated twice in the manifest sweep. That check matters because the path filter and the job step are separate edits, and a PR can go green with the new test never having run.

**2 — Reviewed.** CodeRabbit full review at this exact head: requested 23:33:25Z, `✅ Action performed / Full review finished` 23:38:14Z, head status `Review completed` 23:38:13Z. The auto-generated summary was refreshed at 23:38:11Z, names the full head, and carries no rate-limit marker. Zero review objects, zero inline comments, zero unresolved threads, and the only collapsed section is `🔇 Additional comments (7)` — CodeRabbit's non-actionable one. No finding to fix or refute.

**3 — Tried and evaluated as its user.** The effect was observed, not reasoned about. `kyverno apply --resource <trigger> --target-resource <targets> --output` emits one document per (rule × target) that actually changed, so a target a rule left alone emits nothing — the four CSI sidecar Deployments come back carrying `fsGroupChangePolicy: OnRootMismatch` and a container-level `seLinuxOptions: {}`, and the storage-plane workloads held out of the targets emit nothing at all. That is what drove the design away from the obvious implementation: expressed as `patchStrategicMerge`, `+(seLinuxOptions): {}` left the container with **no securityContext whatsoever** while the sibling `fsGroupChangePolicy` in the same patch landed — strategic merge drops an empty mapping, and C-0211 is satisfied precisely by an empty one. The container half is `patchesJson6902` because that was measured, not assumed.

Judged as the operator who has to live with it: the blast radius is the four named Deployments, pinned by static `mutate.targets` and by a `resourceNames`-pinned Role, so the background controller cannot reach anything else even if the match were wrong; each rolls once on the pod-template hash change; and `failurePolicy: Ignore` means a policy fault cannot wedge admission.

**Merge mechanics for whoever lands this.** `platform` has an active `Require merge queue` ruleset, so this enqueues rather than merges — drop `--squash`, keep `--match-head-commit`. Eviction risk was pre-assessed: the base is one commit behind `main` (`fix(coroot): accept autoscaled Longhorn manager risk #3950`), touching only a coroot risk-acceptance ConfigMap and its test, so the `merge_group` re-run of `🔐 Validate Production Authorization` has no reason to move the fingerprint re-approved on this branch.

Issue #3955, opened 2026-09-20
fix(kubescape): preserve large vulnerability scan writes


devantler:
## Problem

Coroot reports a real `storage` application error for the Backstage vulnerability manifest:

- `Create - save object failed`
- `write metadata: insert metadata: sqlite: clear bindings: interrupted`
- key `/spdx.softwarecomposition.kubescape.io/vulnerabilitymanifests/kubescape/ghcr.io-backstage-backstage-1.52.0-993611`

The exact manifest is absent afterward, so this is data loss and must not be exempted. The same scan also submits 6,604,113-6,604,551 byte `openvulnerabilityexchangecontainers` payloads, which the current 1,000,000 byte kind limit rejects.

## Root cause evidence

The compatibility image sets SQLite `busy_timeout` to 60 seconds, but the pinned v0.0.297 storage code obtains every REST connection with `pool.Take(poolCtx)` where `poolCtx` expires after five seconds. zombiezen/sqlite binds that context to the connection interrupt for its entire checked-out lifetime. A legitimate foreground write waiting longer than five seconds is therefore forcibly interrupted before the reviewed busy timeout can help, after the payload file has been written but before metadata is persisted.

## Acceptance criteria

- Add a failing regression test proving the pool-acquisition timeout cannot interrupt an already acquired connection during a legitimate longer operation.
- Separate the five-second pool-acquisition budget from a finite, longer checked-out connection lifetime without unsafe manual `SetInterrupt` rebinding.
- Apply the fix consistently to the v0.0.297 compatibility image and publish/pin a signed immutable revision.
- Raise only `storage.kindQueues.openvulnerabilityexchangecontainers.maxObjectSize` to a finite value above the observed 6.605 MB payload with documented headroom.
- Keep unrelated storage kinds and failure modes fail-visible.
- Verify the Backstage vulnerability manifest and VEX resource are persisted on a fresh scan, storage logs remain clean, and Coroot reports no storage log error after the historical entry ages out.

PR #3957, opened 2026-09-20
fix(kubescape): preserve large vulnerability scan writes

devantler:
## Problem

Kubescape accepted a 6,604,551-byte Backstage vulnerability exchange record, but its one-megabyte queue limit rejected the record and the storage API later lost the corresponding vulnerability manifest. The storage failure was `sqlite: clear bindings: interrupted`: the five-second pool-acquisition context stayed attached to the checked-out SQLite connection, so a legitimate writer could be interrupted before the configured 60-second busy timeout completed.

## User and product impact

The security inventory can silently lose large vulnerability results while Coroot intermittently reports both storage errors and queue warnings. That makes the live application health view noisy and, more importantly, leaves the vulnerability manifest incomplete.

## Intended outcome

- Limit pool acquisition to five seconds without letting that deadline interrupt a checked-out connection.
- Keep every checked-out connection finitely bounded for at least the full SQLite busy-timeout window.
- Admit the observed VEX record with a narrow eight-megabyte ceiling on that exact object kind.
- Publish a reviewed `.6` compatibility image from the pinned upstream source. A follow-up PR will pin the immutable digest after the main-branch publisher creates it.

## Validation

- RED: `TestCreatePoolAcquireTimeoutDoesNotInterruptCheckedOutConnection` reproduced `sqlite: clear bindings: interrupted` with the existing patch.
- GREEN: the patch applies cleanly to upstream commit `b35788b68337134fc2514574cde1ba7f1225fd43`; `go test ./pkg/registry/file` passes from that clean source.
- `bash scripts/tests/test-kubescape-storage-hotfix.sh`
- `bash scripts/tests/test-coroot-alert-remediation.sh`
- `kubectl kustomize k8s/bases/infrastructure/controllers/kubescape`

Part of #3955


devantler:
Manual non-Copilot current-head review completed for 5b82b77d24fe8f0907700680f0a02486e3675430. Reviewed pool acquisition cancellation, checked-out connection lifetime, cleanup ordering, the RED-to-GREEN SQLite writer reproduction, every storage entry point migrated to the helper, the finite 8 MB VEX ceiling, immutable publisher revisioning, and the authorization delta. No findings. This review requested no Copilot or external reviewer.

PR #3963, opened 2026-09-21
fix(kyverno): serialize CSI baseline retrofit

### LINKED ISSUE #3962: fix(kyverno): serialize Longhorn CSI baseline retrofit

## Why

The Longhorn CSI baseline retrofit writes each of four Deployments from two separate Kyverno mutate-existing rules. On the first production reconciliation, those rules raced on the same resource versions and emitted 30 PolicyError warnings even though all four Deployments eventually converged.

## Outcome

Apply both baseline fields in one serialized mutate-existing rule, retain the same four-name scope and least-privilege Role, prove near misses remain untouched, and verify a clean production rollout with no new PolicyError events.

## Evidence

At 2026-09-20T23:49Z the policy emitted repeated object-modified conflicts for csi-attacher, csi-provisioner, csi-resizer, and csi-snapshotter. Every target is now healthy and carries both fields, which isolates the defect to concurrent idempotent policy writes rather than an unavailable workload.

devantler:

## Why

The Longhorn CSI hardening rollout used two Kyverno rules to update the same four Deployments. Those writers raced on the first production pass and emitted 30 object-modified PolicyError warnings even though the workloads eventually became healthy.

## What

Applies both baseline fields through one serialized rule and keeps the same four exact target names and least-privilege Role. No Coroot exemption is added; the historical rollout events can age out while future policy reconciliations stay clean.

## Verification

- RED: the previous two-rule policy fails the new single-writer contract
- GREEN: the exact policy test produces one combined mutation containing both fields
- Near misses, inherited settings, existing security fields, and the namespaced grant remain covered
- shellcheck and the infrastructure render pass

Fixes #3962


devantler:
Manual non-Copilot current-head review completed for 6d5500a9496677b428eca0e061fda2fec76bdff8. Reviewed the single-writer mutate-existing shape, one-element target loop, parent-versus-leaf JSON Patch preconditions, exact four-target scope, preservation of existing security fields, combined-mutation behavioral fixture, and unchanged least-privilege Role. No findings. This review requested no Copilot or external reviewer.

PR #3972, opened 2026-09-21
fix(observability): classify recovered autoscaler lease retry

### LINKED ISSUE #3971: Accept recovered autoscaler lease retry in Coroot
## Problem

Coroot reports the healthy cluster autoscaler as having an error after a recovered control-plane leader transition. The autoscaler retries its lease update through the slow path and continues normally, but the application-health reconciler only recognizes the existing capacity-buffer conflict pair.

This leaves an intermittent error in the Applications view even though the workload remains ready and later autoscaler loops succeed.

## Intended outcome

Treat only the exact, recovered autoscaler lease-update retry as reviewed operational noise, with the same finite cap and fail-visible behavior used by the existing classifiers. Any different lock, error text, excessive count, or mixed unknown error must remain visible.

## Acceptance criteria

- A failing behavioral fixture covers the exact recovered lease retry before implementation.
- The classifier accepts the exact `kube-system/cluster-autoscaler` slow-path retry caused by the known etcd leader timeout.
- Near misses, different locks, permission failures, excessive counts, and mixed unknown errors remain visible.
- Existing capacity-buffer conflict handling and all application-health reconciliation tests remain green.
- The change is deployed through the protected merge queue and verified in live Coroot.

## Evidence

At 2026-09-21 01:39:10 UTC the autoscaler logged one `Failed to update lease optimistically, falling back to slow path` event with `etcdserver: request timed out, possibly due to previous leader failure` and lock `kube-system/cluster-autoscaler`. The pod remained Ready with zero restarts and subsequent scaling loops continued successfully.

devantler:
## Why

Coroot intermittently reports the healthy cluster autoscaler as erroneous after a recovered control-plane leader transition. The stale signal obscures new application failures even though the autoscaler remains ready and resumes normal work.

## What

Recognize only the exact recovered autoscaler lease retry under a finite ten-event allowance. Wrong locks, permission failures, mixed errors, and higher counts remain visible.

Fixes #3971


devantler:
Manual non-Copilot exact-head review completed for `b26eefa3bc9f2f5edadbe497c06a40915e9475ed` against base `762224e0cc339ea1c33e888a3f56f87fe315cf53`.

I reviewed the combined classifier rather than either accepted shape in isolation: the existing CapacityBuffer pair still requires a same-timestamp two-record correlation, while the new lease path requires the exact leader-election source, recovered previous-leader timeout, and `kube-system/cluster-autoscaler` lock. All returned errors must still be reviewed; a wrong lock, permission failure, or mixed unknown record removes the override, and the ten-event ceiling remains finite so an excessive recurrence is visible.

The behavior contract was observed RED before implementation and is GREEN with the exact, near-miss, mixed, and excessive-count fixtures. The focused suite, shellcheck, rendered infrastructure build, and diff check pass; GitHub verifies the commit signature. Verdict: no blocking findings. This review requested no GitHub Copilot or external reviewer.

PR #3982, opened 2026-09-21
fix(kubescape): bound host-data cleanup API calls

### LINKED ISSUE #3979: fix(kubescape): bound host-data cleanup API calls
## Problem

The `kubescape-hostdata-cleanup` CronJob hit its five-minute active deadline on 2026-09-21 at 03:36Z. Pod events show it scheduled, pulled, created, and started immediately, then ran continuously until the Job controller killed it. The script wraps `kubectl` in retries but gives each API call no request timeout, so one stalled list or delete can consume the entire Job deadline and erase the pod before useful completion evidence is retained.

The current live inventory is small (97 host-data objects across the ten resource types) and the previous daily run completed in 32 seconds, so object volume is not the cause.

## User impact

A transient API stall turns a safe cleanup into an opaque `DeadlineExceeded` warning, leaves stale Kubescape host data unreconciled for a day, and removes the pod before the failure path can report which authoritative read stalled.

## Intended outcome

Bound every cleanup `kubectl` call with an explicit request timeout while preserving the existing retry and fail-closed behavior. The Job must fail with a useful command-level error before the five-minute outer deadline when the API remains unavailable, and transient stalls must still have enough retries to recover.

## Acceptance criteria

- [ ] Every cleanup API call has an explicit bounded request timeout.
- [ ] The worst-case retry budget for one failed authoritative call remains below `activeDeadlineSeconds`.
- [ ] Behavioral contract coverage fails before the manifest change and passes after it.
- [ ] The live CronJob deploys the bounded request policy and the next run completes without `DeadlineExceeded`.
- [ ] Exact-head non-Copilot review, protected merge queue, CI, and Flux deployment readback are complete.

devantler:

## Why

The daily Kubescape host-data cleanup reached its five-minute outer deadline even though only 97 inventory objects existed and the previous run finished in 32 seconds. Its retry loop bounded retry count but not an individual Kubernetes API request, so one stalled call could consume the whole Job budget; Kubernetes then deleted the pod before the fail-closed path could report which authoritative inventory was unavailable.

## What

Give every cleanup `kubectl` request an exact 20-second timeout while retaining five attempts and the existing linear backoff. The full 120-second retry budget stays below the 300-second Job deadline, so a transient API stall can recover and a persistent one fails visibly with diagnostic output before Kubernetes terminates the pod. The cleanup remains fail-closed and deletion logic is unchanged.

## Validation

- RED: the focused contract rejected an unbounded Kubernetes API call.
- GREEN: the contract proves the exact request timeout and verifies the outer deadline exceeds the complete retry budget.
- Existing behavioral fixtures still prove missing authoritative inventories prevent deletion and healthy mode deletes only records whose identically named Node is absent.
- `shellcheck scripts/tests/test-kubescape-hostdata-cleanup.sh`
- `bash scripts/tests/test-kubescape-hostdata-cleanup.sh`
- production authorization validator passes against exact main `098d1719d1538e7d4f6dea56dab6cd7e5547bdbc`
- `git diff --check`

Fixes #3979


devantler:

Exact-head non-Copilot review of `2a09dc81470b4158501a0c7d1c1080421223d01f` (GitHub signature: verified).

No findings. The exact 20-second per-request timeout keeps all five retries plus backoff inside the existing 300-second Job deadline, preserves fail-closed authoritative inventory acquisition, and does not widen deletion scope or authorization. Focused shellcheck, behavioral tests, full Hetzner infrastructure render, production authorization validation against exact main, and diff check pass.

PR #3987, opened 2026-09-21
fix(ci): reacquire Flux parent before restart

### LINKED ISSUE #3986: fix(ci): reacquire Flux parent fence before controller restart

## Problem

Merge-group run `35561812989` for PR #3982 passed every static and production-authorization gate, then failed before publishing a candidate with:

```
The Flux policy fences changed before the kustomize-controller handoff restart.
```

The candidate was never published and the automatic heal correctly began redeploying the last known-good `main`, so production remained on the previous Ready OCI revision.

The transaction deliberately fences the Flux Operator-owned root `flux-system` Kustomization and the child `infrastructure` Kustomization. The script already handles the Operator restoring the exact same-UID, ownerless, unsuspended root object (a) between parent and child acquisition and (b) after the controller restart. It does not handle the same safe state transition after the child fence is acquired but before the restart begins. At that boundary it collapses the expected Operator restoration, an unrecognized fence mutation, and an API read failure into one generic error and evicts an otherwise valid PR.

## User impact

- A safe, expected Flux Operator reconciliation can evict any unrelated PR from the protected merge queue.
- Coroot remediations #3976 and #3982 cannot be delivered until the shared deployment lane is reliable.
- Requeueing without repairing the omitted race would repeat production deployment work and generate more transient Coroot rollout evidence.

## Intended outcome

At the pre-restart boundary, accept only the exact already-supported parent release state: same UID, no owner annotation, and `spec.suspend=false`, while the exact child fence and synchronization Lease remain owned. Reacquire the parent atomically before restarting the controller. Continue to fail closed for a foreign owner, UID change, malformed state, child-fence change, or unreadable API response.

## Acceptance criteria

- [ ] A failing fixture reproduces the Flux Operator restoring the parent after child acquisition and before the controller restart.
- [ ] The transaction reacquires only the exact released parent state and completes the safe handoff.
- [ ] Tests prove foreign, malformed, replaced, and unreadable fence states still block mutation.
- [ ] Failure diagnostics identify whether the parent, child, or read failed instead of emitting only the combined message.
- [ ] Focused safety tests and repository CI pass.
- [ ] The change receives exact-head non-Copilot review and merges through the protected queue.
- [ ] A real production deployment succeeds before the Coroot remediations are requeued.

devantler:

## Why

A fully green merge-group for #3982 was evicted before publication because the Flux Operator restored its generated root Kustomization after the child policy fence was acquired but before the kustomize-controller restart. The deploy script already repaired this exact safe state transition at the adjacent boundaries, but treated it as an unknown fence change at this one boundary. That made an expected Operator reconciliation look like a PR failure and blocked the Coroot cleanup work behind it.

## What

When the exact child fence and synchronization Lease are still owned, recognize only the same root Kustomization UID in the exact released state (no transaction owner and not suspended), then reacquire that parent fence atomically before the controller restart. Foreign ownership, replacement, malformed state, and API read failure still stop before any restart or policy mutation. Split diagnostics now identify whether the parent, child, or read failed.

## Validation

- RED: the new pre-restart Operator race fixture failed with the production error, `The Flux policy fences changed before the kustomize-controller handoff restart.`
- GREEN: the race now reacquires the parent between child acquisition and controller restart, then completes the transaction.
- Fail-closed table tests reject a foreign owner, replacement UID, malformed parent, and unreadable fence state before restart or policy mutation.
- Adjacent before-child and after-restart reacquisition tests pass.
- `go test ./scripts/tests/refresh-flux-ghcr-auth -count=1`
- `go test ./scripts/tests/refresh-flux-ghcr-auth -run TestFluxParentFence -count=1`
- `shellcheck -x scripts/refresh-flux-ghcr-auth.sh scripts/ghcr-auth-lib.sh scripts/refresh-flux-ghcr-auth-safety.sh`
- `bash -n scripts/refresh-flux-ghcr-auth.sh`
- `git diff --check`

Fixes #3986


devantler:

Exact-head non-Copilot review completed for `0534cecab836b7fc64dbf3551af57b77901cb670` (GitHub signature: verified).

Reviewed the fence state machine, cleanup ownership transitions, diagnostics, and all new behavioral fixtures. The new path is bounded to the already-supported exact released-parent state while the synchronization Lease and child fence remain owned. It clears local parent ownership before reacquisition so a failed claim cannot release another actor's fence, then re-proves both fences before the controller restart. Foreign ownership, UID replacement, malformed state, and unreadable objects remain fail-closed before restart or policy mutation.

Validation reviewed:
- RED reproduced the live pre-restart error before implementation.
- Complete `refresh-flux-ghcr-auth` Go safety suite passed at this exact head.
- Source-aware ShellCheck, Bash syntax, and diff checks passed.

No blocking findings.

PR #3991, opened 2026-09-21
fix(coroot): bound reviewed OOM signals

### LINKED ISSUE #3985: fix(coroot): bound one Talos runtime child-process OOM
## Problem

Coroot intermittently reports the Talos host application `_:Unknown:runtime` as warning after one OOM-victim event in the runtime cgroup, even when the container runtime itself stays healthy.

Live evidence on 2026-09-21 showed the event on `prod-worker-2` while the Talos runtime PID remained `2807` before and after it, the node stayed Ready without MemoryPressure, and total node use remained around 7.5 GiB of 16 GiB. Coroot node-agent 1.35.8 increments the container OOM counter for any process exit marked by the kernel `oom/mark_victim` tracepoint in that cgroup; Coroot 1.22.2 then presents the event as an application restart. The evidence therefore supports a child process being killed inside `/talos/runtime`, not containerd restarting or node memory exhaustion.

## User impact

The Applications view carries an intermittent false runtime warning, obscuring a repeated runtime OOM or an actual node-pressure incident that would require action.

## Intended outcome

Accept one process-level OOM-victim event per hour only for the exact Talos `_:Unknown:runtime` application, with an explicit reason tied to Coroot's process-level cgroup semantics. Keep the allowance finite so a second event still warns, and leave every Kubernetes workload and other host application at the global zero threshold.

## Acceptance criteria

- [ ] A failing contract proves the runtime policy is absent.
- [ ] The reconciler applies `MemoryOOM=1` only to `_:Unknown:runtime`.
- [ ] Tests prove the global threshold remains zero and a second hourly event remains visible.
- [ ] The policy reason documents the process-level cgroup semantics and healthy runtime/node evidence.
- [ ] Focused tests and repository CI pass.
- [ ] The change merges through the protected queue and deploys through Flux.
- [ ] Coroot shows no runtime warning while repeated OOMs remain actionable.

### LINKED ISSUE #3988: bug: bound one Longhorn engine-image probe OOM

## Problem

Coroot opened an `Out of memory kills` warning for the exact Longhorn application `95rsc5yp:longhorn-system:DaemonSet:engine-image-ei-a4d05f02` at 2026-09-21 05:15:40Z. The signal is real, but it did not restart or degrade the engine-image container.

## User and platform impact

The retained single OOM event keeps the Applications and Alerts views non-clean and can obscure a genuine recurrence. Weakening the global OOM check would hide actionable failures elsewhere, so any acceptance must be exact, finite, and documented.

## Root-cause evidence

- Coroot attributes the single event to `engine-image-ei-a4d05f02-zd4xs` on `prod-worker-2`.
- The Coroot memory timeline stays roughly flat around the application's low steady-state usage; node unreclaimable memory stays near 52% with no pressure spike.
- The Pod remains Ready. Its container start time is 2026-08-26 12:17:56Z, and its last recorded container termination is also from 2026-08-26, so the 2026-09-21 event did not terminate the main container.
- The engine-image Pod's main process is `bash ... && sleep infinity`; its recurring child processes are the readiness/liveness `/data/longhorn version --client-only` probes. The evidence therefore isolates the OOM victim to a short-lived probe child rather than the storage daemon container.
- All seven nodes are Ready without memory pressure, and Kubernetes recorded no event for the Pod.

## Acceptance criteria

- Add a declarative Coroot `MemoryOOM` threshold of exactly `1` only for `95rsc5yp:longhorn-system:DaemonSet:engine-image-ei-a4d05f02`.
- Record the operational reason: one retained probe-child OOM is tolerated because the Ready main container was not restarted or degraded.
- Prove behaviorally that the first retained event is accepted and a second event still warns.
- Do not weaken the global OOM threshold or any other application.
- Deploy through the protected merge queue and verify the live Coroot alert resolves while the workload and cluster remain healthy.

### LINKED ISSUE #3989: fix(coroot): give the policy reconciler burst memory headroom

## Problem

At 2026-09-21 05:43:40Z Coroot recorded an OOM kill in the exact periodic application `95rsc5yp:observability:CronJob:coroot-application-health-reconciler` on `prod-worker-3`.

## User and platform impact

The reconciler is the declarative owner of application-health policies. An OOM during its bounded parallel work can leave a run incomplete, generate a warning about the health controller itself, and delay application-policy convergence.

## Root-cause evidence

- The container has an explicit 64 MiB memory limit and 32 MiB request, below the namespace's 512 MiB default limit.
- The script deliberately runs four reconciliations concurrently so it can finish within its CronJob deadline. Each worker can hold Coroot JSON plus `jq` and `curl` child processes, including bounded raw-log responses of up to 1,000 entries.
- Coroot's timeline shows low steady-state RSS but a single OOM event at the 64 MiB ceiling, consistent with a short parallel parsing burst rather than a leak.
- `prod-worker-3` remained Ready without MemoryPressure, so this was a container ceiling, not node exhaustion.
- Subsequent runs completed, confirming the workload logic is healthy but its burst headroom is insufficient.

## Acceptance criteria

- Raise only the reconciler container's declarative memory limit from 64 MiB to 128 MiB; preserve the reviewed four-worker concurrency and bounded-response checks.
- Add a non-vacuous manifest contract that fails at the old 64 MiB limit and passes at 128 MiB.
- Do not suppress or exempt the reconciler OOM check.
- Deploy through the protected merge queue and verify subsequent Jobs complete without additional OOM events while the cluster remains healthy.

### LINKED ISSUE #3992: fix(coroot): classify fail-closed truncation without self-alerting

## Problem

The production application-health reconciler deliberately fails closed when Coroot returns exactly the 1,000-message cap for a raw log query: it restores the target application's `LogErrors` threshold to inherited/global visibility. That safety path is functioning, but it also emits its own status line at `error` severity.

The Talos `_:Unknown:init` stream currently reaches that cap every minute, so the healthy reconciler continuously creates an `Errors in logs` signal for itself even though no reconciliation failed and the uncertain target remains fail-visible.

Observed at `2026-09-21T06:15:14Z`:

`refusing talos-runtime: raw-message response may be truncated for 95rsc5yp:_:Unknown:init`

The same run immediately confirmed `talos-runtime-truncated-fail-visible ... /LogErrors=inherit` and completed successfully.

## Expected outcome

Classify only this bounded-cap fail-closed outcome as informational status while preserving the inherited/global `LogErrors` threshold on the queried target. Query failures, malformed responses, unexpected global configuration, and actual reconciliation failures must remain error-level and fail the existing contracts.

## Acceptance criteria

- A response at the raw-message limit still posts `configs[2] = null` for the exact target.
- The truncation status is not emitted at error severity.
- Query failures, invalid responses, and reconciliation failures remain error-level.
- A behavioral RED→GREEN fixture covers both the log classification and retained fail-visible target policy.

devantler:

## Why

Coroot warned that the Talos `runtime` application restarted after one OOM event on `prod-worker-2`, but the live evidence showed a different outcome: containerd kept PID `2807`, the node stayed Ready without MemoryPressure, and total use remained around 7.5 GiB of 16 GiB. Coroot node-agent counts any process marked as an OOM victim inside the runtime cgroup, so one short-lived child process can trigger the application-level check even when the runtime itself remains healthy.

The same process-level behavior later appeared in the exact Longhorn `engine-image-ei-a4d05f02` cgroup. Coroot attributed one event to `engine-image-ei-a4d05f02-zd4xs`, but the Ready container's start time and Kubernetes restart count did not change, the application's memory stayed flat, and `prod-worker-2` remained near 52% unreclaimable memory without pressure. The sleeping installer container's only recurring children are its readiness/liveness `longhorn version --client-only` probes.

A separate OOM at 2026-09-21 05:43:40Z exposed a real resource ceiling in the application-health reconciler itself. Its steady-state RSS is small, but four bounded workers can concurrently hold Coroot responses plus `jq` and `curl` children. The explicit 64 MiB limit was reached on a healthy node; later runs completed, ruling out a leak or node-pressure failure.

After the first policy rollout, the reconciler also exposed a recurring secondary signal: the Talos `init` stream reaches the bounded 1,000-message query cap, so the reconciler correctly restores the target's inherited/global `LogErrors` policy but then logs that deliberate fail-closed outcome as its own error every minute. No reconciliation failed, and the uncertain target already remains visible.

## What

Set `MemoryOOM` to a finite one-event-per-hour threshold only for the exact `_:Unknown:runtime` host application and exact `longhorn-system:DaemonSet:engine-image-ei-a4d05f02` application. The global threshold remains zero, every other workload and host process remains unchanged, and a second event in either exact cgroup in the same hour still warns.

Each declarative reason records the process-level cgroup semantics and the verified healthy main process/node state; these are not broad OOM suppressions.

Raise only the application-health reconciler's memory limit from 64 MiB to 128 MiB while keeping its 32 MiB request, four-worker bound, response caps, and one-minute deadline unchanged. This fixes the real reconciler OOM at source; it does not exempt that application from OOM warnings.

Classify only the bounded-cap fail-closed outcome as informational reconciler status while retaining the inherited/global `LogErrors` threshold on the queried target. Query failures, malformed responses, unexpected global configuration, and actual reconciliation failures remain error-level.

## Validation

- RED: each behavioral contract failed because its exact application policy was absent.
- GREEN: the contract verifies both exact applications, global zero, and independent finite thresholds of one.
- RED→GREEN: the manifest contract rejects the old 64 MiB ceiling and requires the exact 32 MiB request / 128 MiB limit pair.
- RED→GREEN: a capped raw-message response still fails visible on the target but no longer self-alerts as a reconciler error.
- `rtk test ./scripts/tests/test-coroot-application-health-reconciler.sh`
- `shellcheck scripts/tests/test-coroot-application-health-reconciler.sh`
- `yq -r '.spec.jobTemplate.spec.template.spec.containers[0].command[2]' k8s/providers/hetzner/infrastructure/coroot/cron-job-application-health-reconciler.yaml | shellcheck -s sh -`
- `kubectl kustomize k8s/providers/docker/infrastructure`
- `git diff --check`

Fixes #3985
Fixes #3988
Fixes #3989
Fixes #3992


devantler:

Exact-head review of `286109ecd21edbdc9046d71122e5b8c6e07c1b80`: no blocking findings.

- The two exemptions preserve the global `MemoryOOM` threshold of zero and are restricted to the exact Talos runtime identity and the exact Longhorn engine-image DaemonSet. Each permits one retained process-level event per hour, so a recurrence still surfaces.
- The Longhorn identity includes the immutable engine-image name. A future engine upgrade therefore drops the exemption instead of silently inheriting it.
- The reconciler OOM is fixed rather than exempted: only its limit changes from 64 MiB to 128 MiB; the 32 MiB request, four-worker bound, response caps, schedule, concurrency policy, and deadline remain unchanged.
- The behavioral fixture asserts the exact application IDs and thresholds, the exact request/limit pair, and the emitted Coroot API payloads. The focused suite, embedded-shell and test-shell ShellCheck, both provider renders, and `git diff --check` pass on this head.

The change is declarative, finite, and fails visible if the evidence or workload identity changes.

devantler:

Exact-head review of `86967883550f569456b12e9e9a629b95204af206`: no blocking findings.

- The OOM policies still preserve global zero, exact workload identities, and finite one-event application thresholds; the reconciler OOM is still fixed with bounded resource headroom rather than exempted.
- The added truncation change does not weaken the inspected application's check. At the 1,000-message cap it still writes `configs[2] = null`, restoring inherited/global `LogErrors` visibility for the target.
- Only the reconciler's classification of that successful fail-closed decision changes from `error` to `info`. Query failures, malformed responses, unexpected globals, worker failures, and failed reconciliations retain error severity and their existing failure paths.
- The new behavioral fixture demonstrated RED on the prior head, then GREEN while independently asserting both non-error self-classification and retained target visibility. The focused suite, both ShellCheck passes, Bash syntax, provider renders, and diff check pass.

This head remains declarative, finite, and fail-visible where evidence is incomplete.

PR #3995, opened 2026-09-21
fix(coroot): prune stale reconciler failure history

### LINKED ISSUE #3994: fix(coroot): prune stale reconciler failure history

## Problem

The application-health reconciler recovered after the bad candidate and every current minute run succeeds, but Coroot still reports `BackoffLimitExceeded (1 events in the last 1h)` almost two hours after the alert opened. Kubernetes has no live Warning event for the CronJob.

The CronJob retains three failed children (`failedJobsHistoryLimit: 3`). Those are the three Jobs created by the rejected candidate at 05:33Z–05:35Z. Because no later run fails, the history set never rotates; Coroot keeps rediscovering the stale failed Job state and the alert cannot age out.

## Expected outcome

Do not retain failed child objects for this every-minute, externally observed policy reconciler. The real failure remains visible through the Kubernetes event and centralized Coroot logs during its evidence window, while successful current runs and the CronJob itself remain available.

## Acceptance criteria

- Set `failedJobsHistoryLimit: 0` only on `coroot-application-health-reconciler`.
- Retain `successfulJobsHistoryLimit: 1`, `backoffLimit: 0`, the one-minute schedule, concurrency guard, and deadline.
- Add a RED→GREEN manifest contract for the exact history policy.
- Verify existing stale failed children are pruned by the CronJob controller and the Coroot alert ages out without an exemption.

devantler:

## Why

The application-health reconciler recovered after the rejected candidate and every current minute run succeeds, but Coroot continued reporting `BackoffLimitExceeded` almost two hours after the alert opened. The live cluster has no Warning event for the CronJob.

The CronJob retained exactly three failed child Jobs because `failedJobsHistoryLimit` was three. Those were the three failures created by the rejected candidate; healthy later runs never rotate failed history, so Coroot kept rediscovering a stale failed state and the warning could not age out.

## What

Set `failedJobsHistoryLimit: 0` only for the every-minute application-health reconciler. Actual failures remain observable through Kubernetes events and centralized Coroot logs during their evidence window, while the stale failed child object is no longer an indefinite alert source.

Retain one successful child for live readback, `backoffLimit: 0`, the one-minute schedule, the concurrency guard, and the bounded deadline.

## Validation

- RED: the behavioral contract failed while the reconciler retained three failed children.
- GREEN: the contract requires the exact `failedJobsHistoryLimit: 0`, `successfulJobsHistoryLimit: 1`, and `backoffLimit: 0` combination.
- `rtk test ./scripts/tests/test-coroot-application-health-reconciler.sh`
- `shellcheck scripts/tests/test-coroot-application-health-reconciler.sh`
- embedded reconciler shell: `shellcheck -s sh -`
- `kubectl kustomize k8s/providers/docker/infrastructure`
- `git diff --check`

Fixes #3994


devantler:

Closing without merge after live verification disproved the implementation premise. Production currently has zero Kubernetes Warning Event objects, while Coroot still shows the earlier BackoffLimitExceeded alert because the live Kubernetes-events rule uses a 1h query window plus `Keep firing for: 1h`. Coroot upstream resolves event alerts only after the event leaves the query window and that keep-firing interval expires; retained Job objects are not part of the evaluation. Setting `failedJobsHistoryLimit: 0` would therefore weaken post-failure diagnostics without fixing the alert. The warning will be verified through its intended aging window instead.

PR #3997, opened 2026-09-21
fix(flux): stabilize HA controller rollouts

### LINKED ISSUE #3996: fix(flux): stabilize HA controller handoff rollouts
## Problem

Coroot intermittently reports `kustomize-controller` as a single-instance availability risk during production deploy handoffs even though the Flux controller Deployment is configured for two replicas with required cross-node anti-affinity. The same rollout produced a Kubernetes `Unhealthy` readiness-probe event after a terminating pod became unreachable.

## Root cause evidence

At 2026-09-21T07:44Z the production Deployment was 2/2 Ready before and after the rollout, but Coroot observed a transient single-instance state. The old ReplicaSet carried `kubectl.kubernetes.io/restartedAt: 2026-09-21T06:53:20Z`; the desired ReplicaSet did not. The policy-handoff script deliberately adds that restart annotation to replace in-flight controller processes, and the Flux operator subsequently reconciles the imperative annotation away. That creates a second rapid cleanup rollout. With the default `minReadySeconds: 0`, Kubernetes retires the old replica immediately after the replacement first reports Ready, which is shorter than Coroot's observation window and leaves no stability floor for the HA handoff.

## Intended outcome

Strengthen the declarative HA rollout contract for the replicated Flux controllers without suppressing or dismissing a potentially real future single-replica failure.

## Acceptance criteria

- [ ] A focused behavioral test fails against the current production FluxInstance and passes after the fix.
- [ ] `kustomize-controller`, `helm-controller`, and `notification-controller` render with `spec.minReadySeconds: 30`.
- [ ] Their existing two replicas and required pod anti-affinity remain intact.
- [ ] `source-controller` remains intentionally single-replica and unchanged.
- [ ] The policy handoff remains bounded by its existing two-minute rollout timeout.
- [ ] CI and exact-head non-Copilot review pass.
- [ ] The change merges through the protected merge path, deploys, and production reads back 2/2 healthy replicas.
- [ ] Coroot Alerts, Applications, and Risks return to zero active signals across repeated samples, with no check weakening.

devantler:

## Why

Coroot intermittently reports a Flux controller as a single-instance risk during normal production deploy handoffs, even though two healthy replicas exist before and after the rollout. The gap also creates noisy readiness events and makes a real loss of redundancy harder to distinguish.

## What

Require every replicated Flux controller replacement to remain healthy for 30 seconds before an old replica retires. This strengthens rollout availability without dismissing the risk or weakening monitoring, while leaving the intentionally single-instance source controller unchanged.

Fixes #3996


devantler:

## Exact-head review — `563ceaa2860f4e7e8b73251657b3a34747f5cc35`

Reviewer: OpenAI Codex (non-Copilot). GitHub Copilot was not requested or used.

I reviewed the complete diff against `main`, the embedded JSON patches, rollout timing, and the focused behavioral contract. Each replicated controller receives exactly one 30-second readiness floor while retaining two replicas and required cross-node anti-affinity; `source-controller` remains outside replica/readiness-floor patching; the two staged readiness waits plus the existing probe and shutdown grace remain bounded by the handoff timeout; and the change strengthens availability without suppressing Coroot evidence.

One over-broad test assertion found during review was fixed before this head: the contract now forbids only replica/readiness-floor operations for `source-controller`, rather than forbidding all future name-targeted patches. Focused RED→GREEN validation, local and production schema validation, and all 23 current-head CI checks pass.

Verdict: no remaining findings at this exact head.

PR #3999, opened 2026-09-21
security(longhorn): retrofit C-0211 onto the longhorn-csi-plugin DaemonSet

### LINKED ISSUE #3998: security(longhorn): retrofit C-0211 onto the longhorn-csi-plugin DaemonSet

## Evidence

Measured against live prod on 2026-09-21: the `longhorn-csi-plugin` DaemonSet's stored template carries an empty pod `securityContext` and no container `seLinuxOptions`. It was created on 2026-08-21, before the longhorn-system baseline-context opt-in (2026-09-06). Its managed fields show no template writer since, so the CREATE/UPDATE opt-in rules never reach it. It is one of the four DaemonSets left in #3918's residual. The other three are the engine-image DaemonSets, which #3966 / #3967 handle.

## Audience and problem

Whoever reads the platform's C-0211 posture. This DaemonSet keeps a namespace that has already opted in looking unfinished, and it will not change on its own.

## Expected behaviour

The stored DaemonSet template carries `fsGroupChangePolicy` and container-level `seLinuxOptions`, and no existing security field changes.

## Acceptance criteria

- [ ] An exact, add-only mutate-existing retrofit reaches only `longhorn-system/longhorn-csi-plugin`. It reuses the existing resourceNames-pinned grant and adds no new RBAC.
- [ ] A regression test pins scope, shape, convergence and the grant, with ablations recorded.
- [ ] After deploy, the live stored spec is read back: both fields present, and the privileged containers unchanged.
- [ ] No volume is left detached or degraded after the one-node-at-a-time roll.

Size: S. Part of #3918.

devantler:

## Why

The Longhorn storage node plugin was created before its namespace opted in to the baseline security context, and nothing ever rewrites it. So it stays one of the four remaining workloads that the posture scan flags for this control in the storage namespace.

## What

Adds the same targeted, add-only retrofit the CSI sidecars received: the plugin gains the two non-privilege settings, and every existing setting is kept. It reuses the narrow write permission that already exists for this one workload, and it rolls the plugin one node at a time without detaching any volume.

Fixes #3998
Part of #3918


devantler:

## Validation at `d4cb7db0`

- **Live shape measured first** (prod, 2026-09-21): `longhorn-csi-plugin` has an empty pod `securityContext`. `node-driver-registrar` is `privileged`. `longhorn-liveness-probe` has no `securityContext`. `longhorn-csi-plugin` is `privileged` with `allowPrivilegeEscalation` and `SYS_ADMIN`. The update strategy is RollingUpdate with `maxUnavailable: 1`. The fixture reproduces that shape exactly.
- **GREEN:** `scripts/tests/test-longhorn-csi-plugin-baseline-context.sh` passes with kyverno 1.19.1, the cluster's version, and shellcheck is clean. The mutated output has `fsGroupChangePolicy: OnRootMismatch` on the pod, an empty `seLinuxOptions` on all three containers, and every privileged field preserved. Both near misses (same name in another namespace, and an engine-image DaemonSet) emit no document. A second pass over the already-retrofitted shape patches nothing.
- **Ablations**, each failing on its own assertion:
  - existing-container branch neutralised → `container 0 must receive seLinuxOptions`
  - pod-leaf branch neutralised → `the empty pod securityContext must receive fsGroupChangePolicy`
  - match widened to `*` → `the rule must match only longhorn-csi-plugin`
  - absent-container branch made unconditional → `the registrar must keep its existing securityContext — the object must never be replaced`. This one shows the wholesale-replace hazard is caught.
- **Test method note:** the policy has one target, and the Kyverno CLI rejects an object that is both trigger and target. The test therefore evaluates a copy of the policy with only the match name rewritten to a fixture trigger, and it asserts that the copy differs from the shipped policy in its `match` block only.
- **RBAC:** no new grant. The existing `kyverno:background-controller:mutate-longhorn-csi-plugin` Role (daemonsets, `resourceNames: [longhorn-csi-plugin]`, get/patch/update) already covers it, and the test pins it. The ClusterPolicy is outside the EKS authorization aggregate, as it was for #3953.
- **Still to do after merge:** read the stored spec back from prod, and confirm no Longhorn volume is degraded after the roll (#3998 AC3/AC4).

devantler:


@coderabbitai full review

devantler:


@codex review

devantler:

## Ready at `d4cb7db0`

1. **Programmatically tested:** the regression test passes, with four ablations each failing its own assertion. All 24 CI checks pass, including the EKS authorization aggregate validator, which is unchanged because no RBAC changed. There are no threads and no findings.
2. **Reviewed:** Codex clean review at `d4cb7db0`. CodeRabbit was asked first and refused: this repository's hourly quota was already used by #3984.
3. **Tried as a user:** I evaluated the policy with kyverno 1.19.1, the cluster's version, against the DaemonSet shape measured live from prod (details in the validation comment). The mutated result carries exactly the two non-privilege fields, every privileged field is intact, near misses are untouched, and a second pass is a no-op. This change can't be exercised on a cluster before it deploys. After merge, #3998 AC3/AC4 require the stored spec to be read back from prod and every Longhorn volume to be confirmed healthy after the one-node-at-a-time roll.

Issue #3990, opened 2026-09-21
security(kubescape): close the Coroot C-0211 residual through a scoped, convergence-proven admission mutation


devantler:
> 🤖 Generated by the Agentic Engineer

## Evidence

Six of the 14 remaining C-0211 workloads are the Coroot stack in `observability`: the server, prometheus, cluster-agent, node-agent, clickhouse and keeper. All six are templates generated by the Coroot operator. The installed Coroot CRD has no pod-level security field for any component and no field at all for node-agent, so the fix can't be supplied through operator configuration the way it was for the Helm-rendered workloads. Measurement and CRD survey: https://github.com/devantler-tech/platform/issues/3239#issuecomment-5756108492.

## Audience and problem

Until these six stored templates carry `fsGroupChangePolicy` and `seLinuxOptions`, the cluster-wide C-0211 exception can't narrow for observability. `docs/baseline-context-rollout.md` keeps the namespace-wide `baseline-context-controllers` switch off because admitting fields the operator didn't set could make it rewrite its objects repeatedly.

## Hypothesis

Scoping the existing controller mutation to Coroot-owned controllers (not the whole namespace), and gating it on a preflight / after-apply convergence proof, closes all six without an operator write loop. The operator's current write pattern is bounded: its last writes on these objects are days old despite continuous reconciliation.

## Success signal

- The observability residual goes from 6 to 0 under the object-presence predicate, measured on the live stored specs.
- Guardrail: the operator's write count on the six objects doesn't grow across a 30-minute window after apply, and every Coroot pod stays Ready.

## Acceptance criteria

- [ ] The mutation reaches only Coroot-owned Deployments, StatefulSets and DaemonSets, not other observability workloads.
- [ ] A test proves it applies with the scope label present and doesn't apply without it.
- [ ] A before-publish and after-reconcile guard, like the Longhorn UI canary's, checks the stored fields and flags an owner that keeps rewriting.
- [ ] The residual is re-measured after rollout and recorded on #3239.

Size: M.

devantler:
> 🤖 Generated by the Agentic Engineer

## Verified in prod: the observability residual is 0, with no write loop

I measured the six Coroot templates' live stored specs (read-only, 2026-09-22), using the same object-presence predicate as #3239.

[...]

PR #4007, opened 2026-09-21
feat(kyverno): scope the controller baseline-context rules to Coroot-generated workloads

### LINKED ISSUE #4006: security(kubescape): scope the controller baseline-context rules to Coroot-generated workloads, default off

## Evidence

The six C-0211 residual workloads in `observability` are the Coroot operator's own templates, and all six carry `app.kubernetes.io/managed-by: coroot-operator` (read live 2026-09-21). The rest of that namespace (the operator itself, the audit-log forwarder, the crossplane sync exporter) is Helm- or Git-managed and doesn't carry it. The existing controller rules can only be switched on for a whole namespace.

## Audience and problem

#3990 needs a mutation that reaches only the Coroot-generated controllers. Without a narrower scope, activation would also roll every Helm and Git workload in the namespace.

## Change

Add a separate, default-off `baseline-context-coroot` namespace label that makes the existing controller rules match only Deployments, StatefulSets and DaemonSets carrying the Coroot managed-by label. Nothing carries the new label, so this changes nothing in the cluster.

## Acceptance criteria

- [ ] The rules match an operator-generated Deployment and DaemonSet in a scoped namespace, and keep privilege fields untouched.
- [ ] They don't match a non-operator workload in the scoped namespace, or an operator workload in an unscoped one.
- [ ] An inventory pin fails if any namespace carries the new label without a test change.

Size: S. Part of #3990.

devantler:

## Why

Six Coroot monitoring workloads still miss two security settings the posture scanner checks, and the only switch that could add them would also restart every other workload in their namespace.

## What

Adds a separate, switched-off setting that applies those two security settings only to the workloads the Coroot operator creates. Nothing uses it yet, so production is unchanged; turning it on comes later with a check that the operator does not keep rewriting its workloads.

Fixes #4006
Part of #3990


devantler:

Evaluation at `51eaeece`:
- **Scope, read live (read-only):** across the whole cluster, the object selector `app.kubernetes.io/managed-by=coroot-operator` matches exactly the six residual workloads named on #3990: `coroot-cluster-agent`, `coroot-prometheus`, `coroot-clickhouse-keeper`, `coroot-clickhouse-shard-0`, `coroot-coroot` and `coroot-node-agent`. No namespace carries `baseline-context-coroot`, so merging changes nothing in production.
- **Tests:** `test-add-baseline-context.sh` passes, and `kyverno test` reports 45/45.
- **Ablations**, each run against a backup copy of the policy and each failing exactly its own assertions:
  1. The original policy fails the 5 ON-state checks.
  2. Removing the object selector fails the non-operator-workload check.
  3. Removing the namespace scope fails the unscoped-namespace check.
  4. Labelling `observability` fails the inventory pin.
- Both cluster overlays still build.

devantler:


@coderabbitai full review

PR #4016, opened 2026-09-21
feat(security): guard Coroot baseline-context convergence on every deploy

### LINKED ISSUE #4014: security(kubescape): guard Coroot baseline-context convergence on every deploy

## Evidence

#3990's acceptance criterion 3 asks for a before-publish and after-reconcile guard, like the Longhorn UI canary's, that checks the stored fields on the six Coroot-operator templates and flags an owner that keeps rewriting them. The scoped admission rules shipped in #4007; the namespace label that activates them cannot land safely without this guard.

## Audience and problem

The platform operator activating the Coroot C-0211 fix. Without a guard, an operator/admission write loop would only be noticed after it had been churning the monitoring stack.

## Hypothesis

A read-only guard wired into every production deploy, disarmed while the label is absent, is enough to make activation a reversible, evidenced step.

## Success signal

- The guard reports the live population correctly (six templates, operator-owned, ready) and fails after-reconcile today for the expected reason, since the fields are not yet stored.
- Guardrail: with the label absent the guard never contacts the cluster.

## Acceptance criteria

- [ ] The guard disarms without a cluster read while the label is absent, and refuses an unexpected label value.
- [ ] While armed, it refuses a population other than the reviewed six, a template not controller-owned by `Coroot`, and an unready template; an API error is never an empty population.
- [ ] After reconcile it requires both fields on all six and fails on a changed generation/UID, a removed field, lost readiness, or repeated writes to one template.
- [ ] Fixture tests cover each case, with ablations proving each check is load-bearing, and CI runs them.

Size: S.

devantler:

## Why

Six Coroot monitoring workloads still fail a security posture check. They can only be fixed by letting admission add two settings to templates the Coroot operator generates, and that is unsafe to switch on until something proves the operator does not keep rewriting them in a loop.

## What

Adds that proof to every production deploy: while the Coroot opt-in label is set, the deploy checks that all six workloads carry both settings, stay healthy, and are not being rewritten, and fails otherwise. The label is still off, so nothing changes today; switching it on is the next change.

Fixes #4014


devantler:

**Evaluation at `ed2c3f45`** (read-only against production, `oidc@prod`, 2026-09-21 ~12:08Z):

- `before-publish` with the repository's real namespace manifest (label absent) → `rollout_required=false`, no cluster read.
- `before-publish` with a manifest carrying the label → `rollout_required=true`: live population is exactly the six reviewed templates, each controller-owned by `Coroot` and fully ready.
- `after-reconcile` → fails with `the six templates did not all reach both fields and a complete healthy rollout`, which is the correct verdict today: the fields are not stored yet, so activation cannot pass this guard without the per-workload template update.

Fixture test: 25 cases pass; six ablations (loop count, population, field strip, generation/UID, owner kind, owner controller flag) each turn it red for the intended case.

devantler:


@coderabbitai full review

devantler:

**Readiness at `ed2c3f45`: tried as a user, reviewed, CI green.**

I ran the guard at this head read-only against production, in the deploy step's `before-publish` mode:

| Namespace manifest | Result |
|---|---|
| as committed (label absent) | `rollout_required=false`, exit 0. The guard is disarmed and reads nothing from the cluster. |
| copy with `baseline-context-coroot: enabled` | `rollout_required=true`, exit 0. It read the live Coroot templates and confirmed the reviewed six are present, operator-owned and ready. |

The `after-reconcile` mode was exercised live on the previous run. With the label armed but no mutation applied yet, it fails because the fields are not stored, which is the expected refusal. The fixture suite covers the remaining failure paths.

- **Review:** CodeRabbit full review at this head. The summary names `ed2c3f45…` with no actionable comments, the status reads `Review completed`, and there are no threads.
- **CI:** green, and the PR is `CLEAN`.

Merging keeps the guard disarmed. Activation (the label, the per-workload template update and DR wiring) is the next slice of #3990.

PR #4020, opened 2026-09-21
fix(security): narrow the app tenants' publish matchers to their approved revisions

### LINKED ISSUE #3917: security(supply-chain): the app tenants' publish matchers still accept any workflow revision

## Evidence

Measured on `main` at `cf2f6248`, after #3308 landed the generated approved-revision sets:

| consumer | workflow | matcher subject revision |
|---|---|---|
| `.github` | `publish-manifests` | `@(a147ead8…\|3c3a6348…)` — approved pair |
| `aws` | `publish-manifests` | `@(b089a1b0…\|3c3a6348…)` — approved pair |
| `wedding-app` | `publish-app` | `@[0-9a-f]{40}` — **any** fixed revision |
| `ascoachingogvaner` | `publish-app` | `@[0-9a-f]{40}` — **any** fixed revision |

`scripts/publish-workflow-approved-revisions.tsv` already computes an approved pair for both app
tenants, and both are currently diverged — `wedding-app` runs an artifact signed by `883d891a…` while
its pin is `759ff9b2…`, and `ascoachingogvaner` runs one signed by `a147ead8…` against a pin of
`3c3a6348…`. The numbers exist; these two matchers do not read them.

The exclusion is deliberate, not an oversight: `TRUSTED_RELEASE_STREAM_CONSUMERS` in
`scripts/publish-workflow-approved-revisions.lib.sh` names both tenants, and
`write-publish-workflow-matchers.sh` leaves their subject on the shape pattern.

## Problem

For these two consumers, any 40-hex revision of `devantler-tech/actions`' `publish-app.yaml` still
satisfies verification — including a superseded one. That is the substitution the #2818 line of work
exists to close, and it remains open for the two workloads that serve real users.

## The trade-off, which is why this is not simply "narrow them too"

The recorded rationale is sound. An application tenant releases on its own cadence. If its matcher named
only the revisions the platform has observed, then the first tenant release signed by a *new*
`publish-app.yaml` revision would fail verification and stop deploying until a platform change landed —
turning an unrelated change in `devantler-tech/actions` into a tenant outage.

So the question is not whether to narrow, but **how to narrow without coupling a tenant release to a
platform merge**. Options worth weighing, cheapest first:

1. **Regenerate on the event rather than the schedule.** The set already regenerates automatically; if
   that workflow also triggered on a new `publish-app.yaml` revision in `devantler-tech/actions`, the
   approved pair would move ahead of the first tenant release that uses it, and the coupling window
   would close on its own.
2. **Accept a bounded window.** Narrow, and treat the gap between a new workflow revision and the
   regeneration PR as a known, short unavailability — acceptable only if it is measured, not assumed.
3. **Keep the carve-out and compensate elsewhere.** Leave the shape pin and rely on a different control
   to catch a superseded signer, stating plainly which control that is.

## Expected

Either both app tenants verify against their generated approved pair, or the carve-out is a decision
someone made on measured evidence rather than a gap that outlived the issue that would have caught it.

## Acceptance criteria

- [ ] The coupling window is **measured**, not estimated: how often `publish-app.yaml` changes revision,
      and how long the regeneration workflow takes to reflect one.
- [ ] A decision is recorded on this issue — narrow, narrow-with-event-trigger, or keep the carve-out —
      with that measurement behind it.
- [ ] If narrowing: both tenants' matchers carry the approved pair, and a negative case proves an
      out-of-set revision is rejected for them specifically.
- [ ] If keeping the carve-out: `TRUSTED_RELEASE_STREAM_CONSUMERS`' comment names the compensating
      control, so the next reader is not left to infer that none exists.

## Size

Small for the measurement and the decision; small-to-medium for whichever remedy it selects.

Follows #3308.

devantler:

## Why

The two apps that serve real users, the wedding site and the coaching site, still accepted a release signed by any version of the shared publishing workflow, including superseded ones. Every other consumer is already tied to the versions the platform has approved, so these two were the gap.

## What

Both apps now accept only their approved versions: the one that signed the running release, the one they are pinned to, and the newest published release. This is now safe because each app already refuses to move its pin until the platform has approved the new version. A new version therefore holds the update rather than breaking a release.

Fixes #3917


devantler:

**Tried as a user, at `acede63a`.**

- **Writer on the real tree:** `write-publish-workflow-matchers.sh` changed exactly two manifests, the wedding-app and ascoachingogvaner OCIRepositories. Each moved from `@[0-9a-f]{40}` to its three approved revisions. The enforcing guard (`APPROVED_REVISIONS_ENFORCE=1`) passes on all 4 consumers, and the shape guard passes on all 7 shared-workflow subjects.
- **The running releases stay accepted.** Read-only against prod, Flux serves `wedding-app@sha256:4f5eb100…` and `ascoachingogvaner@sha256:1016e41f…`. Both digests equal the approved set's recorded `applied_digest`, and each recorded signer (`759ff9b2`, `a147ead8`) is in that tenant's new subject.
- **Not observed directly:** a direct cosign verification of the signatures. The local token lacks package read, and I did not widen it.

**Tests.**
- RED on the previous scripts: 13 guard cases and the writer suite fail.
- GREEN now: all guard and writer cases, plus the regeneration (1 control + 45 ablations), shape-guard, signing-revision and digest-gate suites.
- Removing the grouping rule fails only the ungrouped-alternation case.

devantler:

**Update at `dd42bce3`.**

`🔐 Validate Production Authorization` failed at `acede63a` with `unapproved rendered authorization surface fingerprint`. The authorization fingerprint maps an exact publish-workflow revision set back to the approved `[0-9a-f]{40}` pattern, so regenerating the set never needs a re-approval. But it only recognised up to **two** revisions, and a tenant's set now has three.

The root-cause fix is in the projection, not a fingerprint refresh:
- It now recognises up to **3** revisions for publish-app consumers (signer, pin, release candidate). publish-manifests consumers stay at 2.
- A 3-revision set is still a strict subset of the pattern the authorization approval already covers.
- The projection test adds the 3-way subset for both tenants, and a set one larger than each workflow carries still moves the fingerprint. All 253 validator tests pass, including `TestValidateAuthorizationAcceptsCommittedPolicy`, the test that failed in CI.

`🧪 Validate Manifests` failed on an upstream `curl … 504` while downloading a supply-chain tool, which is transient. This push re-runs it.

devantler:


@coderabbitai full review

devantler:

Resolution record for the outside-diff finding in https://github.com/devantler-tech/platform/pull/4020#pullrequestreview-5269789688 (obsolete matcher comments, both app manifests): **fixed in 849b6bdc**. The comments no longer describe the former any-commit matcher or its accepted residual risk; they now say the subject is written from the approved set and accepts only the approved revisions (signer, current pin, release candidate). Comment-only change: the rendered authorization aggregate is unchanged (committed approval test passes locally), and the approved-revisions guard, writer test and guard test all pass.

devantler:


@coderabbitai full review

devantler:


CodeRabbit refused this head (rate limited); Codex is out of usage, so the next lane is Cursor Bugbot.

devantler:

devantler:
@cursor review

devantler:

The push to `849b6bdc` changes only the two manifests' comments (the CodeRabbit outside-diff finding); the matchers, and so the behaviour evaluated above at `dd42bce3`, are byte-identical, and the approved-revisions guard still reports all four consumers in agreement. Bugbot is green at this head.

PR #4021, opened 2026-09-21
fix(kubescape-bridge): preview finding titles exactly as they are filed

### LINKED ISSUE #2924: kubescape-backlog-bridge: report() renders titles unsanitized and unbounded, so the preview differs from what write files

## Evidence

Found while doing the user evaluation for #2917 — running the real command against live cluster data,
not by reading the code.

The two modes render the title through different paths:

| path | call site | sanitized? | bounded? |
|---|---|---|---|
| **write** (what gets filed) | `issues.go:395` → `renderTitle(t)` | ✅ | ✅ 256 runes |
| **report** (the preview) | `main.go:1210` → `t.Title()` raw | ❌ | ❌ |

Reproduced against a real `workloadconfigurationscansummaries` object from `admin@prod`, with its
control key replaced by a 1400-character id:

```
rendered title runes = 1432
ends with ellipsis   = False
```

The same theme through the write path renders 256 runes ending in `…`.

Two distinct consequences, both small but real:

1. **The preview does not preview.** `-mode report` is how an operator sees what a write run would
   file. A title that differs between the two means the report is not a faithful dry run.
2. **The report's TSV can be broken by untrusted input.** `report()` emits tab-separated fields, and
   `sanitizeForIssue` is precisely what folds `\r`/`\n` into spaces. Because report bypasses it, a
   control key or CVE key containing a tab or newline would split a row and corrupt the output for
   anything parsing it. Scanner output is untrusted by this tool's own stated design premise, so
   "control ids are always `C-0034`-shaped" is an assumption the rest of the package deliberately
   refuses to make.

Not a security issue: `report()` writes to an `io.Writer` (stdout), never to a GitHub artifact, so
the unsanitized `@`-mention and backtick handling that matters for bodies does not apply here.

## Expected behaviour

Report and write render a theme's title identically, through one function.

## Smallest useful change

Have `report()` call `renderTitle(t)` instead of `t.Title()`. Both live in the same package.

Worth deciding deliberately rather than assuming: bounding the *preview* is arguably wrong if an
operator wants to see the full key. If so, the fix is to sanitize in report (closing consequence 2)
and to display the truncation explicitly, rather than silently rendering a different title from the
one that would be filed.

## Acceptance criteria

- [ ] A theme's title is rendered by one function for both modes, or the divergence is documented
      with its rationale.
- [ ] A control/CVE key containing a tab or newline cannot break the report's TSV row structure;
      proved by a test with that input.
- [ ] The report of a theme whose title would be truncated on filing makes that visible.

## Rough size

XS. Pre-existing — **not** introduced by #2917, which bounded only the write path because that is
where the failure it fixed occurs (an over-long title fails `gh issue create` and, via `applyPlan`'s
stop-at-first-failure, blocks every remaining action in the run).

devantler:

## Why

The security-backlog bridge's preview mode is how an operator checks what it would file before letting it write issues. It printed finding titles raw, so an oversized or oddly formatted scanner value previewed a title that would never actually be filed, and could break one preview row into two.

## What

The preview now shows each title exactly as it would be filed: cleaned up and trimmed to GitHub's title limit.

Fixes #2924


devantler:

**Tried as a user** at the head of this PR, against real production scanner data (read-only `GET` of one `workloadconfigurationscansummary`, `external-secrets/deployment-external-secrets-webhook`, failing C-0234):

- Unmodified object, `-mode report`: exit 0, one row, title `security(posture): C-0234 fails` — unchanged from before.
- Same object with the control key (and its matching `controlID`) stretched to 1406 characters plus an embedded newline, `-mode report`: exit 0, still **one** row with 6 tab-separated fields, title bounded to 256 runes ending `…c8a204c8` — the same form the write path files.

**RED/GREEN:** the new test fails on `main` (the newline splits the row in two) and passes here; the full package suite passes (282).

devantler:


@coderabbitai full review

devantler:


CodeRabbit refused this head (rate limited, 15:45Z) and Codex is at its usage limit, so this round advances to Bugbot.

devantler:
@cursor review

devantler:

**Readiness at `e15c183`:** all checks green; Cursor Bugbot review green at this head with no findings (CodeRabbit refused this round on its hourly limit and Codex is at its usage limit); 0 review threads; tried as a user against a real production posture summary (evaluation comment above). Promoting and adding to the merge queue at this head.

PR #4022, opened 2026-09-21
fix(kubescape-bridge): show scanner severity as literal text in filed issues

devantler:
> 🤖 Generated by the Agentic Engineer

## Why

Issues filed by the security-backlog bridge show the scanner's severity in normal text, so a formatting character in that value (underscores, tildes) still renders as live styling. Earlier fixes blocked such characters one at a time, and each review round found another.

## What

The severity is now shown as plain literal text, the same way the affected components already are, which makes every formatting character harmless at once.

Fixes #2928

PR #4024, opened 2026-09-21
fix(ci): retry brief download failures when installing supply-chain tools

### LINKED ISSUE #4023: Retry transient release-download failures when installing supply-chain tools

## Evidence

On 2026-09-21 between ~14:50Z and ~15:21Z, `🧪 Validate Manifests` failed on platform#4020 and platform#4022 (and again on #4022's re-run) in `🔐 Exercise verified supply-chain tools`, with `curl: (22) The requested URL returned error: 504` while downloading a pinned release asset from GitHub. Neither PR touches that step; a sibling PR from the same base passed it minutes earlier. The same 504 class hit an unrelated golangci-lint install in `devantler-tech/actions` in the same window.

Every platform PR's required manifest validation, and the merge-queue production deploy, which installs the same tools before signing. One momentary 5xx from GitHub's release download fails the job outright, because both downloads run once with no retry.

## Expected behaviour

A transient HTTP 5xx or timeout on either download is retried a few times before the step fails. Digest verification is unchanged: retried bytes are still checked against the reviewed pin, and a persistent failure still fails closed.

## Acceptance criteria

- Both downloads pass curl's retry options (RED/GREEN in the installer's test).
- A persistent download failure is still refused and installs nothing (existing test).

**Size:** XS.

devantler:

## Why

Every platform change is checked by a job that downloads two signing and inventory tools from GitHub. When GitHub briefly answers with a server error, as it did repeatedly this afternoon, the job fails outright and blocks unrelated changes; the production deploy installs the same tools and would fail the same way.

## What

The downloads now retry a few times before giving up. The downloaded tools are still checked against the exact versions we reviewed, and a download that keeps failing still stops the job.

Fixes #4023


devantler:

**Tried as a user.** Ran the exact new download command (same flags, including `--retry 5 --retry-delay 3`) with the real `curl` against the pinned syft 1.51.1 release asset: exit 0, and the bytes match the reviewed SHA-256 pin. The retry path itself is curl's own behaviour on timeouts, 408, 429 and 5xx, so it is asserted rather than reproduced: the installer test now fails on `main` (`download does not retry transient failures`) and passes here, and its existing checks still prove a persistent download failure is refused with nothing installed.

**Evidence for the fix:** the failing step on #4020 and #4022 (and #4022's re-run) was `curl: (22) The requested URL returned error: 504` in this installer.

devantler:


@coderabbitai full review

devantler:


CodeRabbit refused this head (rate limited) and Codex is at its usage limit, so this round advances to Bugbot.

devantler:
@cursor review

devantler:

**Readiness at `f92770f`:** all checks green, including `🧪 Validate Manifests` running the changed installer; Cursor Bugbot green at this head with no findings (CodeRabbit refused this round on its hourly limit, Codex at its usage limit); 0 threads; tried as a user per the comment above. Promoting and adding to the merge queue.

PR #4027, opened 2026-09-21
ci(security): regenerate the approved revisions when actions publishes a release

### LINKED ISSUE #3960: security(supply-chain): accept the next released publish-app revision before tenants move to it

## Evidence

Measured for #3917: a tenant's `publish-app.yaml` pin can move within **1.3 h** of a new `devantler-tech/actions` revision. The approved-revision generator only learns a pin once it has landed on the tenant's default branch, and it regenerates once a day, with regeneration PRs merging a median ~3 h later. So the platform always accepts a new signer *after* the tenant starts using it.

## Problem

A narrowed publish matcher for `wedding-app` and `ascoachingogvaner` would reject the first release signed by a newly bumped pin. Replaying the releases since 2026-07-03, this would have happened in 2–3 of 14 cases.

## Expected

Each tenant's approved set also accepts the latest **released** `devantler-tech/actions` revision as a candidate. Regeneration also runs when `actions` publishes a release, so the platform accepts a revision before a dependency bump can move a tenant to it.

## Acceptance criteria

- [ ] The generator adds the latest released `actions` revision to each app tenant's approved set, derived rather than hand-written, and fails closed when it cannot resolve that revision.
- [ ] Regeneration runs when `devantler-tech/actions` publishes a release, in addition to the daily schedule.
- [ ] Tests cover the candidate column, including a case where the release cannot be read.

## Size

Small to medium.

Part of #3917

devantler:

## Why

The platform only learns about a new shared-workflow release once a day, but a tenant app can start signing with it within about an hour, so its first release after a bump could be rejected.

## What

Every 30 minutes a read-only check compares the approved release with the newest published one and starts the existing refresh when it is behind. No new cross-repository write access is added, and the refresh is not re-run when one is already pending or has already run since the release.

Fixes #3960


devantler:

**Tried as a user, at `84f55a5b`.** I ran the watcher in dry-run mode against the live repositories (read-only; nothing dispatched):

```
watch-actions-release: BEHIND: committed candidate(s) 6f47d8ca360fa3058f57175d54fa148deac08983 differ from release v13.6.7 (7a19c1d0f32e5a99a38d0a4106f1b2cbd76c99e4)
watch-actions-release: dry run; set WATCH_DISPATCH=1 to dispatch regenerate-publish-workflow-approved-revisions.yaml
```

That is the gap this closes, observed on real data: `v13.6.7` was published at 15:46Z today, the last regeneration ran on the 06:20Z schedule, so without the watcher the approved set would not name the new release until tomorrow morning. No regeneration branch is open and none has run since the release, so on merge the first scheduled poll would dispatch it.

The contract test stubs `gh` across 20 cases (current, behind, pending, already-ran, dry run, and every failed or implausible read exiting UNKNOWN without dispatching); three ablations of the pending, header and run-time guards each fail their case. actionlint and zizmor report no findings on the new workflow.

devantler:


@coderabbitai full review

devantler:


CodeRabbit refused this head (rate limited); Codex is out of usage, so the next lane is Cursor Bugbot.

devantler:

devantler:
@cursor review

PR #4033, opened 2026-09-21
fix(scheduling): let hcloud-volume pods preempt stateless pods on baseline workers

### LINKED ISSUE #4031: Volume-bound pods can be stranded Pending because stateless pods fill the baseline workers

## Evidence

On 2026-09-21 the merge-group deploy for #4029 restarted the Coroot ClickHouse shard replicas. One replica came back fine; the other has stayed **Pending** since ~21:28Z with `Insufficient memory`, and the deploy's own rollout gate timed out after 15 minutes, which evicted #4029 from the merge queue.

Read-only checks against prod (aggregate figures only; detail is in private operator notes):

- The replica can only run on a **baseline worker**: its data is an hcloud volume, and `require-baseline-workers-for-hcloud-volumes` pins such pods there. Its sibling replica occupies the third baseline worker (pod anti-affinity).
- The baseline workers are at **87–94 % of memory requested**. The replica's VPA-sized request (~1.3 GiB, matching live usage of ~1.1 GiB, so it is not over-sized) does not fit on either candidate.
- The autoscaler cannot help: autoscaled nodes are not baseline workers, so it reports `didn't match Pod's node affinity` and never scales up.
- Meanwhile, stateless pods with no volumes (for example Velero at ~1.2 GiB and the Kubescape operator at ~0.9 GiB) are sitting on those same baseline workers, because `prefer-baseline-nodes` deliberately steers every pod there for FinOps. The one autoscaled node is at ~23 % memory requested.
- Preemption does not rescue it: the replica runs at priority 0, the same as those stateless pods.

## Who is affected and how

Coroot's ClickHouse shard is running on one replica instead of two, so storage redundancy is lost until capacity frees up. More broadly, **any** restart of a volume-bound pod — a node drain, the pending Talos upgrade (#3713), an eviction — can strand it the same way, because stateless pods fill the only nodes it is allowed to use. Flux reports Ready throughout, so nothing alerts on it.

## Expected behaviour

A pod that can only run on a baseline worker always gets a place there. Stateless pods may use baseline capacity while it is spare, but they give it up (and move to autoscaled nodes, scaling up if needed) when a volume-bound pod needs it.

## Candidate fixes (to decide in the implementing PR)

1. **Priority for volume-bound pods.** A dedicated PriorityClass above 0 for pods carrying hcloud volumes, so the scheduler preempts stateless pods, which then reschedule onto autoscaled nodes. The Coroot operator does not expose `priorityClassName` for ClickHouse, so this needs an admission mutation; whether a webhook-set `priorityClassName` is honoured after the built-in Priority plugin has already resolved `.spec.priority` must be verified first.
2. **Keep large stateless pods off baseline workers.** Narrower, but it trades against the FinOps intent of `prefer-baseline-nodes` and does not protect against the next stateless workload that grows.

Option 1 fixes the class; option 2 only moves the current pods.

## Acceptance criteria

- A volume-bound pod restarted while baseline workers are full is scheduled (by preempting stateless pods) within a bounded time, shown on a fixture or live observation.
- Stateless pods still prefer baseline workers when there is room (the FinOps behaviour is unchanged).
- #4029's merge-group deploy passes its Coroot rollout gate.

Rough size: M.

devantler:

## Why

Services that keep their data on a cloud volume can only run on the fixed baseline servers, but every other service is steered onto those same servers to save money. When they fill up, a restarted data service has nowhere to go and stays down; this left one of the two monitoring database copies offline yesterday.

## What

Data services on cloud volumes now get a slightly higher scheduling priority, so the cluster moves a stateless service elsewhere to make room for them instead of leaving them waiting. A CI check keeps the two halves of the setting in step, because a mismatch would stop those services from starting at all. A service already waiting picks up the new priority when it is next recreated.

Fixes #4031


devantler:

**Tried at 1362b6eb** (agents never run a cluster here, so: offline policy evaluation, rendered output, and read-only prod checks).

- **Why both fields.** The upstream Priority admission plugin re-runs after a webhook mutates a Pod and, on CREATE, refuses a Pod whose `spec.priority` differs from its class's value. It has already set `spec.priority` to 0 by the time Kyverno sees the Pod, so a mutation setting only `priorityClassName` would get every hcloud-volume Pod refused. The policy therefore sets the class name and its value together.
- **Policy fixtures:** 11/11 pass — an hcloud claim and an hcloud ephemeral volume are patched with both fields; a Pod naming its own class, a Pod with no volumes, a default-class ephemeral volume and a Longhorn claim are left alone.
- **Ablations, each confirmed to fail for the intended reason:** dropping the priority value → 2 fixtures fail on resource diff; removing the hcloud storage-class check → the Longhorn fixture fails. The new drift check passes, and fails when the class value is changed to 1001.
- **Rendering:** the bootstrap layer renders the new class; `ksail --config ksail.prod.yaml workload validate` passes (658 files).
- **Read-only prod:** 13 Pods currently mount hcloud claims (Coroot ClickHouse and keepers, Coroot, Coroot Prometheus, OpenBao and its snapshot jobs); all run at priority 0 with no class, so all pick up the new priority on their next creation. No ResourceQuota uses a priority scope, so the new class is not blocked by quota. Both ClickHouse shard replicas are Running again at the moment; the defect remains for the next restart into a full baseline.

devantler:


@coderabbitai full review

PR #4039, opened 2026-09-22
fix(security): read an image referenced by tag and digest by its bare name

### LINKED ISSUE #4038: fix(security): the signature inventory misreads an image referenced by both tag and digest

**Evidence.** The scheduled *Validate Image Verifier Liveness* run on `main` went red on 2026-09-22 (run 35688286178), the first failure after 10 green daily runs. Its inventory job reported one `UNKNOWN`: the Coroot node agent referenced as `name:tag@sha256:…` returned HTTP 404 on its manifest read. The same digest, referenced as `name@sha256:…`, verified as `PASS` in the same run.

**Cause.** Coroot operator v0.10.0 (#4034) pins its node-agent image with both a tag and a digest. The inventory strips only the digest, so the tag stays on the repository name. Two things then go wrong:
- rule matching sees `…/platform-coroot-node-agent:<tag>`, misses the dedicated Coroot rule, and falls through to the catch-all rule, so the image is judged against the wrong signing identity;
- the manifest probe asks the registry for a repository whose name includes the tag, which does not exist, so the read fails with 404.

**Impact.** The daily check is red, and every tag-plus-digest reference is classified against the wrong rule, so the inventory's blast-radius verdict is wrong.

**Expected.** A reference with both a tag and a digest resolves to its bare repository for rule matching, and the probe reads it by digest.

**Acceptance criteria**
- A `name:tag@digest` reference matches the same rule as `name@digest` and is verified under that rule's identity.
- The manifest probe requests `/v2/<name>/manifests/<digest>`.
- A registry port (`host:5000/name:tag`) is not mistaken for a tag.
- Hermetic tests cover all three cases.

Size: S.

devantler:

## Why

The daily image-signature check went red today because it misread an image referenced by both a version label and a fixed fingerprint. It judged that image against the wrong signing rule and looked it up under a name the registry does not have, so it reported the image's signature as unknown.

## What

The check now reads such a reference by its plain image name, so the image is judged against its own signing rule and looked up correctly. The same Coroot image that was reported unknown today now verifies.

Fixes #4038


devantler:

**Readiness evidence @ 47702cc3**

- **RED/GREEN:** the 5 new assertions fail against `main`'s script. Notably, on `main` the catch-all identity *satisfied* a tagged image that has its own exact-name rule, which is the wrong-rule bug. All pass with the fix, and the full suite passes. `shellcheck` is clean on both files.
- **Tried as a user, against real GHCR:** I ran the inventory on the exact reference from run 35688286178 (`platform-coroot-node-agent:v1.35.8-alerts.1-…@sha256:e71f4d60…`).
  - `main`: `UNKNOWN … ghcr.io/devantler-tech/* … HTTP 404`, which reproduces the scheduled failure.
  - This branch: `PASS … ghcr.io/devantler-tech/platform-coroot-node-agent … signature accepted by the rule identity`.
- **Why the tagged reference appeared at all:** the scheduled run started 2s after the cluster autoscaler began scaling a node down (04:48:33Z). The terminating Coroot pod on that node had no resolved image ID, so the enumerator fell back to the pod spec's `tag@digest` form. The liveness job in the same run failed for the same scale-down: it could not reach the departing node. That is a separate fix.

devantler:


@coderabbitai review

PR #4041, opened 2026-09-22
fix(security): skip a node the autoscaler is removing in the verifier liveness check

### LINKED ISSUE #4040: fix(security): the image-verifier liveness check fails when the autoscaler removes a node mid-run

**Evidence.** The scheduled *Validate Image Verifier Liveness* run on `main` failed on 2026-09-22 (run 35688286178) with `cannot reach node 10.0.1.5`. The cluster autoscaler had started scaling that node down at 04:48:33Z, 2 seconds before the run started at 04:48:35Z. The node was still listed, but its server was already going away. It was the first failure after 10 green daily runs.

**Cause.** The check's convergence pass handles nodes that join or are replaced mid-sweep. An unreachable node, however, aborts the whole run before the inventory is re-read, so a node that is leaving turns the daily check red.

**Impact.** Every scheduled run that overlaps a scale-down goes red. A check that flaps with ordinary autoscaling gets ignored, which is as bad as not having it.

**Expected.** A node that is leaving the fleet does not fail the run. A node that is still a fleet member and cannot be reached still does.

**Acceptance criteria**
- An unreachable node carrying the autoscaler's removal taint is reported as skipped, and the run's verdict covers the nodes it checked.
- An unreachable node that disappears from the inventory between readings causes a re-check of the remaining fleet, not an abort.
- An unreachable node with no removal marker, still listed on re-read, aborts as before.
- A removal-marked node that still answers is checked like any other.
- A fleet where every node is leaving never reads as clean.

Size: S.

devantler:

## Why

The daily check that every cluster node enforces image signatures went red today because the autoscaler was removing a node at the moment it ran. A check that fails whenever the cluster scales down gets ignored, which defeats its purpose.

## What

A node that is being removed and has already stopped answering is now reported as skipped. The check re-reads the node list before deciding, and a node that is still a normal member but cannot be reached still fails the check.

Fixes #4040


devantler:

**Readiness evidence @ e1b1f5e8**

- **RED/GREEN:** 5 new cases (23–23d), each with a control that differs in one fixture. Against `main`'s script, case 23 fails (the departing node aborts the run). With the fix the full suite passes. `shellcheck` is clean.
- **Tried against prod:** I ran the fixed script with discovery against the live cluster. All 7 nodes reported `OK`, which confirms the new discovery read (UID, address and removal-taint role) parses real node objects. The departure path itself needs a live scale-down to observe, and the cases above pin it.
- **Scope of the relaxation:** a node is skipped only when it carries the autoscaler's `ToBeDeletedByClusterAutoscaler` taint **and** stops answering. A removal-marked node that still answers is inspected (case 23c). An unmarked unreachable node still aborts (23a). A fleet with nothing checkable never reads clean (23d).
- **Companion:** the inventory job in the same run failed for the same scale-down; its fix is #4039.

devantler:


@coderabbitai review

devantler:


@coderabbitai review

PR #4052, opened 2026-09-22
fix(alertmanager): drop the host clause from Slack titles of hostless alerts

### LINKED ISSUE #3081: Slack alert titles render "on <no value>" for alerts without a host label

**Evidence.** Alertmanager's Slack receiver builds its title as `... {{ .CommonLabels.alertname }} on {{ .CommonLabels.host }}`. `host` is a **node-agent** label, so any alert that is not node-scoped renders a dangling clause. Measured by running the exact rendered template through Go's `text/template`:

- node-agent alert → `[FIRING] KubescapeRuleViolated on worker-1`
- hostless alert → `[FIRING] CrossplaneManagedResourceNotSynced on <no value>`

**Who it affects.** Anyone reading `#platform-alerts`. It is live now: the crossplane-sync-alerter added in #3080 posts cluster-scoped alerts with no `host`, and any future non-node alert on this catch-all route inherits the same title.

**Expected.** The clause is omitted when the label is absent, and node-agent titles are unchanged.

**Acceptance criteria.**
- Guard the label (`{{ with .CommonLabels.host }} on {{ . }}{{ end }}` renders correctly in both cases — verified).
- Re-approve the moved EKS authorization fingerprints, and say in the PR why a Slack title moved them.
- Existing node-agent titles are byte-identical.

**Why it is not folded into #3080.** The Alertmanager HelmRelease sits inside the EKS authorization surface — `isControllerRBACEmitter` pins the whole release, because a chart's values can change RBAC — so this one-line cosmetic change moves an approved fingerprint. Re-approving a security pin deserves its own review rather than riding along in an alerting PR (the same re-approval discipline d365ad0e followed).

**Size.** Small: one template line plus the fingerprint re-approval.

devantler:

## Why

Platform alerts that are not tied to a single machine reach the alerts channel with a placeholder where a machine name should be, which reads as broken and hides which alert fired.

## What

Those titles now end after the alert name; titles for machine-level alerts are unchanged. Because the alerting release sits inside the reviewed permissions boundary, its approval record is refreshed too — nothing is granted.

Fixes #3081


devantler:


@coderabbitai full review

devantler:

**Evaluation at `8c3580b7360c`**

- **Behaviour, RED/GREEN:** the Slack title template taken from `main` and from this head, rendered with Go's `text/template` (the engine Alertmanager uses) for a node-agent alert and a hostless alert:
  - `main`: `[FIRING] KubescapeRuleViolated on worker-1` / `[FIRING] CrossplaneManagedResourceNotSynced on <no value>`
  - this head: `[FIRING] KubescapeRuleViolated on worker-1` / `[FIRING] CrossplaneManagedResourceNotSynced`
  Node-agent titles are byte-identical; the dangling clause is gone.
- **Authorization surface:** the required job on the approved kubectl v1.36.2 renderer (job 106809573713, at `dad8fecd`) reported exactly one moved entry against main `a0626878` — the `kubescape/alertmanager` HelmRelease — and aggregate `65fecc82…`. That value is re-approved in the second commit with the moved entry, the base and the renderer recorded; nothing is granted. `approval-base` passes against main locally, and the full required CI is green at this head.
- **Not observed live:** the change reaches Slack only after deploy; confirming a real hostless alert title in `#platform-alerts` is the post-merge verification (the issue stays in Verifying until then).

